PR #18, opened 2015-12-15
print() for py3

jona-sassenhagen:
These are the only things I ran into problems with with python 3.

### DIFF
--- a/pyriemann/utils/mean.py
+++ b/pyriemann/utils/mean.py
@@ -130,7 +130,7 @@ def mean_logdet(covmats, tol=10e-5, maxiter=50, init=None, sample_weight=None):
 
         C = Cnew
     if k == maxiter:
-        print 'Max iter reach'
+        print('Max iter reach')
     return C
 
 
@@ -182,7 +182,7 @@ def mean_wasserstein(covmats, tol=10e-4, maxiter=50, init=None,
         crit = numpy.linalg.norm(Knew - K, ord='fro')
         K = Knew
     if k == maxiter:
-        print 'Max iter reach'
+        print('Max iter reach')
     C = numpy.dot(K, K)
     return C
 


wmvanvliet:
you should also add `from __future__ import print_function` so it still works with Python 2

jona-sassenhagen:
https://github.com/alexandrebarachant/pyRiemann/search?utf8=✓&q=print

jona-sassenhagen:
Oh, and to make it more explicit:
```
$/usr/bin/python
Python 2.6.6 (r266:84292, Aug 12 2014, 07:57:07)
[GCC 4.4.5] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> print("hi")
hi
>>>
```

wmvanvliet:
Here is mine:

    Python 2.7.11 |Continuum Analytics, Inc.| (default, Dec  6 2015, 18:08:32) 
    [GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
    Type "help", "copyright", "credits" or "license" for more information.
    Anaconda is brought to you by Continuum Analytics.
    Please check out: http://continuum.io/thanks and https://anaconda.org
    >>> x = 1
    >>> print('The value of x is:', x)
    ('The value of x is:', 1)
    >>> from __future__ import print_function
    >>> print('The value of x is:', x)       
    The value of x is: 1

jona-sassenhagen:
Yes, but in this case, it is literally only `print('some_string')` so I don't see any need to introduce an extra line. The rest of the (python 2) code is full of `print('some_string')` already, and no print_function.

wmvanvliet:
otherwise `print('some_string')` looks like a function call, but it is not. It freaks me out :) But it's @alexandrebarachant's call of course.

jona-sassenhagen:
That's all over mne too, isn't it? The examples for example all start with `print(__doc__)` and none import from `__future__`.

jona-sassenhagen:
I don't think the Travis error is related to the commit?

alexandrebarachant:
Thanks a lot for checking the py3 compatibility.

Indeed, the travis error is not related to the issue.

I will rework a bit the travis integration for py3 testing.

PR #21, opened 2016-02-25
FIX: scikit-learn LDA import

kingjr:
The current sklearn import is outdated, and will be deprecated in .19

This PR checks the sklearn version and import and LDA accordingly.

Also switch the LDA of the `plot_single` example into a LogisticRegression so as to keep it clean.

Note that the cv will be changing from v.18, (they are now in `model_selection`) and have a slighlty different behavior (`for train, test in cv.split()` instead of `for train, test in cv`)

### DIFF
--- a/examples/motor-imagery/plot_single.py
+++ b/examples/motor-imagery/plot_single.py
@@ -28,7 +28,7 @@ from pyriemann.estimation import covariances
 from sklearn.cross_validation import cross_val_score, KFold
 from sklearn.pipeline import Pipeline
 from sklearn.linear_model import LogisticRegression
-from sklearn.lda import LDA
+from sklearn.lda import LDA  # XXX has changed since v.17
 
 ###############################################################################
 # Set parameters and read data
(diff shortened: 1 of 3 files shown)


alexandrebarachant:
Thanks JR :)
I was too lazy to do it ....

i'm actually thinking about removing the FGDA for something more generic, so there is a good chance the LDA will be removed soon. But that's another story.

kingjr:
I agree, it'd make more sense to just provide the Riemann specificity in this package, and let the user build the estimators pipelines. It'd also be less work for the future.

alexandrebarachant:
the FGDA is actually a legacy function inherited from my covariance toolbox.
The general idea is to project data in the tangent space, apply some
preprocessing (here LDA, but it could be a PCA (like in the PGA
<http://www.cs.unc.edu/Research/MIDAG/pubs/papers/TMI04_Fletcher_PGA.pdf>),
or an ICA), and projecting back the filtered datapoints in the manifold
where they will be classified with a simple distance based classifier.

It can actually be more efficient than just doing the classification in the
tangent space (for some reasons ...), but it is actually very
computationally.

The plan is to move that to a new file (geodesicfiltering.py ?) where you
can pick your own transformer to do the filtering.
However, i'm not sure we can put this in a form of a sklearn pipeline,
because the back-projection need to be done by the same object that did the
projection.

Anyway, it's low priority for now, so thanks for your PR :)

PR #23, opened 2016-02-29
FIX: compatibility with sklearn get_params 

kingjr:
Some of the  covariance classes didn't respect the sklearn `TransformerMixin` syntax where the params have to be stored in attributes with the same names.

Consequently, we couldn't `get_params()` them, and thus cannot `sklearn.base.clone()`, and thus can't use the `mne.decoding.GeneralizationAcrossTime` class.

(Also there are a bunch of params test missing to ensure that the class params are `float`, `int`, `str` etc)

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -28,6 +28,8 @@ v0.2.4 (Unreleased)
 
 - Added Multiclass CSP.
 
+- API: param name changes in `estimations.CospCovariances` to comply to Scikit-Learn.
+
 v0.2.3 (November 2015)
 ----------------------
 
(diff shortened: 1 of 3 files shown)


kingjr:
Also, I haven't found where `phase_correction` is used.

alexandrebarachant:
This part of the code has not been through my code review, i think most of the other classes respect this convention.

phase_correction is not used (see utils.covariance.cospectrum) so you can remove it.

Let me know when your done, i will merge.
Also, if you want to add some test for the cloning process, be my guess :)

kingjr:
> phase_correction is not used (see utils.covariance.cospectrum) so you can remove it.

ok

> Also, if you want to add some test for the cloning process, be my guest :)

From what I know, sklearn clone fully depends on `get_params`, so these new tests should suffice.

kingjr:
ready for review

alexandrebarachant:
Looks good to me,
If you want, you can add something in the doc/whatsnew.rst

kingjr:
Ok, ready on my ends

Issue #19, opened 2016-01-22
Fix Data generation in test suite


alexandrebarachant:
in the test suite, the generation of covariance matrices is flawed and can create non-SPD matrice, causing the test to fail.
- possible fix : add test data

kingjr:
You can set the random state in numpy.

alexandrebarachant:
the seed is set in most of the tests. 

However, i want to use realistic data and results obtained with a reference implementation to make sure that no regressions are introduced.

So far, the tests are just verifying that the code is not trowing errors.

aj-ptw:
Ah functional tests. I have one of these written for the Potato class and am working on one for the TSClassifier. I could make a PR for the Potato one with a minimum dataset i've been using.

qbarthelemy:
I think that problem of non SPD matrices is solved in #149, ensuring that `make_covariances` generates positive eigen values and orthogonal eigen vectors:
https://github.com/pyRiemann/pyRiemann/pull/149/files#diff-9cd79c829466352eb2b1c6531e7f8b71b3157b3fa43264f8df15db931d1d4544R39-R40

sylvchev:
I agree, this is solve now. We could close the issue. Thanks for the POC with covmatest @gcattan

Issue #1, opened 2015-04-19
DocString documentation for all the classes


alexandrebarachant:
Add doctstring for the different classes :
- [x] classification
- [x] clustering
- [x] tangent space
- [x] estimation
- [x] spatial filter
- [x] stats
- [x] channel selection

PR #24, opened 2016-03-01
FIX: attribute convention ala sklearn

kingjr:
sklearn naming convention is to have init params given by the subject without a `_` at the end and attributes created by the method with a `_`.

e.g.
```
myclass = MyClass(foo=1)
myclass.foo
myclass.fit()
myclass.bar_
```

That's a lot of changes, some of which are backward incompatible, but I think it's generally safer to stick to their naming convention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ For the latest version, you can install the package from the sources using the s
 python setup.py install
 ```
 
-## how to use it
+## How to use it
 
 Most of the functions mimic the scikit-learn API, and therefore can be directly used with sklearn. For example, for cross-validation classification of EEG signal using the MDM algorithm described in [4] , it is easy as :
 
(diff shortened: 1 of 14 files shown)


alexandrebarachant:
wow, thats a lot of works :)

i agree, the toolbox is still young so we can do this kind of change, and it's better to stick to the sklearn conventions.

Can you check the examples, i think i access internal variables to plot things.

Don't worry for travis, it's related to #19

kingjr:
Ok. I also removed the channel renaming in the examples, because mne-python doesn't allow manual renaming manually anymore https://github.com/mne-tools/mne-python/pull/1124

alexandrebarachant:
So far, i'm happy with this PR. 

Perhaps we could add a "contribution guideline" in the Readme, stating the different conventions.

When your ready, update the What's new and i will merge.

Note that i plan to rewrite almost completly the stats functions in a near future (the code is old and too convoluted)

kingjr:
Ok. I haven't looked at the way stats work yet. I've kept the contribution guidelines minimalistic 1) because I'm lazy) and 2) because they can become super overwhelming. Most packages actually refer to an external doc.

BTW, you may want to go through all files with flake8.

alexandrebarachant:
Thanks.

File i edited recently are supposed to be Flake8 verified and properly documented.
There are only a few files where i did not finished it: stats.py and estimation.py + some of the examples.
This is actually one of my milestone to release a new major version (see #1)

kingjr:
all right thanks!

PR #27, opened 2016-07-01
ENH: allow precomputed baseline covariance

kingjr:
Here is what I was suggesting to compute the "baseline" cov with unlabeled data.

LTKWYT

### DIFF
--- a/pyriemann/estimation.py
+++ b/pyriemann/estimation.py
@@ -206,17 +206,21 @@ class XdawnCovariances(BaseEstimator, TransformerMixin):
     """
 
     def __init__(self, nfilter=4, applyfilters=True, classes=None,
-                 estimator='scm', xdawn_estimator='scm'):
+                 estimator='scm', xdawn_estimator='scm', baseline_cov=None):
         """Init."""
         self.applyfilters = applyfilters
         self.estimator = estimator
         self.xdawn_estimator = xdawn_estimator
         self.classes = classes
         self.nfilter = nfilter
+        self.baseline_cov = baseline_cov
 
     def fit(self, X, y):
+        # FIXME self.Xd should be init at XdawnCovariances.__init__ and be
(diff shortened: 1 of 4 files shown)


alexandrebarachant:
I don't see any objection.

Can we find a better name for the parameter ? like baseline_cov or something more straingtforward.

just 2 requests: 
- make the doc more explicit, what the parameter does, etc
- pipe it for the `XdawnCovariances` estimator from 'estimation.py`

pyriemann/estimation.py
```diff
@@ -206,17 +206,21 @@ class XdawnCovariances(BaseEstimator, TransformerMixin):
...
         self.applyfilters = applyfilters
         self.estimator = estimator
         self.xdawn_estimator = xdawn_estimator
         self.classes = classes
         self.nfilter = nfilter
+        self.baseline_cov = baseline_cov
 
     def fit(self, X, y):
+        # FIXME self.Xd should be init at XdawnCovariances.__init__ and be
+        # replaced as param not attribute
```
kingjr: I didn't fix this in this PR. I'm not sure whether you want to break the API or whether you think deprecation cycles are necessary.

alexandrebarachant: I think in encountered problem when object are initialized in the init. it caused issues for `deepcopy()` of the object in scikit-learn pipeline. it's something to check because it's all over the place in the toolbox.

alexandrebarachant: see  d45a374ff44dd497e32bf780dcb7a2db99753cba

alexandrebarachant: but if we add it as a param of the init, it may works.

kingjr: Ah OK I see.

alexandrebarachant:
Thanks :)

PR #29, opened 2016-07-11
ENH: handle flat channels in Xdawn

kingjr:
Xdawn generates empty filters if there s a bad channels:

```
import numpy as np
from pyriemann.spatialfilters import Xdawn
X = np.random.rand(100, 30, 20)
y = np.random.randint(0, 2, 100)
X[:, 20:, :] = 0
Xdawn(estimator='oas').fit_transform(X, y)
```

This fixes it

(I also added some tests to the baseline covariance)

### DIFF
--- a/pyriemann/spatialfilters.py
+++ b/pyriemann/spatialfilters.py
@@ -41,7 +41,9 @@ class Xdawn(BaseEstimator, TransformerMixin):
         type, concatenated, else empty.
     evokeds_ : ndarray
         If fit, the evoked response for each event type, concatenated.
-
+    picks_ : ndarray
+        Channels indices used in the computation. Can be different from
+        range(n_channels) if some channels do not vary over time.
 
     See Also
     --------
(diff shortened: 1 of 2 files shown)


kingjr:
I'm not sure whether `picks` should be stored publicly, probably a private attribute should be better

alexandrebarachant:
Hum,
Would it be preferable to drop the bad channels before feeding the data to xDawn ?

alexandrebarachant:
Hi again,
I appreciate the contrib, and generally speaking we will face similar problem for any other method. 

Flat channels creates rank-deficient mat-cov, and despite the regularization it leads to error in the joint diagonalization.

I would suggest you refactor this in a `FlatChannelDetector` transformer (or something with a better name) so that it can be pipelined before any other method.

WDYT ?

kingjr:
yes, definitely a better idea. I'll close this PR, and open a new one to keep it clean.

PR #31, opened 2016-07-21
FIX: raises in CSP

kingjr:
There was a buggy raise when using less than 2 classes e.g. `CSP().fit(X, y=np.zeros(len(X)))`, so I added a series of tests.


alexandrebarachant:
@kingjr Nice catch !

one small change, the check for the metric is for the mean and not the distance. Also, it should accept custom callable object. It would be preferable to do this check in `utils.mean.mean_covariance`, or to create a (private ?) method in `utils.mean` that makes this check.

kingjr:
I don't have this error locally, is it a linalg problem?

alexandrebarachant:
ho yep, that's related to #19 i believe. 
i started again travis, that should be fine.

Issue #33, opened 2016-07-28
Tracking: Tutorials


andrewjaykeller:
An issue to track the creation and testing of tutorials.

alexandrebarachant:
Thank AJ,

What kind of tutorial would you like, i'm planning to do at least 2 : 
- [ ] Motor imagery classification
- [ ] Evoked potential classification

it might worth it do to a small one about spatial filtering as a way to reduce dimension. 

@kingjr : any other idea ?

pushtheworldllc:
If you can provide those that would be great!

kingjr:
What is the question exactly?

alexandrebarachant:
i should create some tutorial to help people getting started. beside the two obvious tutorial on how to use pyRiemann for motor imagery and ERP classification, do you have any other idea/request.

It could be a way to explain and devellop non trivial questions or tackle some common problems. I think one about dimensionality reduction and regularization of covariance matrices is appropriate.

kingjr:
yes, I would explain the cov regularization, that seems like a common blocker

kingjr:
maybe another for the new objects you recently implemented?

pushtheworldllc:
Potato!
* Building your first potato* Offline use of a potato for post processing* Online use for real time artifact rejection  

AJ Keller
Founder 
Push The World
843-471-0039




On Thu, Jul 28, 2016 at 9:28 PM -0400, "Jean-Rémi KING" <notifications@github.com> wrote:












maybe another for the new objects you recently implemented?

qbarthelemy:
Regarding all examples added in the gallery since v0.2.7, I think this issue could be closed.

PR #35, opened 2016-08-12
FIX: decommented imports in __init__.py to make README example work

emanuele:
Currently, the example in the README file does not work because imports fail, due to issues in `pyriemann/__init__.py`. This little PR fixes that.

### DIFF
--- a/pyriemann/__init__.py
+++ b/pyriemann/__init__.py
@@ -1,7 +1,8 @@
-import classification
-import tangentspace
-import channelselection
-import estimation
-# import permutations
-import spatialfilters
-import clustering
+from . import classification
+from . import tangentspace
+from . import channelselection
+from . import estimation
+# from . import permutations
+from . import spatialfilters
+from . import clustering
+from . import stats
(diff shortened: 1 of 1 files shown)


kingjr:
LGTM

alexandrebarachant:
looks like travis with python 3 raise some import error.

i think we should do some relative import.

can you add clustering too

emanuele:
OK for importing clustering (stats too?).
Unfortunately I cannot directly test for Python 3 on my laptop at the moment. Will try to dig into that tomorrow.

alexandrebarachant:
you can import stats too, it will be cleaner once i merged #37 

i think you need to do relative import, ie : 

```python
from . import classification
```

emanuele:
Thanks for the suggestion and sorry for the latency. I'm on holidays these days :) but couldn't spent time on this one :(. Will do now.

alexandrebarachant:
no problem, take your time.
the example on the readme is a bit outdated, you can check here for working examples : http://pythonhosted.org/pyriemann/index.html

i'm planning some tutorial too (see #33 ), i hope it will be ready for the next release

alexandrebarachant:
Okay, all good. 
thanks

emanuele:
Thanks to you for pyRiemann!!

PR #45, opened 2016-10-25
Add predict_proba function to potato

aj-ptw:
Add: `predict_proba` to `Potato`. 

`0.0` if the trial is clean, and `1.0` if the trial contains an artifact


alexandrebarachant:
Hey,

Thanks for that.
this is already covered by `predict` !

predict_proba should return a probability i.e. a scalar between 0 and 1.
it should be relatively easy to convert the z score into a probability score.

PR #47, opened 2017-02-06
Spectral Transfer Learning with Information Geometry

andrewjaykeller:
Adds support for spectral learning with information geometry!


alexandrebarachant:
Thanks AJ.
I'm going to test this on my side, and give you feedback.
It should take a few days as i'm quite busy lately.

aj-ptw:
@alexandrebarachant great! I will be continually working on this as well as I integrate it into thinker :) very much looking forward to your feedback

alexandrebarachant:
Okay, i made a few tests. some comments

- please modify the fit function so it can fit estimators like it is done in `VotingClassifier`. you can directly copy the code from sklearn.

- I only get prediction of the same classes (i.e accuracy 50% on a balanced dataset). something does not works properly.

I'm joining a script so you can test it. this is cross-session with multiple pipeline
[test_stig.zip](https://github.com/alexandrebarachant/pyRiemann/files/801156/test_stig.zip)

PR #54, opened 2017-04-01
Visualizing Covariance matrices with Diffusion maps

plcrodrigues:
Created the Embedding class
Created a new example using data from the MNE dataset with ERP EEG
Created some basic tests for the new class

### DIFF
--- a/examples/ERP/plot_embedding_EEG.py
+++ b/examples/ERP/plot_embedding_EEG.py
@@ -12,7 +12,7 @@ import numpy as np
 
 from pyriemann.estimation import ERPCovariances, XdawnCovariances
 from pyriemann.tangentspace import TangentSpace
-from pyriemann.embedding import CovEmbedding
+from pyriemann.embedding import Embedding
 
 import mne
 from mne import io
@@ -21,7 +21,6 @@ from mne.datasets import sample
 from sklearn.pipeline import make_pipeline
 from sklearn.cross_validation import KFold
 from sklearn.linear_model import LogisticRegression
-
 from matplotlib import pyplot as plt
 from mpl_toolkits.mplot3d import Axes3D
 
@@ -62,8 +61,8 @@ y = epochs.events[:, -1]
 # Embedding the Xdawn covariance matrices with Diffusion maps
 
 #covs = XdawnCovariances(nfilter=4).fit_transform(X,y)
-covs = ERPCovariances(estimator='oas', classes=[3,4]).fit_transform(X,y)
-lapl = CovEmbedding(metric='riemann', n_components=3)
+covs = ERPCovariances(estimator='oas', classes=[3, 4]).fit_transform(X, y)
+lapl = Embedding(metric='riemann', n_components=3)
 embd = lapl.fit_transform(covs)
 
 #%%
@@ -71,19 +70,11 @@ embd = lapl.fit_transform(covs)
 ###############################################################################
 # Plot the three first components of the embedded points
 
-fig,ax = plt.subplots(figsize=(7,8), facecolor='white')
-      
+fig, ax = plt.subplots(figsize=(7, 8), facecolor='white')
+
 for label in np.unique(y):
-    idx = (y==label)
-    ax.scatter(embd[idx,0], embd[idx,1], s=36)      
-    
-ax.set_xlabel(r'$\varphi_1$', fontsize=16)        
-ax.set_ylabel(r'$\varphi_2$', fontsize=16)        
+    idx = (y == label)
+    ax.scatter(embd[idx, 0], embd[idx, 1], s=36)
 
-
-
-
-
-
-
- 
+ax.set_xlabel(r'$\varphi_1$', fontsize=16)
+ax.set_ylabel(r'$\varphi_2$', fontsize=16)
--- a/examples/ERP/plot_embedding_EEG.py
+++ b/examples/ERP/plot_embedding_EEG.py
@@ -63,22 +63,19 @@ y = epochs.events[:, -1]
 
 #covs = XdawnCovariances(nfilter=4).fit_transform(X,y)
 covs = ERPCovariances(estimator='oas', classes=[3,4]).fit_transform(X,y)
-diff = CovEmbedding(metric='riemann', n_components=3)
-
-#%%
-
-u = diff.fit_transform(covs)
+lapl = CovEmbedding(metric='riemann', n_components=3)
+embd = lapl.fit_transform(covs)
 
 #%%
 
 ###############################################################################
 # Plot the three first components of the embedded points
 
-fig,ax = plt.subplots(figsize=(6.54, 5.66), facecolor='white')
+fig,ax = plt.subplots(figsize=(7,8), facecolor='white')
       
 for label in np.unique(y):
     idx = (y==label)
-    ax.scatter(u[idx,0], u[idx,1], s=36)      
+    ax.scatter(embd[idx,0], embd[idx,1], s=36)      
     
 ax.set_xlabel(r'$\varphi_1$', fontsize=16)        
 ax.set_ylabel(r'$\varphi_2$', fontsize=16)        
--- a/examples/ERP/plot_embedding_EEG.py
+++ b/examples/ERP/plot_embedding_EEG.py
@@ -1,6 +1,6 @@
 """
 ====================================================================
-Embedding ERP EEG data in 2D Euclidean space with Diffusion maps
+Embedding ERP EEG data in 3D Euclidean space with Diffusion maps
 ====================================================================
 
 """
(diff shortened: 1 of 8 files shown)


pyriemann/embedding.py
```diff
@@ -0,0 +1,134 @@
...
+from pyriemann.utils.distance import distance
+    
+class Embedding(BaseEstimator, TransformerMixin):
+
+    """Embed SPD matrices into an Euclidean space of smaller dimension.
+
+    It uses diffusion maps to embed the SPD matrices into an Euclidean space.
+    The Euclidean distance between points in this new space approximates 
+    the Diffusion distance (also called commute distance) between vertices 
+    of a graph where each SPD matrix is a vertex.
```
alexandrebarachant: Do you have a reference for this work ?

pyriemann/embedding.py
```diff
@@ -0,0 +1,134 @@
...
+            ndarray of SPD matrices.
+
+        Returns
+        -------
+        self : Embedding instance
+        The Embedding instance.
+        """
+        return self
+
+    def transform(self, X, eps=None, tdiff=0.0):
```
alexandrebarachant: keep things compatible with sklearn API.
eps and tdiff should be set as class arguments

pyriemann/embedding.py
```diff
@@ -0,0 +1,134 @@
...
+    # normalize rows of the kernel so it becomes a prob transition matrix    
+    d = np.sqrt(np.dot(kernel, np.ones(len(kernel))))
+    P = np.divide(kernel, np.outer(d, d))  
+    
+    return P
+
+def get_Embedding(points, metric, eps=None, tdiff=0):
+    
+    # this implementation follows the algorithm in page 34 of 
+    # Stephane's Lafon PhD thesis "Diffusion Maps and Geometric Harmonics"
```
alexandrebarachant: This looks super similar to the Spectral Embedding of scikit learn : http://scikit-learn.org/stable/modules/generated/sklearn.manifold.SpectralEmbedding.html#sklearn.manifold.SpectralEmbedding

Can you comment on the difference

plcrodrigues: Yeah, I agree with you. 

However, please note that this function actually implements the **Laplacian Eigenmaps** algorithm and not the Diffusion Maps (it is stated in the docstring).

The differences between these two algorithms is really just **one** line: Diffusion maps do an extra normalization in the affinity matrix before doing the eigendecomposition, making it more robust to what the authors call "non uniform sampling of the datapoints". This means that the embedding of your points is less dependent of the probability density used to sample your data points and reflects more precisely the actual geometry of your dataset. 

To be honest, I haven't seen much difference in practice between the results with these two methods, so if you think it would be better to keep things simple, we can use directly the implementation from sklearn. I could make things with a wrapper so that it would be more adapted to covariance matrices and the context of Pyriemann.

pyriemann/embedding.py
plcrodrigues: One of my favourite references for this topic is

S.Lafon, Y. Keller, R. Coifman, "Data fusion and multicue data matching by diffusion maps" (2006) on IEEE Transactions of Pattern Analysis and Machine Intelligence.

pyriemann/embedding.py
alexandrebarachant: Okay, thanks for the detailed explanation.

They have exposed a method that does the decomposition (i guess) and [require just the affinity matrix](http://scikit-learn.org/stable/modules/generated/sklearn.manifold.spectral_embedding.html#sklearn.manifold.spectral_embedding). So if it is possible to do the extra step before the decomposition i would prefer to use the sklearn method.

pyriemann/embedding.py
alexandrebarachant: Perfect, 
try to put it in the docstring following the [numpydoc convention](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) (see the reference section)

pyriemann/embedding.py
alexandrebarachant: And if it's not possible to use their code, then go ahead with your own implementation.

pyriemann/embedding.py
plcrodrigues: Ah ok, haha, thought you were asking a reference for reading yourself :P

pyriemann/embedding.py
plcrodrigues: Ok, cool, I will see how compatible things are and get back soon to you.

alexandrebarachant:
Once you are done with the comments, add something to `doc/whatsnew.rst` (with your name, see how they do it in [MNE](https://raw.githubusercontent.com/mne-tools/mne-python/master/doc/whats_new.rst)) as well as in the README.md.

then add a new section for the embedding in the `doc/api.rst` so in can appears in the documentation.

alexandrebarachant:
Any update ?

plcrodrigues:
Hey, sorry for such a long delay :(

So, I changed things so that we use the `SpectralEmbedding` from Sklearn and also the `pairwise_distance` function from PyRiemann. I think now everything is compatible, right?

In fact, now the embedding is done using Laplacian Eingenmaps (LE) instead of Diffusion Maps (DM). I think is better like this: less confusion (LE is better known than DM) and I have not seen so much superiority from DM to LE in practice (apart from theoretical explanations and interpretations). 

I included a line to the `doc/whatsnew.rst`, however the file does not follow the same standard from MNE as you mentioned. If I include my name if will be only line with an author...

examples/ERP/plot_embedding_EEG.py
```diff
@@ -0,0 +1,68 @@
...
+epochs = mne.Epochs(raw, events, event_id, tmin, tmax, proj=False,
+                    picks=picks, baseline=None, preload=True, verbose=False)
+
+X = epochs.get_data()
+y = epochs.events[:, -1]
+
+###############################################################################
+# Embedding the Xdawn covariance matrices with Diffusion maps
+
+covs = ERPCovariances(estimator='oas', classes=[3, 4]).fit_transform(X, y)
```
alexandrebarachant: ERPCov is also Supervised. 
Using XdawnCov was fine (and probably faster), just cross-validates the estimation i.e

```
train = range(0, len(y), 2)
test = range(1, len(y), 2)
covs = XdawnCov().fit(X[train], y[train]).transform(X[test])
```

pyriemann/embedding.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self, n_components=2, metric='riemann', eps=None):
+        """Init."""
+        self.metric = metric
+        self.n_components = n_components
+        self.eps = eps
+
+    def _get_affinity_matrix(self, X, eps):
+
+        # make matrix with pairwise distances between points
+        distmatrix = pairwise_distance(X, metric='riemann')
```
alexandrebarachant: self.metric

pyriemann/embedding.py
```diff
@@ -0,0 +1,99 @@
...
+            eps = np.median(distmatrix)**2 / 2
+
+        # make kernel matrix from the distance matrix
+        kernel = np.exp(-distmatrix**2 / (4 * eps))
+
+        # normalize the kernel matrix
+        q = np.dot(kernel, np.ones(len(kernel)))
+        kernel_n = np.divide(kernel, np.outer(q, q))
+
+        self.affinity_matrix_ = kernel_n
```
alexandrebarachant: Do you need to store the matrix ?

pyriemann/utils/distance.py
```diff
@@ -133,6 +133,23 @@ def distance(A, B, metric='riemann'):
 
     return d
 
+def pairwise_distance(X, Y=None, metric='riemann'):
+    """Pairwise distance matrix"""
```
alexandrebarachant: While you are at it, can you add docstring about the arguments.
I probably forgot to do it myself.

tests/test_embedding.py
```diff
@@ -0,0 +1,24 @@
...
+    covmats = np.empty((Nt, Ne, Ne))
+    for i in range(Nt):
+        covmats[i] = np.dot(np.dot(A, np.diag(diags[i])), A.T)
+    return covmats, diags, A
+
+
+def test_embedding_firstcomponent():
+    """Test Embedding."""
+    covmats, diags, A = generate_cov(100, 3)
+    u, l = Embedding(metric='riemann').fit_transform(covmats)
```
alexandrebarachant: This creates an error in Travis

plcrodrigues:
Here's the figure generated by the embedding function.
figure_1

alexandrebarachant:
looks good to me. Out of curiosity, what would be the results with the 4 classes contained in the dataset ?

plcrodrigues:
Here's the result with 4 classes.
figure_2

alexandrebarachant:
very nice 3D plot. 

I actually switched the example to MEG data, and 4 classes and fixed a couple of issues. Its in the master branch. 

here is the nex plot : 
image

PR #74, opened 2019-10-25
Avoid joblib warning

maxdolle:
Replace the `from sklearn.externals.joblib import ...` into `from joblib import ...` to avoid the following warning (that will eventually turn into an error with sklearn==0.23):
```
sklearn.externals.joblib is deprecated in 0.21 and will be removed in 0.23. Please import this functionality directly from joblib, which can be installed with: pip install joblib. If this warning is raised when loading pickled models, you may need to re-serialize those models with scikit-learn 0.21+.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
   apt:
     packages:
       - libatlas-dev
-      - libatlas3gf-base
+      - libatlas3-base
       - libblas-dev
       - liblapack-dev
       - python-matplotlib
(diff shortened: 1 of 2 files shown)


alexandrebarachant:
Thanks, this is great. 

Could you look at the test and try to fix them. I dont think its related to this PR but would be awesome to bring them back to life.

maxdolle:
Here it is.
I switched to travis virtualenv instead of conda as there are no hard binary dependencies that whould require it.

PR #76, opened 2020-01-24
Update plot_classify_EEG_tangentspace.py

davestanley:
Fixed error `No module named 'sklearn.cross_validation'` for import kfold


alexandrebarachant:
Awesome thanks, this is much appreciated !

Would it be possible to fix the other example as well, the same issue is present in a few places.

PR #79, opened 2020-03-18
various fix and improvements

alexandrebarachant:


examples/ERP/plot_classify_MEG_mdm.py
```diff
@@ -62,7 +71,7 @@
 n_components = 3  # pick some components
 
 # Define a monte-carlo cross-validation generator (reduce variance):
```
agramfort: it's a plain kfold you use here.

pyriemann/utils/tangentspace.py
```diff
@@ -50,3 +57,14 @@ def untangent_space(T, Cref):
         covmats[i] = numpy.dot(numpy.dot(C12, covmats[i]), C12)
 
     return covmats
+
+
+def transport(Covs, Cref, metric='riemann'):
+    """Parallel transport of two set of covariance matrix.
```
agramfort: Parallel transport of two sets of covariance matrices.

also you have no docstring :)

examples/ERP/plot_classify_MEG_mdm.py
alexandrebarachant: thats probably some older comment that was left from using the sufflesplit ... thanks for catching it.

PR #81, opened 2020-05-08
Adding an example for SSVEP

sylvchev:
This PR aims to extend the example gallery with an SSVEP example.
This is a joint work with Emmanuel Kalunga  (@emmanuelkalunga) and Quentin Barthélemy. We made an emphasis on how to build the extended covariance matrices.

### DIFF
--- a/examples/SSVEP/README.txt
+++ b/examples/SSVEP/README.txt
@@ -1,4 +1,4 @@
 Classification of SSVEP
----------------
+-----------------------
 
 Using Riemannian geometry for processing Steady-State Event Related Potentials.
--- a/examples/SSVEP/README.txt
+++ b/examples/SSVEP/README.txt
@@ -1,0 +1,4 @@
+Classification of SSVEP
+---------------
+
+Using Riemannian geometry for processing Steady-State Event Related Potentials.
--- a/examples/SSVEP/plot_classify_ssvep_mdm.py
+++ b/examples/SSVEP/plot_classify_ssvep_mdm.py
@@ -7,15 +7,14 @@ Building extended covariance matrices for SSVEP-based BCI. The
 obtained matrices are shown. A Minimum Distance to Mean classifier
 is trained to predict a 4-class problem for an offline setup.
 """
-# Authors: Sylvain Chevallier <sylvain.chevallier@uvsq.fr> ,
-# Emmanuel Kalunga , Quentin Barthélemy, David Ojeda
+# Authors: Sylvain Chevallier <sylvain.chevallier@uvsq.fr>,
+# Emmanuel Kalunga, Quentin Barthélemy, David Ojeda
 #
 # License: BSD (3-clause)
 
 # generic import
 import os
 import numpy as np
-from scipy.signal import butter, lfilter, filtfilt
 import matplotlib.pyplot as plt
 
 # mne import
@@ -76,7 +75,6 @@ def download_sample_data(dataset="ssvep", subject=1, session=1):
         _fetch_file(url, destination, print_destination=False)
     return destination
 
-
 # Download data
 destination = download_sample_data(dataset="ssvep", subject=12, session=1)
 # Read data in MNE Raw and numpy format
@@ -101,7 +99,7 @@ plt.figure(figsize=(10, 4))
 plt.plot(time.T, eeg_data[np.array(raw.ch_names) == 'Oz', :n_seconds*sfreq].T,
          color='C0', lw=0.5)
 plt.xlabel("Time (s)")
-_ = plt.ylabel(r"Oz ($\mu$V)")
+plt.ylabel(r"Oz ($\mu$V)")
 
 ###############################################################################
 # And of all electrodes:
@@ -123,18 +121,16 @@ raw.plot(duration=n_seconds, start=0, n_channels=8, scalings={'eeg': 4e-2},
 ###############################################################################
 # Extended signals for spatial covariance
 # ---------------------------------------
-# Using the approach proposed by [1]_, the SSVEP signal is extended to include
+# Using the approach proposed by [1], the SSVEP signal is extended to include
 # the filtered signals for each stimulation frequency. We stack the filtered
 # signals to build an extended signal
 
-
 def _bandpass_filter(signal, lowcut, highcut):
-    """ Bilateral bandpass filter for offline filtering """
+    """ Bandpass filter using MNE """
     return signal.copy().filter(l_freq=lowcut, h_freq=highcut,
                                 method="iir").get_data()
 
-
-# we stack the filtered signals to build an extended signal
+# We stack the filtered signals to build an extended signal
 frequencies = [13., 17., 21.]
 freq_band = 0.1
 ext_signal = np.vstack([_bandpass_filter(raw,
@@ -174,7 +170,7 @@ plt.legend(loc='upper right')
 ###############################################################################
 # As it can be seen on this example, the subject is watching the 13Hz
 # stimulation and the EEG activity is showing an increase activity in this
-# frequency band while other frequency have a lower amplitude.
+# frequency band while other frequencies have lower amplitudes.
 #
 # Spatial covariance for SSVEP
 # ----------------------------
@@ -190,8 +186,7 @@ plt.figure(figsize=(7, 7))
 for i, l in enumerate(event_id):
     ax = plt.subplot(2, 2, i+1)
     plt.imshow(cov_ext_trials[events[:, 2] == event_id[l]][0],
-               cmap=plt.get_cmap('RdBu_r'),
-               interpolation='nearest')
+               cmap=plt.get_cmap('RdBu_r'))
     plt.title('Cov for class: '+l)
     plt.xticks([])
     if i == 0 or i == 2:
@@ -204,10 +199,10 @@ for i, l in enumerate(event_id):
 # It appears clearly that each class yields a different structure of the
 # covariance matrix. Each stimulation (13, 17 and 21 Hz) generating higher
 # covariance values for EEG signal filtered at the proper bandwith and no
-# activation at all for the other bandwith. The resting state, where the
+# activation at all for the other bandwiths. The resting state, where the
 # subject focus on the center of the display and far from all blinking
 # stimulus, shows an activity with higher correlation in the 13Hz frequency
-# and lower but still visible activity in the other bandwith.
+# and lower but still visible activity in the other bandwiths.
 #
 # Classify with MDM
 # -----------------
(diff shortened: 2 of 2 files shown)


alexandrebarachant:
This is great. You need to add a README.txt in the folder if you want the doc to be generated. Otherwise looks good.

sylvchev:
Thank you! I added a README to add this script to the generated doc.

examples/SSVEP/README.txt
```diff
@@ -0,0 +1,4 @@
+Classification of SSVEP
+---------------
```
agramfort, line 2: line is to short
```suggestion
-----------------------
```
it should match the number of characters above otherwise it's an RST bug

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,239 @@
...
+ext_signal = np.empty_like(eeg_data[0, :])
+for f in frequencies:
+    ext_signal = np.vstack((ext_signal,
+                            _butter_bandpass(eeg_data,
+                                             lowcut=f-freq_band,
+                                             highcut=f+freq_band,
+                                             fs=sfreq)))
+ext_signal = ext_signal[1:, :]
+
+# and we build an array with the signal for each trial
```
dojeda, line 127: I think mne can do this by creating a `mne.Epochs` from your events, and you have both of these on mne already.
I always have trouble mne correctly so I can't propose a piece of code.

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,239 @@
...
+ext_trials = list()
+for t in events[:, 0]:
+    start = t + 2 * sfreq
+    stop = t + 5 * sfreq
+    ext_trials.append(ext_signal[:, start:stop])
+ext_trials = np.array(ext_trials)
+n, c, t = ext_trials.shape
```
dojeda, line 134: These are not used anywhere

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,239 @@
...
+    else:
+        plt.yticks([])
+
+###############################################################################
+# Minimum distance to mean is a simple and robust algorithm for BCI decoding
+
+cv = RepeatedKFold(n_splits=2, n_repeats=10, random_state=42)
+mdm = MDM(metric=dict(mean='riemann', distance='riemann'))
+scores = cross_val_score(mdm, cov_ext_trials, events[:, 2], cv=cv, n_jobs=1)
+print("MDM accuracy: {:.2f}% +/- {:.2f}".format(np.mean(scores)*100,
```
dojeda, line 229: I suggest you leave a comment with the expected output of this final accuracy... just as a reference on what it should have given if everything went well

examples/SSVEP/README.txt
sylvchev: Thank you, I thought a minimal number of dashes was enough!

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,239 @@
...
+eeg_data = raw.get_data()
+
+###############################################################################
+# Visualization of raw EEG data
+# -----------------------------
+#
+# Plot few seconds of signal from the `Oz` electrode using matplotlib
+
+n_seconds = 2
+time = np.linspace(0, n_seconds, n_seconds*sfreq).reshape((1, n_seconds*sfreq))
```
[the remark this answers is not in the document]
sylvchev: Brilliant! Thank you so much, I didn't catch that one.
Still, I prefer the linspace version (with the endpoint correction you suggested) to keep it simple (just `n_seconds` and `sfreq`).

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,239 @@
...
+# Extended signals for spatial covariance
+# ---------------------------------------
+# Using the approach proposed by [1]_, the SSVEP signal is extended to include
+# the filtered signals for each stimulation frequency. We define a function to
+# filter the signal:
+
+def _butter_bandpass(signal, lowcut, highcut, fs, order=4):
+    """ Bilateral Butterworth filter for offline filtering """
+    nyq = 0.5 * fs
+    low = lowcut / nyq
```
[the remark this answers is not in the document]
sylvchev: Thanks, it is much more readable like this!

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,239 @@
...
+frequencies = [13., 17., 21.]
+freq_band = 0.1
+ext_signal = np.empty_like(eeg_data[0, :])
+for f in frequencies:
+    ext_signal = np.vstack((ext_signal,
+                            _butter_bandpass(eeg_data,
+                                             lowcut=f-freq_band,
+                                             highcut=f+freq_band,
+                                             fs=sfreq)))
+ext_signal = ext_signal[1:, :]
```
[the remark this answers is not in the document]
sylvchev: Yes! Very nice one. I like the list comprehension version!

examples/SSVEP/plot_classify_ssvep_mdm.py
sylvchev: Not sure how to do this in a simple piece of code, but it could be possible to use tailored events. I couldn't figure a piece of code either.

examples/SSVEP/plot_classify_ssvep_mdm.py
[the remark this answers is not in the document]
dojeda: There is also the loopless one if you prefer 😉  
```python
ext_signal = np.apply_along_axis(_butter_bandpass, 1, eeg_data, lowcut=f-freq_band, highcut=f+freq_band)
```

sylvchev:
@dojeda Thank you, this is really nice. 
About your suggestion on the requirements, this is a good idea. Maybe we could use something like https://github.com/rasbt/watermark to indicate the libraries used (to have something dynamic) or write a frozen requirement file. 
Regarding the publication, this is more about a methodology than a specific result. IMO we already have a high impact regarding reproducibility with this example script.

examples/SSVEP/plot_classify_ssvep_mdm.py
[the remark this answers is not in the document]
sylvchev: This one is amazing! 
I think I'll keep the list comprehension one, as there is no computation issue and it is more readable IMO : it made clear that we vstack the same signal filtered in different frequency bands.

bertrandlalo:
@sylvchev, we could also implement this new covariance that you designed as a scikit estimator and then use is in your exemple to allow for easier cross validation :  
https://github.com/bertrandlalo/pyRiemann/blob/91da51bedcb5942586cba32fe17ed568ee571d7c/pyriemann/estimation.py#L600-L668
Or is it too specific to deserve a place in module `estimation.py`?

examples/SSVEP/plot_classify_ssvep_mdm.py
sylvchev: Yes, good idea.

sylvchev:
Thank you @bertrandlalo, this is a very good point. 
We wrote a scikit estimator for the SSVEP with Quentin Barthélemy, but I think this is outside the scope of this PR. We thought about opening a PR to add a `SSVEPCovariance` class in `estimation.py` but we decided against it. The main reason is that the filtering part is central and burden the class with details specific to acquisition (that are avoided in other classes). Also, it seems better to rely on MNE for this level of processing, as it is done in [MOABB](https://github.com/NeuroTechX/moabb/blob/master/examples/plot_cross_subject_ssvep.py).

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,234 @@
...
+key, key_dest = 'MNE_DATASETS_SSVEPEXO_PATH', 'MNE-ssvepexo-data'
+
+# Use MNE _fetch_file to download EEG file
+if get_config(key) is None:
+    set_config(key, os.path.join(os.path.expanduser("~"), "mne_data"))
+path = _get_path(None, key, sign)
+destination = _url_to_local_path(url, os.path.join(path, key_dest))
+os.makedirs(os.path.dirname(destination), exist_ok=True)
+if not os.path.exists(destination):
+    _fetch_file(url, destination, print_destination=False)
```
alexandrebarachant, line 56: I would suggest to make a method for this. We want to keep the example short and on point. If we happens to extend the set of example we might put it somewhere else to, but for now its fine to have it here in its own function.

examples/SSVEP/plot_classify_ssvep_mdm.py
```diff
@@ -0,0 +1,234 @@
...
+    """ Bilateral Butterworth filter for offline filtering """
+    b, a = butter(order, (lowcut, highcut), btype='bandpass', fs=fs)
+    filtered = filtfilt(b, a, signal, axis=-1)
+    return filtered
+
+
+# we stack the filtered signals to build an extended signal
+frequencies = [13., 17., 21.]
+freq_band = 0.1
+ext_signal = np.vstack([_butter_bandpass(eeg_data, fs=sfreq,
```
alexandrebarachant, line 116: could you use the MNE filter function instead ?

examples/SSVEP/plot_classify_ssvep_mdm.py
alexandrebarachant: i agree, lets try to use the mne.Epochs for this

examples/SSVEP/plot_classify_ssvep_mdm.py
sylvchev: Yes, this is a very good idea, it could be useful to have some functions to bridge pyriemann and moabb.

examples/SSVEP/plot_classify_ssvep_mdm.py
sylvchev: Okay! I forgot that I did not use mne.Epoch at all in this PR. I rewrote the code to use the epochs correctly.

examples/SSVEP/plot_classify_ssvep_mdm.py
sylvchev: Yes, you are right it is better than creating an ad hoc function. I updated the code to use MNE's filter.

examples/SSVEP/plot_classify_ssvep_mdm.py
sylvchev: Good idea! I added a small function, more generic to download a few samples. It could be used as a starting point for a correct helper function.

sylvchev:
Thanks for your feedback, @alexandrebarachant I updated code from your comment. Could you review this version?

qbarthelemy:
@alexandrebarachant , in which module can we put the function `download_sample_data`?

PR #83, opened 2020-05-10
fixed Xdawn filter parameter storing

### LINKED ISSUE #82: Cloned Xdawn fails
When I clone Xdawn filter and try to use it an error occurs. Minimal example to reproduce:

```python
import numpy as np
from sklearn.base import clone
from pyriemann.spatialfilters import Xdawn

xd = Xdawn()
cl_xd = clone(xd).set_params(estimator='oas')
cl_xd.fit(np.random.randn(2, 10, 10), [1, 0])

>>> Traceback (most recent call last):
>>>   File "<stdin>", line 1, in <module>
>>>   File "../lib/python3.7/site-packages/pyriemann/spatialfilters.py", line 94, in fit
>>>     Cx = numpy.matrix(self.estimator(tmp.reshape(Ne, Ns * Nt)))
>>> TypeError: 'str' object is not callable
```

I'll make a PR to fix it by recording only string to class attribute and getting actual function on demand instead of storing function as an attribute. This is common for sklearn-like transformers.

P.S. Actually I've encountered this error by using sklearn's cross_validate function where joblib couldn't clone this class instance due to this behaviour

v-goncharenko:
Fixes #82

### DIFF
--- a/pyriemann/spatialfilters.py
+++ b/pyriemann/spatialfilters.py
@@ -64,8 +64,12 @@ class Xdawn(BaseEstimator, TransformerMixin):
         """Init."""
         self.nfilter = nfilter
         self.classes = classes
-        self.estimator = _check_est(estimator)
+        self.estimator = estimator
         self.baseline_cov = baseline_cov
+
+    @property
+    def estimator_fn(self):
+        return _check_est(self.estimator)
 
     def fit(self, X, y):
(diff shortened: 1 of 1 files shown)


alexandrebarachant:
Thanks a lot for this. I'm surprised i didn't catch this before since it frequently use this method in cross validation.

PR #85, opened 2020-06-28
add classes_ to FgMDM

sylvchev:
The attribute `classes_` is required for ClassifierMixin, as indicated on this [scikit-learn page](https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator) on how to define your own estimator.

I also removed a dead code part.

### DIFF
--- a/pyriemann/classification.py
+++ b/pyriemann/classification.py
@@ -114,12 +114,6 @@ class MDM(BaseEstimator, ClassifierMixin, TransformerMixin):
             self.covmeans_ = [mean_covariance(X[y == l], metric=self.metric_mean,
                                     sample_weight=sample_weight[y == l])
                                         for l in self.classes_]
-            """
-            for l in self.classes_:
-                self.covmeans_.append(
-                    mean_covariance(X[y == l], metric=self.metric_mean,
-                                    sample_weight=sample_weight[y == l]))
-            """
         else:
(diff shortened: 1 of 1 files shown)


agramfort:
I am surprised you don’t reuse it in predict 

The idea of this attributes is that you can pass y as list of string and get string when you predict

sylvchev:
Yes, you're right, it is not reused in predict. FgMDM calls FGDA and MDM, so all the attributes are handled in those estimator. Predict is : 
```
cov = self._fgda.transform(X)
return self._mdm.predict(cov)
```

The main problem is that FgMDM should be able to expose this `classes_` attribute as it is required for ensemble learning, meta-classifier or, in some cases, model selection. Right now, it is exposed in MDM but not in FgMDM, hence the PR.

pyriemann/classification.py
```diff
@@ -281,6 +280,7 @@ def fit(self, X, y):
         self : FgMDM instance
             The FgMDM instance.
         """
+        self.classes_ = numpy.unique(y)
```
alexandrebarachant, line 283: it would be cleaner to just do `self.classes_ = self._mdm.classes_` after the fit. this way we are sure there isn't any discrepancy

agramfort:
thx @sylvchev 

I've tested locally.

PR #90, opened 2020-10-26
Add AJDC and an example on artifact correction

qbarthelemy:
This PR aims to:

- add functions `cross_spectrum` and `normalize_trace` in `utils.covariance`, fix `cospectrum` and complete tests,

- add weights in function `ajd_pham` in `utils.ajd`,

- add a class `AJDC` in `spatialfilters`, for SOS-based BSS and group BSS,

- add an example in the gallery, on artifact correction by AJDC, and a comparison with HOS-BSS (ICA-Infomax).


examples/artifacts/plot_correct_ajdc_EEG.py
```diff
@@ -0,0 +1,163 @@
...
+from mne.time_frequency import psd_welch
+from mne.preprocessing import ICA
+from pyriemann.spatialfilters import AJDC
+from matplotlib import pyplot as plt
+
+import warnings
+from mne import __version__ as v
+if v != '0.18.2':
+    warnings.warn('This example has only been tested with mne version 0.18.2',
+                  UserWarning)
```
agramfort, line 27: you can take this commit to avoid this:

https://github.com/agramfort/pyRiemann/commit/927459c0958feb95ddaf4e471ca49b6e1f21ca3c

qbarthelemy: Thank you Alexandre! I am going to upgrade mne and sklearn.

agramfort (approved):

examples/artifacts/plot_correct_ajdc_EEG.py
qbarthelemy: Upgrading mne from 0.18 to 0.21, it seems that channel positions have changed when displaying the topographic map of the blink source, using `plot_topomap()`. 

MNE 0.18
blink_topomap_mne018

MNE 0.21
blink_topomap_mne021


The difference is significant for frontal channels: Fp1, Fp2, F7, F3, F4, F8.
How can I fix that? @agramfort

agramfort:
ping @mmagnuski @cbrnr @larsoner

question of channel locations for EEG...

examples/artifacts/plot_correct_ajdc_EEG.py
cbrnr: I don't have enough background to tell what the problem could be. What exactly is different between MNE 0.18 and 0.21? The topomaps? I assume it's not just the visualization that's different (because (1) we changed the spherical projection at some point and (2) the sign is irrelevant). Which MNE functions does your method use that could have caused the result to change?

mmagnuski: @qbarthelemy Try using the `sphere` argument - the default sphere radius is `0.095` (in meters, so that's 9.5 cm) which may not be adequate for your channel positions. Take a look at the tutorial [here](https://mne.tools/dev/auto_tutorials/intro/plot_40_sensor_locations.html#controlling-channel-projection-mne-vs-eeglab) and [this example](https://mne.tools/dev/auto_examples/visualization/plot_eeglab_head_sphere.html#sphx-glr-auto-examples-visualization-plot-eeglab-head-sphere-py).

qbarthelemy: Thank you for your quick help!

@cbrnr Yes I agree: due to BSS uncertainty on sign and order of sources, the sign is not a problem. My question is about channel locations: in MNE 0.21, Fp1 and Fp2 are right next to the nose, and F7, F3, F4, F8 are curved downward. This is different from MNE 0.18.
I have updated my comment above, to highlight the differences between my example coded with MNE 0.18, and then with MNE 0.21.

 @mmagnuski Thanks for the links! I have read the tutorial several times, but I can't properly set the values of  `sphere` argument in MNE 0.21, because the documentation of MNE 0.18 does not state this information. So, I try to manually tune these values in order to reproduce the locations of MNE 0.18. But, I don't find (yet) the configuration ensuring backward compatibility with previous versions of MNE.

cbrnr: When you say you try to reproduce the original locations, are you measuring this by eye or do you have 3D coordinates for 0.18 and 0.21?

mmagnuski: @qbarthelemy The locations are actually the same, they are just differently projected to 2d with reference to a sphere. If you don't like how the channels are projected by default in 0.21 - now you have full control to specify the origin and radius of the sphere  used in the 3d -> 2d projection. To get exactly what you saw in 0.18 might take a bit of trial and error, but in general try moving the sphere origin to the front and reduce the sphere radius.

qbarthelemy: My approach was a simple try-and-error optimization, with a subjective visual evaluation. I have not reached the solution, in spite of several iterations. So, this "à la hussarde" strategy does not seem appropriate to converge.
It probably exists a closed-form solution to convert this 3D to 2D projection from MNE 0.18 to 0.21, requiring to dig into linear algebra.

But on reflection, this topomap is not the crucial part of my example, so, I have decided to keep the current display. Thanks for all!

mmagnuski: @qbarthelemy No problem. In your specific case it might indeed be impossible to get exactly the same projection as in 0.18. First, 0.18 had a different srategy for topo channel placement - first the channels were projected to a sphere and then they would be stretched until they filled the whole head circle. Now the channels are not projected to the sphere surface, only with respect to its position and radius. Then, because `'standard_1020'` used by mne currently contains realistic, not schematic, channel positions. Because these channels are not projected to the sphere surface now, their constellation may appear more eliptical - but that is due to the shape of the standard_1020 positions used by mne. This may be resolved one day though: mne-tools/mne-python#7141 :)

pyriemann/utils/ajd.py
```diff
@@ -148,11 +167,11 @@ def ajd_pham(X, eps=1e-6, n_iter_max=15):
                 c1 = A[ii, Ii]
                 c2 = A[jj, Ij]
 
-                g12 = np.mean(A[ii, Ij] / c1)
-                g21 = np.mean(A[ii, Ij] / c2)
+                g12 = np.average(A[ii, Ij] / c1, weights=sample_weight)
```
qbarthelemy, line 170: @gouypailler : can you confirm the processing of weights in ajd_pham?

qbarthelemy:
Dev finished. 
Tests OK. 
CodeCov OK.
Documentation with sphinx OK.
@alexandrebarachant

agramfort:
tests don't run anymore here... need to move to GitHub actions...

pyriemann/spatialfilters.py
```diff
@@ -562,24 +576,40 @@ def fit(self, X, y=None):
...
+                    raise ValueError('Unequal number of conditions')
+                if self.n_channels_ != cosp_.shape[1]:
+                    raise ValueError('Unequal number of channels')
+            cosp.append(cosp_)
+        cosp = numpy.transpose(numpy.array(cosp), axes=(0, 1, 4, 2, 3))
+
+        # trace-normalization of cospectra, Eq(3) in [2]
+        cosp = self._normalize_trace(cosp)
+        # average of cospectra across subjects, Eq(7) in [2]
+        cosp = numpy.mean(cosp, axis=0, keepdims=False)
```
qbarthelemy, line 605: Is there a study showing the improvement of gBSS when using a Riemannian mean across subjects? @Marco-Congedo

Marco-Congedo: I am not aware of such a study. Cheers

pyriemann/spatialfilters.py
```diff
@@ -462,3 +464,312 @@ def fit(self, X, y):
...
+        """Fit.
+
+        Compute and diagonalize cospectra, to estimate forward and backward
+        spatial filters.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_subjects, n_conditions, n_channels, n_samples) | list of n_subjects of list of n_conditions ndarray of shape (n_channels, n_samples), with same n_conditions and n_channels but different n_samples
+            ndarray of signal in channel space, acquired for different subjects
+            and under different experimental conditions.
```
lkorczowski, line 570: fit and transform have different input. 
`(n_trials, n_channels, n_samples) # in transform`

should it be better to have also `X shape (n_trials, n_channels, n_samples)` and use `y` to find the conditions ?

lkorczowski (approved):
Great. Thanks @qbarthelemy !

I see only minor improvements. The biggest issue is that `fit` and `transform` are not used the same input while it could be possible by using `y`.

pyriemann/utils/ajd.py
```diff
@@ -2,6 +2,23 @@
 import numpy as np
 
 
+def _get_normalized_weight(sample_weight, data):
```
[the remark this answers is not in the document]
qbarthelemy: In fact, I have mimicked the function `utils.mean._get_sample_weight(sample_weight, data)`, where the argument `sample_weight ` has no default value.

pyriemann/utils/ajd.py
```diff
@@ -148,11 +168,11 @@ def ajd_pham(X, eps=1e-6, n_iter_max=15):
...
-                g12 = np.mean(A[ii, Ij] / c1)
-                g21 = np.mean(A[ii, Ij] / c2)
+                g12 = np.average(A[ii, Ij] / c1, weights=normalized_weight)
+                g21 = np.average(A[ii, Ij] / c2, weights=normalized_weight)
 
-                omega21 = np.mean(c1 / c2)
-                omega12 = np.mean(c2 / c1)
+                omega21 = np.average(c1 / c2, weights=normalized_weight)
+                omega12 = np.average(c2 / c1, weights=normalized_weight)
                 omega = np.sqrt(omega12 * omega21)
```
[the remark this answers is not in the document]
qbarthelemy: Yes, figures look similar.

But I tried to check numerically the equivalence with the following code, but the difference is too high.
```
n_repeats, n_trials, n_channels = 100, 100, 5
n_iter_max = 50

mse = np.zeros((n_repeats))
rmse = np.zeros((n_repeats))
for r in range(n_repeats):
    covs, _, _ = generate_cov(n_trials, n_channels)
    w = _get_normalized_weight(np.abs(np.random.randn(n_trials)), covs)
    
    V1, D1 = ajd_pham(covs, sample_weight=w, n_iter_max=n_iter_max)
    V2, D2 = ajd_pham(w[:, np.newaxis, np.newaxis] * covs, n_iter_max=n_iter_max)
    
    mse[r] = np.linalg.norm(V1 - V2, 'fro')**2
    rmse[r] = mse[r] / np.linalg.norm(V1, 'fro')**2

print("MSE : {:.2f} +/- {:.2f}".format(np.mean(mse), np.std(mse)))
print("rMSE : {:.2f}% +/- {:.2f}".format(np.mean(rmse)*100, np.std(rmse)*100))
```

I think that, if we apply the pre-multiplication trick, weights are used only at the first iteration, and then, they are lost during the update of matrix A, from https://github.com/alexandrebarachant/pyRiemann/blob/master/pyriemann/utils/ajd.py#L171 to https://github.com/alexandrebarachant/pyRiemann/blob/master/pyriemann/utils/ajd.py#L178.

pyriemann/spatialfilters.py
qbarthelemy: Thanks, it's a good idea! 

But, what could be the optimal shape and content of `y`? 

It could be a ndarray of shape `(n_subjects, n_conditions, 3)`, where the final dimension contains 3 values: 1 condition index, and 2 sample indices, defining the beginning and the end of the condition.

pyriemann/utils/ajd.py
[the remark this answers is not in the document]
Marco-Congedo: Hi guys, using the log-det criterion (Pham's), weights must be used explicitly within the algorithm, not multiplied to the input matrices.

pyriemann/spatialfilters.py
lkorczowski: More I think about this and more it seems not trivial (so not urgent).

The only default I see have having X different in fit/transform is that it can be confusing for the user but the proposal bellow is even more confusing. So keep it with a grain of salt just for discussing the design, and your solution seems good enough.


- if `X` is `(n_trials, n_channels, n_samples)` 3D epochs. Note that most of the time it will be 2D as users just want to find the sources of the full EEG signal, i.e.  `(n_channels, n_samples)`.
- then `y`, taking inspiration of the sklearn API would be either (n_trials,) with n_classes unique labels or (n_trials, n_classes).

So we would have the following options for `y`
```
- None  # (no condition, no specific subject) : simple BSS or gBSS scheme of AJD where `n_trials` cross-spectral matrices are averaged. So user would have to unsure herself conditions and subjects balance, i.e. giving the same number of trial for a given subject or condition. Easy to go. Should work for most situation.
- ndarray, shape (n_trials,)  # supervised training n_condition*n_subjects unique labels for AJD of cross-spectral matrices. SCM are averaged for each unique label (condition or subject or subject+condition)*
- ndarray, shape (n_trials, n_classes)  # same than before but where n_classes =  n_condition*n_subjects and boolean instead
```

*e.g. if you want to average subjects for gBSS, you could add an input "labels_to_average" which can be a list or list of list, e.g.
```
labels_to_average :  list | list of list (default: None)
    List of unique labels in `y` which will be averaged together.
    By default (None), each unique labels has one cross-spectrum matrix estimation.
    Example for group Blind Source Separation (gBSS) with uniform cohort:
    >>> labels_to_average = ['subject_1', 'subject_2', ...., 'subject_M']
    Example for group Blind Source Separation (gBSS) with non-uniform cohort:
    >>> labels_to_average = [['subject_1', 'subject_2'],
    >>>                                   ['subject_3', 'subject_4']]  # subjects 1&2 belong to different cohort than 3&4

    Example for grouping conditions together
    >>> labels_to_average = [['condition_1', 'condition_2'],
    >>>                                    ['condition_3']]  # two first conditions are merged together

``` 

I should ask @Marco-Congedo because IMO, averaging CSM for all subjects may not be the only use of gBSS as sources similarities may be just as important as sources dissimilarities. Therefore I wonder if BSS could use be used for different cohorts or a completely blind scheme where each subject has his own Cross-Spectral Matrix.

qbarthelemy: For `transform` and `inverse_transform` functions, `trials` can relate to any batch of epochs, coming from different subjects, sessions, conditions, trials, etc.

For the `fit` function, the name `trials` seems confusing: we don't know if this quantity relates to `subjects` or `conditions`. It is a crucial problem because these dimensions are not interchangeable in `AJDC`. They are not processed in the same way: cospectra are averaged across subjects but concatenated along conditions. So, the shape of `y` must make it possible to contain these two dimensions and distinguish them. Adding an argument `labels_to_average` breaks the compatibility with `sklearn` API.

Unfortunately, we can't mimic `sklearn` or `mne` APIs, which provide functions for BSS (notably ICA) but not for group BSS (pyRiemann is going to fast!).

**Suggestion for `fit` arguments**
```
X : ndarray, shape (n_trials, n_channels, n_samples)
    ndarray of trials in channel space
y : list of strings of shape (n_trials,) | None
    List of labels, characterizing each trial with respect its subject and/or condition (semicolon separated)
```

Example of `y` for:
BSS under a single condition : `None`
BSS under several conditions : `['condition_baseline', 'condition1', 'condition_2']`
gBSS under a single condition : `['subject1', ..., 'subject_17', ..., 'subject-42bis']`
 gBSS under several conditions : `['subject1;condition1', 'subject_2;condition1', 'subject3;condition-2', 'subject3;condition1', 'subject1;condition-2', 'subject_2;condition-2']`

With this specification, `y` could be decoded in two steps:
- using regex on `subject` and `condition` expressions, to know what the label refers to,
- and then counting the number of different labels for each of the two expressions.

Then, we could reshape trials X with respect found labels. Then AJDC could begin...
It makes the job, but it still seems cumbersome compared to the current implementation.

pyriemann/utils/ajd.py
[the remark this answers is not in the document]
qbarthelemy: Thank you Marco! I keep the current implementation, coming from your Pascal code.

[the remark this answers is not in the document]
Marco-Congedo: Ah! Ca risque d'être peu performant en Python. Ca doit être implémenté dans MNE, non?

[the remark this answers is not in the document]
qbarthelemy: For `ajd_pham`, MNE uses a copy of pyRiemann (cf https://github.com/mne-tools/mne-python/blob/master/mne/decoding/csp.py#L618).

In this PR, I have just updated the Python code of pyRiemann, adding weights multiplication to the appropriate variables, replacing `np.mean(x)` by `np.average(x, weights=weights)`. So, I think that the new implementation is still optimal.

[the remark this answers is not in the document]
Marco-Congedo: Right! Code replication is something we don't see much in Julia.

pyriemann/spatialfilters.py
lkorczowski: indeed. Let's keep your implementation.

pyriemann/clustering.py
```diff
@@ -2,7 +2,11 @@
 import numpy
 from sklearn.base import (BaseEstimator, ClassifierMixin, TransformerMixin,
                           ClusterMixin)
-from sklearn.cluster._kmeans import _init_centroids
+try:                # pragma: no cover
+    from sklearn.cluster._kmeans import _init_centroids
+except ImportError: # pragma: no cover
+    from sklearn.cluster import KMeans
+    _init_centroids = KMeans._init_centroids
```
ErikBjare, line 9: This won't work, see: https://github.com/alexandrebarachant/pyRiemann/issues/92#issuecomment-740641197

ErikBjare (changes requested):

qbarthelemy:
Obviously, I have missed one of the multiple rebases of the branch, so this PR is broken (the regular storry of non merged PR).

I will try to push AJDC from a new fork.

PR #99, opened 2021-03-29
fix for new sklearn version

gabelstein:
sklearn 0.24 changed the KMeans class, making the _init_centroinds method a class method.


gabelstein:
Whoops, hadn't noticed this has already been done.

PR #100, opened 2021-04-15
Correct probas of MDM

qbarthelemy:
When a Riemannian Gaussian distribution is used to model each class k=1... K, the probability density function is [Said2017](https://ieeexplore.ieee.org/abstract/document/7819552):
mdm_likelihood
where mdm_bar_Sigma is the center of the Riemannian Gaussian distribution of class k, mdm_sigma encodes the dispersion of the distribution, and mdm_ksi is a normalization factor.

MDM assumes that: (i) all classes have identical dispersions mdm_sigma, and (ii) all classes are likelly (equal priors). 

Under these assumptions, Bayes' law of total probability gives the posterior:
mdm_proba
using squared distances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,11 @@ The package aims at adopting the [Scikit-Learn](http://scikit-learn.org/stable/d
 
 # changelog
 
+### v0.2.X
+- Add example on SSVEP classification
+- Fix compatibility with scikit-learn v0.24
+- Correct probas of MDM
+
 ### v0.2.6
 - Updated for scikit-learn v0.22
 - Remove support for python 2.7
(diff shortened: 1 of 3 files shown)


agramfort:
LGTM

we should have a release notes pages here to document bug fixes and API changes and enhancements. See example:

https://mne.tools/stable/whats_new.html#whats-new

it also gives public credit to contributors for their work.

agramfort:
@qbarthelemy can you add a line in https://github.com/pyRiemann/pyRiemann/blob/master/doc/whatsnew.rst 
to document the bug fix?

agramfort:
thanks @qbarthelemy

PR #106, opened 2021-05-08
add action to build doc and fix doc build

agramfort:


agramfort:
looks good https://github.com/pyRiemann/pyRiemann/suites/2685650347/artifacts/59206569 !

PR #112, opened 2021-05-20
Add weights to Pham's AJD algorithm

qbarthelemy:
Add a weight for each matrix into Pham's AJD algorithm. Some discussions about implementation can be found in #90.

### DIFF
--- a/tests/test_ajd.py
+++ b/tests/test_ajd.py
@@ -1,7 +1,8 @@
-from numpy.testing import assert_array_almost_equal, assert_array_equal
+from numpy.testing import assert_array_equal
 import numpy as np
+import pytest
 
-from pyriemann.utils.ajd import rjd, ajd_pham, uwedge
+from pyriemann.utils.ajd import rjd, ajd_pham, uwedge, _get_normalized_weight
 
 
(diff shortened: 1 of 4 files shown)


tests/test_ajd.py
```diff
@@ -24,9 +39,15 @@ def test_rjd():
...
     """Test pham's ajd"""
-    covmats, diags, A = generate_cov(100, 3)
+    Nt = 100
+    covmats, diags, A = generate_cov(Nt, 3)
     V, D = ajd_pham(covmats)
 
+    w = 5 * np.ones(Nt)
+    Vw, Dw = ajd_pham(covmats, sample_weight=w)
+    assert_array_equal(V, Vw)
+    assert_array_equal(D, Dw)
```
agramfort, line 49: what I would test is that if the weight of one covmat is 0 it's like it was not passed. It's what sklearn does for testing.

agramfort (commented):
thx @qbarthelemy 

can you also update what's new page?

thx

agramfort (approved):

agramfort:
thx @qbarthelemy

Issue #116, opened 2021-05-24
Incorrect cospectrum


qbarthelemy:
**Problems**

Function `cospectrum` had been modified here https://github.com/pyRiemann/pyRiemann/commit/9a6dbeef03f61c7658c8902923d0f22d84b57c1c#diff-b256f1ac840adebcb1b8bd17d18f050b62b1b6b7f9ef86331083bfbccd0d16d6
but the output is defined as the squared modulus of the cross-spectra, instead of the real part, as mentionned in #90.

Moreover, `cospectrum(x, fmin=20)` raises the `TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'`
at line https://github.com/pyRiemann/pyRiemann/commit/9a6dbeef03f61c7658c8902923d0f22d84b57c1c#diff-b256f1ac840adebcb1b8bd17d18f050b62b1b6b7f9ef86331083bfbccd0d16d6R136 : `f = numpy.arange(0, 1, 1.0 / number_freqs) * (fs / 2.0)`, because `fs` is None.

**Cross-spectra and co-spectra**

![Issue_CoSpectrum](https://user-images.githubusercontent.com/5346669/119399642-6d250580-bcd9-11eb-906f-8cf46312dc56.png)

**Solution**

PR in progress...

alexandrebarachant:
this is intended to be the squared modulus (cross-spectrum). it could be great to have the option to choose what you want if you make a PR.

qbarthelemy:
If the output of the function is the (squared) modulus, the phase is definitively lost.

It is better to return the raw complex cross-spectrum, and then users can compute the feature they need: real part (for co-spectrum), imaginary part, modulus (useful for coherence) or phase.

alexandrebarachant:
yeah, phase is lost. its intented to be a "unnormalized" version of the coherence (i.e. what covariance is to correlation). I would say the best option is to add a parameter that allows you to switch between the different version (you could even throw in normalization to have coherence estimation). This way it won't break existing pipeline that expect non-complex outputs.

qbarthelemy:
Closed by #118.

PR #118, opened 2021-05-27
Add cross_spectrum and fix cospectrum

qbarthelemy:
This PR solves #116:
- adding a function `cross_spectrum` in `utils.covariance` to compute cross-spectral matrices,
- adding tests for `cross_spectrum`,
- defining `cospectrum` (and `CospCovariances`) as the real part of cross-spectrum,
- adding frequencies in attributes of `CospCovariances`, and completing its documentation,
- modifying `coherence`, to return the same output as before.

Related to #90.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -125,6 +125,7 @@ Covariance preprocessing
     :toctree: generated/
 
     covariances
+	cross_spectrum
     normalize
 
 
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -125,7 +125,9 @@ Covariance preprocessing
     :toctree: generated/
 
     covariances
-	cross_spectrum
+    cross_spectrum
+    cospectrum
+    coherence
     normalize
 
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -20,7 +20,7 @@ v0.2.7.dev
 
 - Add weights to Pham's AJD algorithm
 
-- Add cross_spectrum, fix cospectrum; coherence output is kept unchanged
+- Add :func:`pyriemann.utils.covariance.cross_spectrum`, fix :func:`pyriemann.utils.covariance.cospectrum`; :func:`pyriemann.utils.covariance.coherence` output is kept unchanged
 
 v0.2.6 (March 2020)
 -------------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -20,6 +20,8 @@ v0.2.7.dev
 
 - Add weights to Pham's AJD algorithm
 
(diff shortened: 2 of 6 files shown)


doc/api.rst
```diff
@@ -125,6 +125,7 @@ Covariance preprocessing
     :toctree: generated/
 
     covariances
+	cross_spectrum
```
agramfort, line 128:
```suggestion
    cross_spectrum
```

pyriemann/utils/covariance.py
```diff
@@ -151,21 +180,47 @@ def cospectrum(X, window=128, overlap=0.75, fmin=None, fmax=None, fs=None):
...
+            raise ValueError('Parameter fmax must be inferior to fs/2')
         f = numpy.arange(0, 1, 1.0 / number_freqs) * (fs / 2.0)
         Fix = (f >= fmin) & (f <= fmax)
         fdata = fdata[:, :, Fix]
+        freqs = f[Fix]
+    else:
+        if fmin is not None:
+            raise Warning('Parameter fmin not used because fs is None')
+        if fmax is not None:
+            raise Warning('Parameter fmax not used because fs is None')
```
agramfort, line 201: the standard way is to do:

```python
import warnings
warnings.warn("Parameter fmax not used because fs is None")
```

pyriemann/utils/covariance.py
```diff
@@ -151,21 +180,47 @@ def cospectrum(X, window=128, overlap=0.75, fmin=None, fmax=None, fs=None):
...
-            number_windows * normval)
+        S[:, :, i] = numpy.dot(fdata[:, :, i].conj().T, fdata[:, :, i])
+    S /= number_windows * numpy.linalg.norm(win)**2
+
+    return S, freqs
 
-    return numpy.abs(S)**2
+
+def cospectrum(X, window=128, overlap=0.75, fmin=None, fmax=None, fs=None):
+    """Compute co-spectral matrices, the real part of cross-spectra."""
```
agramfort, line 214: public functions should have docstirngs

pyriemann/utils/covariance.py
agramfort: unless you want to reallly raise an Exception in which case I would suggest a ValueError

doc/whatsnew.rst
```diff
@@ -20,6 +20,8 @@ v0.2.7.dev
 
 - Add weights to Pham's AJD algorithm
 
+- Add cross_spectrum, fix cospectrum; coherence output is kept unchanged
```
agramfort, line 23: we should make the function and class references hyperlinks in what's new...

agramfort (approved):

agramfort:
thx @qbarthelemy

Issue #121, opened 2021-06-08
PyPi


mesca:
Any hope for a new PyPi package soon? 
The current version is broken and we need to add PyRiemann as a dependency to several PIP packages. Unfortunately, because of [this issue](https://github.com/pypa/warehouse/issues/7136), we cannot directly link to the git repository.
Currently, we prompt users to install PyRiemann separately, but this is far from ideal.

lkorczowski:
FYI I had to use `pip install --upgrade --force-reinstall git+https://github.com/pyRiemann/pyRiemann` because I inadvertently did `pip install pyriemann`. I think the version in setup  needs to be updated
https://github.com/pyRiemann/pyRiemann/blob/acd7d34f4f8d97ebe7e7d0f164332d56511e54c4/setup.py#L4

qbarthelemy:
Yep, it should be '0.2.7.dev'.

agramfort:
shall we release after the Whitening PR?

>

mesca:
> shall we release after the Whitening PR?

I guess so.

sylvchev:
We could release the 0.27 !

agramfort:
0.2.7 has been on pypi for a few days now

https://pypi.org/project/pyriemann/

>

PR #120, opened 2021-06-03
Add AJDC and an example on artifact correction, 2nd try

qbarthelemy:
This PR:

- adds a class `AJDC` in `spatialfilters`, useful for SOS-based BSS and group BSS,

- adds an example in the gallery, on EEG artifact correction by AJDC, and a comparison with HOS-BSS (ICA-Infomax).

Discussions about implementation can be found in #90.

PR #122, opened 2021-06-08
Add Whitening

qbarthelemy:
On second thoughts, the whitening and dimension reduction applied in `AJDC.fit()`, and added in #120, can be useful for many other covariance-based pipelines. @ygerf @sylvchev

This PR:

- adds a new class `Whitening` in `spatialfilters`, with an optional dimension reduction (3 different ways), with covariance matrices as inputs;

- updates `AJDC` to use the new whitening, and replaces parameter `expl_var` by `dim_red`: this replacement is not backward-compatible, but it should not be a problem because it was merged yesterday, and not published in a release (so, this PR should be merged before #121).


pyriemann/spatialfilters.py
```diff
@@ -1,17 +1,203 @@
...
+            elif self._dr_key == 'max_cond':
+                if self._dr_val <= 1:
+                    raise ValueError(
+                        'Value max_cond must be strictly superior to 1 '
+                        '(Got %d)' % self._dr_val)
+            else:
+                raise ValueError(
+                    'Unknown key in parameter dim_red: %r' % self._dr_key)
+        else:
+            raise ValueError('Unknown type for parameter dim_red')
```
agramfort, line 102: all this parameter validation should be done in fit and not in the init. This violates the scikit-learn API.
The init should not add any other attribute that the params passed to the init.
If someone uses set_params from BaseEstimator after the init you will have a problem otherwise.

pyriemann/spatialfilters.py
```diff
@@ -549,17 +736,15 @@ class AJDC(BaseEstimator, TransformerMixin):
     """
 
     def __init__(self, window=128, overlap=0.5, fmin=None, fmax=None, fs=None,
-                 expl_var=0.999, verbose=True):
+                 dim_red={'max_cond': 100}, verbose=True):
```
agramfort, line 739: I don't like this. To me you should be using a scikit-learn pipeline and not do any preprocessing in the init. It also makes the API more complex with dict which should not be a default parameter as it's a mutable object.

agramfort (commented):
would it make sense to move Whitening in a preprocessing submodule rather than in spatialfilters ?

pyriemann/spatialfilters.py
qbarthelemy: You are right: I will test parameters in `fit` rather than in `_init_`.

pyriemann/spatialfilters.py
qbarthelemy: What do you suggest? Defining `None` as default value for `dim_red`?

agramfort: I would not offer any dim red option in AJDC and tell users to use a pipeline with one step that does the dim red

qbarthelemy: I don't see how to use a `sklearn` pipeline inside `AJDC`.
Because dimension reduction is an inner step of `AJDC.fit()`, applied between cospectra estimation and ajd_pham.

Can you point an example using the suggested architecture?

agramfort:
see eg how we use make_pipeline in :

https://pyriemann.readthedocs.io/en/latest/auto_examples/ERP/plot_classify_MEG_mdm.html

qbarthelemy:
Yes, I know sklearn pipelines. But, once again, cospectra dimension reduction is an inner step inside AJDC.fit(), applied between cospectra estimation and Pham's AJD. This nested sub-step is not at the same level as AJDC, and consequently, it can't be pipelined before or after fitting AJDC.

agramfort:
would it make sense to have a pipeline with cospectra estimation >
whitening > AJDC?

if not you need to explain to me on a call it would be easier

>

agramfort (commented):
I'll try to push directly to this branch @qbarthelemy

agramfort:
Great team work @qbarthelemy 💪

qbarthelemy:
Yep! 

Just a minor mistake: new classes `AJDC` and `Whitening` are declared in version `0.2.8`, but it should be `0.2.7` (the actual next release).

agramfort:
good catch

fixed in master

>

PR #128, opened 2021-06-27
add not end-of life Python 3 versions

stonebig:


doc/installing.rst
```diff
@@ -16,7 +16,7 @@ The install script will install the required dependencies.
 Dependencies
 ~~~~~~~~~~~~
 
--  Python 2.7
+-  Python 2.7  or Python 3.6+
```
agramfort, line 19:
```suggestion
-  Python (>= 3.6)
```

although we test only on 3.7 now so I am not actually sure 3.6 is supported ... I would need to add a CI test for this...

maybe we just say 3.7?

can you also add in setup.py

`python_requires=">=3.7",`

@qbarthelemy wdyt?

agramfort (approved):

agramfort:
thx @stonebig 

it turns out it works fine with 3.6

https://github.com/pyRiemann/pyRiemann/pull/129

PR #135, opened 2021-08-16
Make matplotlib and seaborn optional dependencies

### LINKED ISSUE #134: Matplotlib dependency
There is a direct matplotlib dependency in pyriemann/stats.py:4

This dependency is implicit in [requirements.txt](https://github.com/pyRiemann/pyRiemann/blob/master/requirements.txt#L6) as seaborn is indicated, but [setup.py](https://github.com/pyRiemann/pyRiemann/blob/master/setup.py#L36) do not mention this.

This cause pytest to fail on a clean install.

sylvchev:
Closes #134 

Matplotlib and seaborn are listed in requirements.txt but not in setup.py, while they are needed for test. 

This PR makes import of matplotlib in pyriemann/stats.py optional. For pyriemann/viz.py, the tests are skipped if matplotlib or seaborn are not installed. 

The setup file now includes an "extras_require" section, to allow installation of doc requirements with either `pip install -r docs/requirements.txt` or with `pip install -e .[docs]`. Extra requirements for tests could be installed with `pip install -e .[tests]` or `pip install -e .[tests,docs]`.

I started to refactor the tests, using fixiture to unify the generation of covariance matrices in all tests.


tests/test_viz.py
```diff
@@ -1,35 +1,34 @@
...
-from pyriemann.utils.viz import (plot_confusion_matrix, plot_embedding,
-                                 plot_cospectra)
+from conftest import covmats
 
+try:
+    from pyriemann.utils.viz import (
+        plot_confusion_matrix,
+        plot_embedding,
+        plot_cospectra,
+    )
```
agramfort, line 10: this is not the right fix for me.

You should import plt inside the plot functions. Not at the top of the file at:

https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/utils/viz.py#L8

then you should decorate each test with something that skips the test if matplotlib is not installed

sylvchev: Ok, I have done it like that for stats.py. For the viz.py, I was wondering if it makes sense as all function are for generating plots.

sylvchev:
In viz.py there is a `plot_confusion_matrix` that is redundant with the sklearn.metrics.confusion_matrix and ConfusionMatrixDisplay.
@agramfort do you think I could deprecate this function and update the plot example where it is used (to use sklearn version)?

pyriemann/utils/viz.py
```diff
@@ -1,44 +1,57 @@
...
-import seaborn as sns
 
 
-def plot_confusion_matrix(targets, predictions, target_names,
-                          title='Confusion matrix', cmap="Blues"):
-    """Plot Confusion Matrix."""
+def plot_confusion_matrix(
+    targets, predictions, target_names, title="Confusion matrix", cmap="Blues"
+):
+    """Plot Confusion Matrix. **Deprecated**
```
qbarthelemy, line 11: You should complete whatsnew.rst .

tests/test_viz.py
```diff
@@ -1,35 +1,34 @@
...
-    A /= np.atleast_2d(np.sqrt(np.sum(A**2, 1))).T
-    covmats = np.empty((Nt, Ne, Ne))
-    for i in range(Nt):
-        covmats[i] = np.dot(np.dot(A, np.diag(diags[i])), A.T)
-    return covmats, diags, A
+plot = importlib.util.find_spec("matplotlib")
 
 
-def test_embedding():
+@pytest.mark.skipif(plot is None, reason="Matplotlib not installed")
```
agramfort, line 15: I would suggest to use such a function:

https://github.com/mne-tools/mne-python/blob/7636a58a744bb95b59816144cce7d01746fe6db7/mne/utils/_testing.py#L87

agramfort (commented):
just a suggestion but ok to merge as is.

@sylvchev no worries but I would avoid black formatting in PRs that do real changes.

if we switch to black we would do it in a PR that does just this.

pyriemann/utils/viz.py
sylvchev: done !

sylvchev:
Sorry for the black formatting, I'll avoid to change style in improvement specific PR.

pyriemann/utils/viz.py
```diff
@@ -1,44 +1,57 @@
...
-
 from sklearn.metrics import confusion_matrix
 from pyriemann.embedding import Embedding
-import matplotlib.pyplot as plt
 
 
-def plot_confusion_matrix(targets, predictions, target_names,
-                          title='Confusion matrix', cmap="Blues"):
-    """Plot Confusion Matrix."""
+def plot_confusion_matrix(
```
agramfort, line 8: when you deprecate a function you should say in which version is will be removed.

it should also warn with a DeprecationWarning when the function is called.

You can borrow the deprecated decorator in mne or sklearn

🙏

tests/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    covmats = np.empty((n_trials, n_channels, n_channels))
+    for i in range(n_trials):
+        covmats[i] = A @ np.diag(diags[i]) @ A.T
+    return covmats, diags, A
+
+
+@pytest.fixture
+def covmats():
+    """Generate covariance matrices for test"""
+    covmats, _, _ = generate_cov(6, 3)
```
qbarthelemy, line 40:
```suggestion
def covmats(n_trials, n_channels):
    """Generate covariance matrices for test"""
    covmats, _, _ = generate_cov(n_trials, n_channels)
```

qbarthelemy: With the parameter `n_trials` in `covmats()`, function `many_covmats()` will be useless.

sylvchev: That is what I tried at first, but fixitures are made to pass the same object for all tests : https://docs.pytest.org/en/latest/how-to/fixtures.html
The idea is that we generate a some covmats and these covmats are teared down in each test. Main point here is to avoid duplication of the covmat code across tests.

setup.py
```diff
@@ -34,5 +34,7 @@
       platforms='any',
       python_requires=">=3.6",
       install_requires=['numpy', 'scipy', 'scikit-learn',  'joblib', 'pandas'],
+      extras_require={'docs': ['sphinx-gallery', 'sphinx-bootstrap_theme', 'numpydoc', 'mne', 'seaborn'],
```
qbarthelemy, line 37: You should update the file `doc/installing.rst`

qbarthelemy:
No other remark on this PR for me.

agramfort (approved):

agramfort:
Thx @sylvchev !

agramfort:
Next time @sylvchev try to open PRs from your fork 🙏

PR #144, opened 2021-10-23
Add example on ERP

qbarthelemy:
This PR:
- adds `plot_erp` and complete test;
- add an example on ERP display, especially with 2D histogram.

It also renames the ERP example on embedding, because the script selects MEG channels instead of EEG ones. @plcrodrigues

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -30,6 +30,8 @@ v0.2.8.dev
 
 - Add new function :func:`pyriemann.datasets.make_gaussian_blobs` for generating random datasets with SPD matrices
 
+- Add an example on ERP visualization
+
 v0.2.7 (June 2021)
 ------------------
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -30,7 +30,7 @@ v0.2.8.dev
 
 - Add new function :func:`pyriemann.datasets.make_gaussian_blobs` for generating random datasets with SPD matrices
 
-- Add an example on ERP visualization
+- Add module ``pyriemann.utils.viz`` in API, add :func:`pyriemann.utils.viz.plot_waveforms`, and add an example on ERP visualization
 
 v0.2.7 (June 2021)
 ------------------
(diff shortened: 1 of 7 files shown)


pyriemann/utils/viz.py
```diff
@@ -100,3 +107,83 @@ def plot_cospectra(cosp, freqs, ylabels=None, title="Cospectra"):
             plt.yticks([])
 
     return fig
+
+
+def plot_waveforms(X, display, *, time=None, **kwargs):
```
agramfort, line 112: kwargs are not documented

tests/test_viz.py
```diff
@@ -36,3 +37,15 @@ def test_cospectra():
...
+@requires_matplotlib
+@pytest.mark.parametrize("display", ["all", "mean", "mean+/-std", "hist"])
+def test_plot_waveforms(display):
+    """Test plot_waveforms"""
+    n_matrices, n_channels, n_times = 16, 3, 50
+    X = np.random.randn(n_matrices, n_channels, n_times)
+    plot_waveforms(X, display)
+
+    X = np.random.randn(n_matrices, 1, n_times)
+    plot_waveforms(X, display)
```
agramfort, line 51: passing time param is not tested. I would also call it "times" like in mne.

doc/whatsnew.rst
```diff
@@ -30,6 +30,8 @@ v0.2.8.dev
 
 - Add new function :func:`pyriemann.datasets.make_gaussian_blobs` for generating random datasets with SPD matrices
 
+- Add an example on ERP visualization
```
agramfort, line 33: you also added a public function that should be documented in the public API page.

qbarthelemy:
@sylvchev , any idea on documentation build failed?
https://github.com/pyRiemann/pyRiemann/runs/4109284600?check_suite_focus=true#step:4:118

pyriemann/utils/viz.py
```diff
@@ -109,29 +113,46 @@ def plot_cospectra(cosp, freqs, *, ylabels=None, title="Cospectra"):
     return fig
 
 
-def plot_waveforms(X, display, *, time=None, **kwargs):
+def plot_waveforms(X, display, *, times=None, **kwargs):
```
agramfort, line 116: reading the code I see that using kwargs is really not necessary here.

when you do things like:

`n_bins = kwargs.get('n_bins', 50)`

just put a parameter in the function signature with `n_bins=50`

it makes the documentation of such functions so much easier.

agramfort:
thx @qbarthelemy !

PR #149, opened 2021-11-17
Add masked and NaN means

qbarthelemy:
This PR aims to:
- add the masked Riemannian mean, for semi-orthogonal masks;
- add the Riemannian NaN-mean, a special case of masked Riemannian mean applied to covariance matrices corrupted by symmetric NaN values (in this case, masks are automatically estimated);
- improve `make_covariances`, ensuring positive eigen values and orthogonal eigen vectors (solve #19 ?);
- add a simple example comparing Euclidean NaN-mean, Riemannian NaN-mean and Riemannian mean after matrix deletion;
- complete tests.

F. Yger, S. Chevallier, Q. Barthélemy, S. Sra. "Geodesically-convex optimization for averaging partially observed covariance matrices", ACML 2021, [PDF](http://proceedings.mlr.press/v129/yger20a/yger20a.pdf)

Disclaimer: this is a pymanopt-free re-implementation of the original work coded with @sylvchev and @ygerf .

For comparison with your method: @ahippert

### DIFF
--- a/tests/test_utils_mean.py
+++ b/tests/test_utils_mean.py
@@ -121,20 +121,36 @@ def test_riemann_mean_masked_shape(init, get_covmats, get_masks):
 
 @pytest.mark.parametrize("init", [True, False])
 def test_riemann_mean_nan_shape(init, get_covmats, rndstate):
-    """Test the riemann nan mean"""
-    n_matrices, n_channels = 50, 4
+    """Test the riemann nan mean shape"""
+    n_matrices, n_channels = 50, 6
     covmats = get_covmats(n_matrices, n_channels)
     emean = np.mean(covmats, axis=0)
     for i in range(n_matrices):
-        corrup_channels = rndstate.randint(n_channels, size=n_channels // 2)
-        for chan in corrup_channels:
-            covmats[i, chan] = np.nan
-            covmats[i, :, chan] = np.nan
+        corrup_channels = rndstate.choice(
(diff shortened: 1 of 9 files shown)


tests/test_utils_mean.py
```diff
@@ -64,45 +67,76 @@ def test_riemann_mean(init, get_covmats_params):
...
+
+
+@pytest.mark.parametrize("init", [True, False])
+def test_riemann_mean_nan_shape(init, get_covmats):
+    """Test the riemann nan mean"""
+    n_matrices, n_channels = 10, 6
+    covmats = get_covmats(n_matrices, n_channels)
+    emean = np.mean(covmats, axis=0)
+    for i in range(n_matrices):
+        corrup_channels = np.random.randint(n_channels, size=n_channels // 2)
```
qbarthelemy, line 129: @sylvchev , how can we use the reproducible random seed here? because it seems not explicitely available...

sylvchev: there's a fixture for that ©️  I'll push the modification.

sylvchev (approved):

sylvchev:
I proposed a new plot, mimicking one of the figure from the article. It is just a suggestion, you could remove it.

qbarthelemy:
> I proposed a new plot, mimicking one of the figure from the article. It is just a suggestion, you could remove it.

The new plot is great! Thanks!

sylvchev:
Ok, LGTM!
Thanks @qbarthelemy 🙏

PR #153, opened 2021-12-03
Add correlation normalization

qbarthelemy:
This PR:
- adds a new covariance normalization "corr", providing correlation matrices, with values in [-1, 1] and diagonal values equal to 1;
- completes doc and tests.


sylvchev (approved):

sylvchev:
Do you want to add an item in what's new?

PR #154, opened 2021-12-08
Implementation of Block Covariance estimation

gabelstein:
This includes block diagonal covariance estimation. 
Useful for tasks where multiple band-pass filters are applied to the original data and covariance between frequencies do not hold information.


pyriemann/estimation.py
```diff
@@ -651,3 +651,74 @@ def transform(self, X):
             covmats[ii] = shrunk_covariance(x, self.shrinkage)
 
         return covmats
+
+
+class BlockCovariances(BaseEstimator, TransformerMixin):
+    """Estimation of block covariance matrix.
+    Perform a block covariance matrix estimation for each given trial. The
+    resulting matrices are block diagonal matrices
+    Parameters
```
sylvchev, line 660:
```suggestion

    Parameters
```

pyriemann/estimation.py
```diff
@@ -651,3 +651,74 @@ def transform(self, X):
...
+    ----------
+    estimator : string (default: 'scm')
+        covariance matrix estimator. For regularization consider 'lwf' or 'oas'
+        For a complete list of estimator, see `utils.covariance`.
+    block_size : list or int
+        Sizes of individual blocks given as int for same-size block or list for
+        varying block sizes.
+    See Also
```
sylvchev, line 668:
```suggestion

    See Also
```

pyriemann/estimation.py
```diff
@@ -651,3 +651,74 @@ def transform(self, X):
...
+    HankelCovariances
+    """
+
+    def __init__(self, block_size, estimator='scm'):
+        """Init."""
+        self.estimator = estimator
+        self.block_size = block_size
+
+    def fit(self, X, y=None):
+        """Fit.
```
sylvchev, line 682:
```suggestion
        """Fit.

```

pyriemann/estimation.py
```diff
@@ -651,3 +651,74 @@ def transform(self, X):
...
+        Do nothing. For compatibility purpose.
+        Parameters
```
sylvchev, line 684:
```suggestion

        Parameters
```

pyriemann/estimation.py
```diff
@@ -651,3 +651,74 @@ def transform(self, X):
...
+        ----------
+        X : ndarray, shape (n_trials, n_channels, n_samples)
+            ndarray of trials.
+        y : ndarray shape (n_trials,)
+            labels corresponding to each trial, not used.
+        Returns
```
sylvchev, line 690:
```suggestion

        Returns
```

pyriemann/estimation.py
```diff
@@ -651,3 +651,74 @@ def transform(self, X):
...
+        -------
+        self : Covariances instance
+            The Covariances instance.
+        """
+        return self
+
+    def transform(self, X):
+        """Estimate block covariance matrices.
+        Parameters
```
sylvchev, line 699:
```suggestion

        Parameters
```

pyriemann/estimation.py
```diff
@@ -651,3 +651,74 @@ def transform(self, X):
...
+        ----------
+        X : ndarray, shape (n_trials, n_channels, n_samples)
+            ndarray of trials.
+        Returns
```
sylvchev, line 703:
```suggestion

        Returns
```

sylvchev (approved):

sylvchev:
Thanks for the PR, this is a nice code. There is some linting errors in CI, that you coud see by running `flake8` on your files.

Could you update what's new page in `doc/whatsnew.rst`?
Could it be interesting to update the `examples/SSVEP/plot_classify_ssvep_mdm.py` with this new estimator?

gabelstein:
Ah yes, sorry for not checking before. 
I'll check the examples and see if it makes sense.

qbarthelemy:
Obviously, block covariance processing is interesting after filter-bank, like in SSVEP paradigm. But I think that the suggested implementation is sub-optimal.

After a filter-bank with non-overlapped frequency bands, you usually build an extended version of the input signal with filter outputs concatenated along channel axis. 
The resulting matrix is block-diagonal, because the off-diagonal blocks are almost null as there is a null inter-frequency covariance.

```
S = [[A, 0, 0],
     [0, B, 0],
     [0, 0, C]]
```

However, this type of matrices has a uselessly too high dimension, leading to: 
ill-conditioned covariance matrices, 
larger memory storage,
and longer computations (complexity of EVD is cubic with the number of channels).

If you consider two block-diagonal matrices:
```
S1 = [[A1, 0, 0],            and            S2 = [[A2, 0, 0],
      [0, B1, 0],                                 [0, B2, 0],
      [0, 0, C1]]                                 [0, 0, C2]]
```
then, you can see that:
- the squared distance between S1 and S2 is simply the sum of squared distance between diagonal blocks:
```
d^2(S1,S2) = d^2(A1, A2) + d^2(B1, B2) + d^2(C1, C2)
```
processing only low-dimensional matrices.
- the mean is the concatenation of means computed on each blocks.

My two cents:
- conceptually, block covariance matrix `S` could be simply stored as `[A, B, C]`, and processed accordingly;
- concretly, I have no idea how to implement in pyRiemann a kind of "block-diagonal matrix frontend view", being actually processed by a "by-block optimized backend computation".

I hope that’s clear...

gabelstein:
I had thought about that as well, but I couldn't find a reasonable solution that didn't include writing and own BlockCovariance class and rewriting all things like eigenvalue decomposition etc for that. 

Interestingly, although it is stored as a n*n matrix, the runtime for EVD and such reduces to about a half from the full covariances (at least in some toy examples that I wrote up). 

So although this is definitely far from optimal code, it could have some usefulness. But if you think that this shouldn't be included like this, I could try to figure out a way to make it more efficient.

sylvchev:
Yes, apart from the heavy solution to write dedicated classes for SPD and blockSPD matrices, which increase code complexity for an unclear efficiency gain, I see another possibility. This is to use [scipy sparse matrices](https://docs.scipy.org/doc/scipy/reference/sparse.html) and there is even block diagonal matrix structure. But the efficiency gain is still unclear as most of our codebase rely on numpy's linear algebra, so the whole codebase requires a major rewriting.

I agree with your general remark @qbarthelemy but I think we could use the code of this PR as it is, especially if this could save some runtime when performing EVD.

sylvchev (approved):

sylvchev:
@gabelstein Do you think you could updating the SSVEP example?

sylvchev (approved):

gabelstein:
> @gabelstein This seems good to me; ok to merge?

Jup, looks good.

sylvchev:
Could you correct the CI error?

qbarthelemy (changes requested):
Ok for me to integrate the current implementation.

pyriemann/utils/covariance.py
```diff
@@ -444,6 +445,55 @@ def coherence(X, window=128, overlap=0.75, fmin=None, fmax=None, fs=None,
     return C, freqs
 
 
+def block_covariances(X, blocks, estimator='cov'):
+    """Compute block diagonal covariance.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_times)
```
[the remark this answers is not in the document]
gabelstein: Depends. This implementation does allow for blocks of varying size. E.g. when you only want to filter a subset of channels in a given filter band. somewhat strange use case, I know, but at my group this is used every now and then. I could replace n_channels by n_filtered_channels.

pyriemann/utils/covariance.py
```diff
@@ -444,6 +445,55 @@ def coherence(X, window=128, overlap=0.75, fmin=None, fmax=None, fs=None,
...
+            Multi-channel time-series.
+        blocks: list of int
+            list of block sizes.
```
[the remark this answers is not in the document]
gabelstein: see above, to allow different block sizes.

pyriemann/utils/covariance.py
```diff
@@ -444,6 +445,55 @@ def coherence(X, window=128, overlap=0.75, fmin=None, fmax=None, fs=None,
...
+              https://scikit-learn.org/stable/modules/generated/sklearn.covariance.MinCovDet.html
+            * 'sch' for Schaefer-Strimmer covariance,
+              http://doi.org/10.2202/1544-6115.1175,
+            * 'corr' for correlation coefficient matrix,
+              https://numpy.org/doc/stable/reference/generated/numpy.corrcoef.html
+
+        Returns
+        -------
+        C : ndarray, shape (n_channels, n_channels, n_freqs)
+            Squared coherence matrices, for each frequency bin.
```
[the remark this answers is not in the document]
gabelstein: See above

pyriemann/utils/covariance.py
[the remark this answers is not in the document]
qbarthelemy: Ok, but in this case, description must be completed.
Users must understand the processing done by the function without reading the code.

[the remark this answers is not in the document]
sylvchev: Yes, I agree with @gabelstein here. For SSVEP, channels could be ["O1-filter1", "Oz-filter1", "O2-filter1", "O1-filter2", "Oz-filter2", "O2-filter2", ..., "O1-filtern", "Oz-filtern", "O2-filtern"]

pyriemann/utils/covariance.py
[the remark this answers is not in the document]
sylvchev: Using different block size with a list of int seems pretty transparent to me, especially with the test examples. Is that ok for you @qbarthelemy ?

pyriemann/utils/covariance.py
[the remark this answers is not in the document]
gabelstein: Added some description, hopefully it's understandable.

sylvchev (approved):

pyriemann/estimation.py
```diff
@@ -651,3 +651,89 @@ def transform(self, X):
...
+                                 'of number of channels of X.')
+
+            blocks = [self.block_size for b in range(n_blocks)]
+
+        else:
+            blocks = self.block_size
+
+        if np.sum(blocks) != n_channels:
+            raise ValueError('Sum of individual block sizes '
+                             'must match number of channels of X.')
```
qbarthelemy, line 737: If the class `BlockCovariances` calls the function `block_covariances`, parameters checking must be placed into the function, to avoid a future bad re-use of this function.

doc/whatsnew.rst
```diff
@@ -36,6 +36,8 @@ v0.2.8.dev
 
 - Add masked and NaN means with Riemannian metric: :func:`pyriemann.utils.mean.maskedmean_riemann` and :func:`pyriemann.utils.mean.nanmean_riemann`
 
+- Add block covariance matrix: :func:`pyriemann.utils.covariance.block_covariances`
+
 - Add ``corr`` option in :func:`pyriemann.utils.covariance.normalize`, to normalize covariance into correlation matrices
```
qbarthelemy, line 41:
```suggestion
- Add ``corr`` option in :func:`pyriemann.utils.covariance.normalize`, to normalize covariance into correlation matrices

- Add block covariance matrix: :func:`pyriemann.utils.covariance.block_covariances`
```

pyriemann/estimation.py
```diff
@@ -651,3 +651,89 @@ def transform(self, X):
...
+
+    Parameters
+    ----------
+    estimator : string (default: 'scm')
+        covariance matrix estimator. For regularization consider 'lwf' or 'oas'
+        For a complete list of estimator, see `utils.covariance`.
+    block_size : list or int
+        Sizes of individual blocks given as int for same-size block or list for
+        varying block sizes.
+
```
qbarthelemy, line 675:
```suggestion

    Notes
    -----
    .. versionadded:: 0.2.8

```

pyriemann/estimation.py
```diff
@@ -651,3 +651,89 @@ def transform(self, X):
...
+    See Also
+    --------
+    ERPCovariances
+    XdawnCovariances
+    CospCovariances
+    HankelCovariances
```
qbarthelemy, line 681:
```suggestion
    Covariances
```

pyriemann/estimation.py
```diff
@@ -651,3 +651,81 @@ def transform(self, X):
...
+        covmats : ndarray, shape (n_matrices, n_channels, n_channels)
+            ndarray of covariance matrices for each trials.
+        """
+        n_matrices, n_channels, n_times = X.shape
+
+        if isinstance(self.block_size, int):
+            n_blocks = n_channels // self.block_size
+            blocks = [self.block_size for b in range(n_blocks)]
+
+        else:
```
qbarthelemy, line 728:
```suggestion
        elif isinstance(self.block_size, list):
```
and code must raise an error otherwise.

pyriemann/utils/covariance.py
```diff
@@ -444,6 +445,53 @@ def coherence(X, window=128, overlap=0.75, fmin=None, fmax=None, fs=None,
...
+                         'must match number of channels of X.')
+
+    covmats = np.empty((n_matrices, n_channels, n_channels))
+    for i in range(n_matrices):
+        blockcov = []
+        idx_start = 0
+        for j in blocks:
+            blockcov.append(est(X[i, idx_start:idx_start+j, :]))
+            idx_start += j
+        covmats[i, :, :] = block_diag(*tuple(blockcov))
```
qbarthelemy, line 490:
```suggestion
        blockcov, idx_start = [], 0
        for j in blocks:
            blockcov.append(est(X[i, idx_start:idx_start+j, :]))
            idx_start += j
        covmats[i] = block_diag(*tuple(blockcov))
```

pyriemann/estimation.py
gabelstein: What if people pass the blocks as ndarrays for whatever reason? Is there a way to check if it's array like?

gabelstein: I just added check for ndarray. If anyone passes any other array-like, it's on them I guess.

qbarthelemy: I don't see the check for ndarray.
And can you raise an error when it is neither a `int` nor a `list`/`ndarray`?

qbarthelemy (approved):
@gabelstein , can you check last modifications?

gabelstein:
> @gabelstein , can you check last modifications?

Everything looks good.

qbarthelemy:
Thx @gabelstein for the contribution!

PR #158, opened 2021-12-14
UMAP wrapper for SPD matrices

gabelstein:
[UMAP](https://umap-learn.readthedocs.io/en/latest/) embedding wrapper for covariance matrices.

I'm also finishing up [Riemannian LLE](http://www.cis.jhu.edu/~rvidal/publications/cvpr08-Riemannian.pdf) in the next days, should I include it in this one or make a new PR when it's done?
When LLE is implemented, I could write an example comparing the three embedding methods.


gabelstein:
Uch, there's a dependency issue with the numba package used in umap which wants an older numpy version. I'll just wait for them to fix that (hopefully soon) and then try again. In the meantime I'll do the LLE.

sylvchev:
Thanks, this is a good idea to have different embedding and it could be very nice to compare the different approaches. Beware that adding a dependency is a choice with lots of long term consequences. Even if UMAP is update its numba dependency, we should have very good reasons to add a dependency to UMAP in pyRiemann (adding thus indirectely numba and others). 

Will you recode the Riemannian LLE from scratch or do you plan to use an existing implementation?

gabelstein:
Good point. I really like UMAP, it often gives nicer results than the other methods, but the dependencies can be quite a pain.

I coded LLE basically from scratch based on the structure in sklearn.manifold._locally_linear and it doesn't require any new packages.

PR #159, opened 2021-12-22
Riemannian Locally Linear Embedding

gabelstein:
As promised, this implements a new embedding method, Locally Linear Embedding using the Riemannian distance (as in [http://www.cis.jhu.edu/~rvidal/publications/cvpr08-Riemannian.pdf](url)). 

There are a few conceptual things to discuss here:

1.  kernel.py: I made a new file in utils as there could be several kernel methods depending on the metric and there are other use cases for the kernel. There could be a generic kernel function consisting of ts-projection and scalar product calculation but it would be rather slow, so I wrote this one standalone. 
2. Should the two functions (riemann_lle and barycenter_weights) that are used in RiemannLLE be moved to another file in utils?
3. with another embedding method present, should Embedding be renamed to SpectralEmbedding?
4. I relied heavily on preexisting code from skelarn for the riemann_lle and barycenter_weights functions. Is there a way this needs to be quoted? I only mentioned it in the notes.

### DIFF
--- a/pyriemann/embedding.py
+++ b/pyriemann/embedding.py
@@ -1,14 +1,20 @@
 """Embedding covariance matrices via manifold learning techniques."""
 
 import numpy as np
+from scipy.linalg import solve
+from scipy.sparse import csr_matrix
+
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.manifold import spectral_embedding
 from sklearn.manifold._locally_linear import null_space
 
-from pyriemann.utils.mean import mean_riemann
-
-from pyriemann.utils.base import logm, invsqrtm
-from sklearn.base import BaseEstimator
-from sklearn.manifold import spectral_embedding
+from .utils.mean import mean_riemann
+from .utils.base import logm, invsqrtm
 from .utils.distance import pairwise_distance
+
+
+
+
 
 
 class Embedding(BaseEstimator):
@@ -107,20 +113,6 @@ class Embedding(BaseEstimator):
         """
         self.fit(X)
         return self.embedding_
-
-
-import numpy as np
-from scipy.linalg import eigh, svd, qr, solve
-from scipy.sparse import eye, csr_matrix
-from scipy.sparse.linalg import eigsh
-
-from sklearn.base import BaseEstimator, TransformerMixin, _UnstableArchMixin
-from sklearn.utils import check_random_state, check_array
-from sklearn.utils._arpack import _init_arpack_v0
-from sklearn.utils.extmath import stable_cumsum
-from sklearn.utils.validation import check_is_fitted
-from sklearn.utils.validation import FLOAT_DTYPES
-from sklearn.neighbors import NearestNeighbors
 
 
 class RiemannLLE(BaseEstimator, TransformerMixin):
--- a/pyriemann/embedding.py
+++ b/pyriemann/embedding.py
@@ -140,6 +140,15 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
     **kwargs
         Keyword arguments passed to sklearn.manifold._locally_linear.null_space
 
+    Attributes
+    ----------
+    embedding_ : ndarray, shape (n_samples, n_components)
+        Stores the embedding vectors
+    error_ : float
+        Reconstruction error associated with `embedding_`
+    data_ : int
+        Training data.
+
     Notes
     -----
     .. versionadded:: 0.2.8
@@ -173,21 +182,20 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
             Returns the instance itself.
 
         """
-        self.data = X
-        self.embedding, self.reconstruction_error = \
-            riemann_lle(X,
-                        self.n_components,
-                        self.n_neighbors,
-                        self.reg,
-                        self.null_space_args)
+        self.data_ = X
+        self.embedding_, self.error_ = riemann_lle(X,
+                                                   self.n_components,
+                                                   self.n_neighbors,
+                                                   self.reg,
+                                                   self.null_space_args)
         return self
 
     def transform(self, X, y=None):
-        pairwise_distances = pairwise_distance(X, self.data, metric='riemann')
+        pairwise_distances = pairwise_distance(X, self.data_, metric='riemann')
         ind = np.array([np.argsort(dist)[1:self.n_neighbors + 1]
                         for dist in pairwise_distances])
 
-        weights = barycenter_weights(X, self.data, ind, reg=self.reg)
+        weights = barycenter_weights(X, self.data_, ind, reg=self.reg)
 
         X_new = np.empty((X.shape[0], self.n_components))
         for i in range(X.shape[0]):
@@ -195,13 +203,13 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
         return X_new
 
     def fit_transform(self, X, y=None):
-        self.data = X
-        self.embedding, self.reconstruction_error = riemann_lle(X,
-                                                                self.n_components,
-                                                                self.n_neighbors,
-                                                                self.reg,
-                                                                self.null_space_args)
-        return self.embedding
+        self.data_ = X
+        self.embedding_, self.error_ = riemann_lle(X,
+                                                   self.n_components,
+                                                   self.n_neighbors,
+                                                   self.reg,
+                                                   self.null_space_args)
+        return self.embedding_
 
 
 def barycenter_weights(X, Y, indices, reg=1e-3):
@@ -249,11 +257,16 @@ def barycenter_weights(X, Y, indices, reg=1e-3):
     return B
 
 
-def riemann_lle(X, n_components=2, n_neighbors=5, reg=1e-3, null_space_args={}):
+def riemann_lle(X,
+                n_components=2,
+                n_neighbors=5,
+                reg=1e-3,
+                null_space_args={}):
     """Riemannian Locally Linear Embedding (LLE).
 
-    Riemannian LLE as proposed in [1]_. LLE is a non-linear, neighborhood-preserving
-    dimensionality reduction algorithm which consists of three main steps:
+    Riemannian LLE as proposed in [1]_. LLE is a non-linear, neighborhood-
+    preserving dimensionality reduction algorithm which consists of three
+    main steps:
     For each point x,
     1.  find its k nearest neighbors KNN(x) and
     2.  calculate the best reconstruction of x based on its KNN.
@@ -278,12 +291,14 @@ def riemann_lle(X, n_components=2, n_neighbors=5, reg=1e-3, null_space_args={}):
     reg : float, default: 1e-3
         Regularization parameter.
     null_space_args : dict, default {}
-        Keyword arguments passed to sklearn.manifold._locally_linear.null_space.
+        Keyword arguments passed to
+        sklearn.manifold._locally_linear.null_space.
 
     Notes
     -----
     .. versionadded:: 0.2.8
-    See sklearn.manifold._locally_linear.locally_linear_embedding for original code.
+    See sklearn.manifold._locally_linear.locally_linear_embedding for original
+    code.
 
     References
     ----------
@@ -300,7 +315,8 @@ def riemann_lle(X, n_components=2, n_neighbors=5, reg=1e-3, null_space_args={}):
     B = barycenter_weights(X, X, neighbors, reg=reg)
 
     indptr = np.arange(0, n_matrices * n_neighbors + 1, n_neighbors)
-    W = csr_matrix((B.ravel(), neighbors.ravel(), indptr), shape=(n_matrices, n_matrices))
+    W = csr_matrix((B.ravel(), neighbors.ravel(), indptr), shape=(n_matrices,
+                                                                  n_matrices))
     M = (W.T * W - W.T - W).toarray()
     M.flat[:: M.shape[0] + 1] += 1  # W = W - I = W - I
     return null_space(
--- a/pyriemann/embedding.py
+++ b/pyriemann/embedding.py
@@ -8,13 +8,8 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.manifold import spectral_embedding
 from sklearn.manifold._locally_linear import null_space
 
-from .utils.mean import mean_riemann
-from .utils.base import logm, invsqrtm
+from .utils.kernel import kernel_riemann
 from .utils.distance import pairwise_distance
-
-
-
-
 
 
 class Embedding(BaseEstimator):
@@ -107,7 +102,7 @@ class Embedding(BaseEstimator):
 
         Returns
         -------
-        X_new: array-like, shape (n_matrices, n_components)
+        X_new: ndarray, shape (n_matrices, n_components)
             Coordinates of embedded matrices.
 
         """
@@ -116,14 +111,43 @@ class Embedding(BaseEstimator):
 
 
 class RiemannLLE(BaseEstimator, TransformerMixin):
-    """
-    Wrapper for UMAP for dimensionality reduction on positive definite matrices with a Riemannian metric
+    """Riemannian Locally Linear Embedding (LLE).
+
+    Riemannian LLE as proposed in [1]_. LLE is a non-linear, neighborhood-preserving
+    dimensionality reduction algorithm which consists of three main steps:
+    For each point x,
+    1.  find its k nearest neighbors KNN(x) and
+    2.  calculate the best reconstruction of x based on its KNN.
+    3.  Then calculate a low-dimensional embedding for all points based on
+        the weights in step 2.
+
     Parameters
     ----------
-    metric : str (default 'riemann')
-        string code for the metric from .utils.distance
-    **kwargs : dict
-        arguments to pass to umap.
+    n_components : int, default: 2
+        Dimensionality of projected space.
+    n_neighbors : int, default: 5
+        Number of neighbors for reconstruction of each point.
+    n_jobs : int, default: 1
+        The number of jobs to use for the computation. This works by computing
+        each of the class centroid in parallel.
+        If -1 all CPUs are used. If 1 is given, no parallel computing code is
+        used at all, which is useful for debugging. For n_jobs below -1,
+        (n_cpus + 1 + n_jobs) are used. Thus for n_jobs = -2, all CPUs but one
+        are used.
+    reg : float, default: 1e-3
+        Regularization parameter.
+    **kwargs
+        Keyword arguments passed to sklearn.manifold._locally_linear.null_space.
+
+    Notes
+    -----
+    .. versionadded:: 0.2.8
+
+    References
+    ----------
+    .. [1] A. Goh and R. Vidal, "Clustering and dimensionality reduction
+        on Riemannian manifolds", 2008 IEEE Conference on Computer Vision
+        and Pattern Recognition, June 2008
     """
 
     def __init__(self, n_components=2, n_neighbors=5, n_jobs=1, reg=1e-3, **kwargs):
@@ -134,14 +158,31 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
         self.null_space_args = kwargs
 
     def fit(self, X, y=None):
+        """Fit the model from data in X.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            SPD matrices.
+
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+
+        """
         self.data = X
-        self.embedding, self.reconstruction_error = riemann_lle(X, self.n_components, self.n_neighbors, self.reg,
+        self.embedding, self.reconstruction_error = riemann_lle(X,
+                                                                self.n_components,
+                                                                self.n_neighbors,
+                                                                self.reg,
                                                                 self.null_space_args)
         return self
 
     def transform(self, X, y=None):
         pairwise_distances = pairwise_distance(X, self.data, metric='riemann')
-        ind = np.array([np.argsort(dist)[1:self.n_neighbors + 1] for dist in pairwise_distances])
+        ind = np.array([np.argsort(dist)[1:self.n_neighbors + 1]
+                        for dist in pairwise_distances])
 
         weights = barycenter_weights(X, self.data, ind, reg=self.reg)
 
@@ -152,7 +193,10 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
 
     def fit_transform(self, X, y=None):
         self.data = X
-        self.embedding, self.reconstruction_error = riemann_lle(X, self.n_components, self.n_neighbors, self.reg,
+        self.embedding, self.reconstruction_error = riemann_lle(X,
+                                                                self.n_components,
+                                                                self.n_neighbors,
+                                                                self.reg,
                                                                 self.null_space_args)
         return self.embedding
 
@@ -161,33 +205,36 @@ def barycenter_weights(X, Y, indices, reg=1e-3):
     """Compute Riemannian barycenter weights of X from Y along the first axis.
     We estimate the weights to assign to each point in Y[indices] to recover
     the point X[i] by geodesic interpolation. The barycenter weights sum to 1.
+
     Parameters
     ----------
-    X : array-like, shape (n_samples, n_dim)
-    Y : array-like, shape (n_samples, n_dim)
-    indices : array-like, shape (n_samples, n_dim)
-            Indices of the points in Y used to compute the barycenter
+    X : ndarray, shape (n_matrices, n_dim)
+    Y : ndarray, shape (n_matrices, n_dim)
+    indices : ndarray, shape (n_matrices, n_dim)
+        Indices of the points in Y used to compute the barycenter
     reg : float, default=1e-3
         amount of regularization to add for the problem to be
         well-posed in the case of n_neighbors > n_dim
+
     Returns
     -------
-    B : array-like, shape (n_samples, n_neighbors)
+    B : ndarray, shape (n_matrices, n_neighbors)
+
     Notes
     -----
-    See developers note for more information.
+    .. versionadded:: 0.2.8
+    See sklearn.manifold._locally_linear.barycenter_weights for original code.
     """
-    # indices = check_array(indices, dtype=int)
 
-    n_samples, n_neighbors = indices.shape
-    assert X.shape[0] == n_samples
+    n_matrices, n_neighbors = indices.shape
+    assert X.shape[0] == n_matrices
 
-    B = np.empty((n_samples, n_neighbors), dtype=X.dtype)
+    B = np.empty((n_matrices, n_neighbors), dtype=X.dtype)
     v = np.ones(n_neighbors, dtype=X.dtype)
 
     for i in range(len(X)):
         X_neighbors = Y[indices[i]]
-        G = riemann_kernel_matrix(X_neighbors, X_neighbors, X[i])
+        G = kernel_riemann(X_neighbors, Cref=X[i])
         trace = np.trace(G)
         if trace > 0:
             R = reg * trace
@@ -200,14 +247,57 @@ def barycenter_weights(X, Y, indices, reg=1e-3):
 
 
 def riemann_lle(X, n_components=2, n_neighbors=5, reg=1e-3, null_space_args={}):
-    n_samples = X.shape[0]
+    """Riemannian Locally Linear Embedding (LLE).
+
+    Riemannian LLE as proposed in [1]_. LLE is a non-linear, neighborhood-preserving
+    dimensionality reduction algorithm which consists of three main steps:
+    For each point x,
+    1.  find its k nearest neighbors KNN(x) and
+    2.  calculate the best reconstruction of x based on its KNN.
+    3.  Then calculate a low-dimensional embedding for all points based on
+        the weights in step 2.
+
+    Parameters
+    ----------
+    X : ndarray, shape (n_matrices, n_channels, n_channels)
+        ndarray of SPD matrices.
+    n_components : int, default: 2
+        Dimensionality of projected space.
+    n_neighbors : int, default: 5
+        Number of neighbors for reconstruction of each point.
+    n_jobs : int, default: 1
+        The number of jobs to use for the computation. This works by computing
+        each of the class centroid in parallel.
+        If -1 all CPUs are used. If 1 is given, no parallel computing code is
+        used at all, which is useful for debugging. For n_jobs below -1,
+        (n_cpus + 1 + n_jobs) are used. Thus for n_jobs = -2, all CPUs but one
+        are used.
+    reg : float, default: 1e-3
+        Regularization parameter.
+    null_space_args : dict, default {}
+        Keyword arguments passed to sklearn.manifold._locally_linear.null_space.
+
+    Notes
+    -----
+    .. versionadded:: 0.2.8
+    See sklearn.manifold._locally_linear.locally_linear_embedding for original code.
+
+    References
+    ----------
+    .. [1] A. Goh and R. Vidal, "Clustering and dimensionality reduction
+        on Riemannian manifolds", 2008 IEEE Conference on Computer Vision
+        and Pattern Recognition, June 2008
+    """
+
+    n_matrices, n_channels, n_channels = X.shape
     pairwise_distances = pairwise_distance(X, metric='riemann')
-    neighbors = np.array([np.argsort(dist)[1:n_neighbors + 1] for dist in pairwise_distances])
+    neighbors = np.array([np.argsort(dist)[1:n_neighbors + 1]
+                          for dist in pairwise_distances])
 
     B = barycenter_weights(X, X, neighbors, reg=reg)
 
-    indptr = np.arange(0, n_samples * n_neighbors + 1, n_neighbors)
-    W = csr_matrix((B.ravel(), neighbors.ravel(), indptr), shape=(n_samples, n_samples))
+    indptr = np.arange(0, n_matrices * n_neighbors + 1, n_neighbors)
+    W = csr_matrix((B.ravel(), neighbors.ravel(), indptr), shape=(n_matrices, n_matrices))
     M = (W.T * W - W.T - W).toarray()
     M.flat[:: M.shape[0] + 1] += 1  # W = W - I = W - I
     return null_space(
@@ -216,34 +306,3 @@ def riemann_lle(X, n_components=2, n_neighbors=5, reg=1e-3, null_space_args={}):
         k_skip=1,
         **null_space_args
     )
-
-
-def riemann_kernel_matrix(X, Y=None, Cref=None):
-    if Cref is None:
-        G = mean_riemann(X)
-        G_invsq = invsqrtm(G)
-
-    else:
-        G_invsq = invsqrtm(Cref)
-
-    Ntx, Ne, Ne = X.shape
-
-    X_ = np.zeros((Ntx, Ne, Ne))
-    for index in range(Ntx):
-        X_[index] = logm(G_invsq @ X[index] @ G_invsq)
-
-    if Y is None:
-        Nty, Ne, Ne = X.shape
-        Y_ = X_
-
-    else:
-        Nty, Ne, Ne = Y.shape
-        Y_ = np.zeros((Nty, Ne, Ne))
-        for index in range(Nty):
-            Y_[index] = logm(G_invsq @ Y[index] @ G_invsq)
-
-    res = np.zeros((Nty, Ntx))
-    for i in range(Nty):
-        for j in range(Ntx):
-            res[i][j] = np.trace(X_[i] @ Y_[j])
-    return res
--- a/pyriemann/embedding.py
+++ b/pyriemann/embedding.py
@@ -113,8 +113,9 @@ class Embedding(BaseEstimator):
 class RiemannLLE(BaseEstimator, TransformerMixin):
     """Riemannian Locally Linear Embedding (LLE).
 
-    Riemannian LLE as proposed in [1]_. LLE is a non-linear, neighborhood-preserving
-    dimensionality reduction algorithm which consists of three main steps:
+    Riemannian LLE as proposed in [1]_. LLE is a non-linear,
+    neighborhood-preserving dimensionality reduction algorithm which
+    consists of three main steps:
     For each point x,
     1.  find its k nearest neighbors KNN(x) and
     2.  calculate the best reconstruction of x based on its KNN.
@@ -137,7 +138,7 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
     reg : float, default: 1e-3
         Regularization parameter.
     **kwargs
-        Keyword arguments passed to sklearn.manifold._locally_linear.null_space.
+        Keyword arguments passed to sklearn.manifold._locally_linear.null_space
 
     Notes
     -----
@@ -150,7 +151,8 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
         and Pattern Recognition, June 2008
     """
 
-    def __init__(self, n_components=2, n_neighbors=5, n_jobs=1, reg=1e-3, **kwargs):
+    def __init__(self, n_components=2, n_neighbors=5,
+                 n_jobs=1, reg=1e-3, **kwargs):
         self.n_components = n_components
         self.n_neighbors = n_neighbors
         self.n_jobs = n_jobs
@@ -172,11 +174,12 @@ class RiemannLLE(BaseEstimator, TransformerMixin):
 
         """
         self.data = X
-        self.embedding, self.reconstruction_error = riemann_lle(X,
-                                                                self.n_components,
-                                                                self.n_neighbors,
-                                                                self.reg,
-                                                                self.null_space_args)
+        self.embedding, self.reconstruction_error = \
+            riemann_lle(X,
+                        self.n_components,
+                        self.n_neighbors,
+                        self.reg,
+                        self.null_space_args)
         return self
 
     def transform(self, X, y=None):
--- a/pyriemann/embedding.py
+++ b/pyriemann/embedding.py
@@ -1,6 +1,11 @@
 """Embedding covariance matrices via manifold learning techniques."""
 
 import numpy as np
+from sklearn.manifold._locally_linear import null_space
+
+from pyriemann.utils.mean import mean_riemann
+
+from pyriemann.utils.base import logm, invsqrtm
 from sklearn.base import BaseEstimator
 from sklearn.manifold import spectral_embedding
 from .utils.distance import pairwise_distance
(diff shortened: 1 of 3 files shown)


pyriemann/embedding.py
```diff
@@ -1,8 +1,14 @@
 """Embedding covariance matrices via manifold learning techniques."""
 
 import numpy as np
-from sklearn.base import BaseEstimator
+from scipy.linalg import solve
+from scipy.sparse import csr_matrix
+
+from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.manifold import spectral_embedding
+from sklearn.manifold._locally_linear import null_space
```
agramfort, line 9: do not import private functions from sklearn. THere is no backward compat guarantee to these. It can break pyriemann at any time

pyriemann/embedding.py
```diff
@@ -96,9 +102,218 @@ def fit_transform(self, X, y=None):
...
+
+def barycenter_weights(X, Y, indices, reg=1e-3):
+    """Compute Riemannian barycenter weights of X from Y along the first axis.
+    We estimate the weights to assign to each point in Y[indices] to recover
+    the point X[i] by geodesic interpolation. The barycenter weights sum to 1.
+
+    Parameters
+    ----------
+    X : ndarray, shape (n_matrices, n_dim)
+    Y : ndarray, shape (n_matrices, n_dim)
```
agramfort, line 215: explain what X and Y are

pyriemann/embedding.py
```diff
@@ -96,9 +102,218 @@ def fit_transform(self, X, y=None):
...
+    indices : ndarray, shape (n_matrices, n_dim)
+        Indices of the points in Y used to compute the barycenter
```
agramfort, line 217: why not passing Y[indices] directly to the function?

pyriemann/embedding.py
```diff
@@ -96,9 +102,218 @@ def fit_transform(self, X, y=None):
...
+    neighbors = np.array([np.argsort(dist)[1:n_neighbors + 1]
+                          for dist in pairwise_distances])
+
+    B = barycenter_weights(X, X, neighbors, reg=reg)
+
+    indptr = np.arange(0, n_matrices * n_neighbors + 1, n_neighbors)
+    W = csr_matrix((B.ravel(), neighbors.ravel(), indptr), shape=(n_matrices,
+                                                                  n_matrices))
+    M = (W.T * W - W.T - W).toarray()
+    M.flat[:: M.shape[0] + 1] += 1  # W = W - I = W - I
```
agramfort, line 313: here you add 1 to M and not remove 1 to W

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+
+from .mean import mean_riemann
+from .base import invsqrtm, logm
+
+
+def kernel_riemann(X, Y=None, Cref=None):
```
agramfort, line 7: this is to me ambiguous as you can use many metrics. Here one is implied

pyriemann/embedding.py
gabelstein: to make it more readable and universal I guess. mostly because they did it the same way in sklearn. can be changed, though.

pyriemann/embedding.py
gabelstein: that's how they did it in sklearn and the two lines should do just what is asked.

pyriemann/utils/kernel.py
gabelstein: this is the way it's been done in this package, e.g. in distance, geodesic. they're called distance_riemann, geodesic_riemann and so on, fo other metrics they're called distance_logeuclid, geodesic_logeuclid etc. the generic thing is then called distance or geodesic with a string passed. I can implement the generic thing with a string passed as well (although it takes only one string so far).

agramfort: @qbarthelemy what do you think?

pyriemann/embedding.py
agramfort: it does not mean it should copy pasted :)

disclaimer I wrote a quiet some part of sklearn myself so I can criticise :)

pyriemann/embedding.py
agramfort: what do you think? is there a computational reason or a reason elsewhere?

pyriemann/embedding.py
gabelstein: Seems that the comment is wrong. it should be
M = (W - I).T@(W - I) 
which boils down to 
W.T@W - W - W.T + I
which is exactly what they did although I don't know why they write it up so weirdly. I'll change it just so it's more readable.

pyriemann/embedding.py
gabelstein: Uhm, good question. I think not. But I'll check more thoroughly as I go through it again

pyriemann/embedding.py
agramfort: feel free to send a PR on scikit-learn to fix the comment there too 🙏

gabelstein: will do!

pyriemann/utils/kernel.py
gabelstein: I added the generic kernel function (which obviously doesn't do anything but the riemannian kernel) to conform with the previous style

agramfort:
@gabelstein would you have time to add an example to the documentation so this feature is more easily discoverable?

gabelstein:
> @gabelstein would you have time to add an example to the documentation so this feature is more easily discoverable?

Yup. Can't test it unfortunately due to some SSL-error, but just copied the code so should work fine anyways.

qbarthelemy (changes requested):
Good contribution!

Do not forget to complete the API file (`doc\api.rst`), with `RiemannLLE`, but also with `kernel` and `kernel_riemann`.

pyriemann/embedding.py
```diff
@@ -92,13 +99,277 @@ def fit_transform(self, X, y=None):
...
+        """
+        self.data_ = X
+        self.embedding_, self.error_ = riemann_lle(X,
+                                                   self.n_components,
+                                                   self.n_neighbors,
+                                                   self.reg)
+        return self.embedding_
+
+
+def barycenter_weights(X, Y, indices, reg=1e-3):
```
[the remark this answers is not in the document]
gabelstein: I'd say to make them usable outside of the class (I'm not a very big fan of class functions).

pyriemann/embedding.py
```diff
@@ -92,13 +99,277 @@ def fit_transform(self, X, y=None):
...
+    """Compute Riemannian barycenter weights of X from Y along the first axis.
+    We estimate the weights to assign to each point in Y[indices] to recover
+    the point X[i] by geodesic interpolation. The barycenter weights sum to 1.
+
+    Parameters
+    ----------
+    X : ndarray, shape (n_matrices, n_dim)
```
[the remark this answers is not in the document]
gabelstein: Oh, it should actually say (n_matrices, n_channels, n_channels). Those are still the original matrices.

pyriemann/embedding.py
```diff
@@ -92,13 +99,277 @@ def fit_transform(self, X, y=None):
...
+        ndarray of SPD matrices.
```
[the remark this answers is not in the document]
gabelstein: see above

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,97 @@
...
+    if Cref is None:
+        G = mean_riemann(X)
+        G_invsq = invsqrtm(G)
+
+    else:
+        G_invsq = invsqrtm(Cref)
+
+    n_matrices_X, n_channels, n_channels = X.shape
+
+    X_ = np.matmul(G_invsq, np.matmul(X, G_invsq))
```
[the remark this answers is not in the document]
gabelstein: I don't think that works, because X has shape (n_matrices, n_channels, n_channels) and G_invsq has shape(n_channels, n_channels). It needs to be multiplied from both sides to each matrix in X and matmul can do this vectorized (and thus in parallel).

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,97 @@
...
+    X_ = np.array([logm(X_[index]) for index in range(n_matrices_X)])
+
+    if isinstance(Y, type(None)) or np.array_equal(X, Y):
+        Y_ = X_
+
+    else:
+        n_matrices_Y, n_channels, n_channels = Y.shape
```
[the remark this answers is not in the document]
gabelstein: Doesn't have to be, the kernel can be calculated for inputs with different number of matrices. For example needs to be possible in SVCs where the test data kernel matrix is calculated with the original data. 
I could implement a check that the n_channels of both are the same.

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Y_ = np.matmul(G_invsq, np.matmul(Y, G_invsq))
```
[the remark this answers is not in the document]
gabelstein: see above

tests/test_embedding.py
```diff
@@ -22,3 +26,46 @@ def test_fit_independence(get_covmats):
     # retraining with different size should erase previous fit
     new_covmats = covmats[:, :-1, :-1]
     embd.fit_transform(new_covmats)
+
+
+def test_rlle_embedding(get_covmats):
+    """Test RiemannLLE embedding fit_transform."""
+    n_trials, n_channels, n_comp = 6, 3, 2
```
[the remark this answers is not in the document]
gabelstein: will do

tests/test_utils_kernel.py
```diff
@@ -0,0 +1,51 @@
...
+
+from pyriemann.utils.test import is_sym
+
+
+def test_riemann_kernel(rndstate, get_covmats):
+    """Test Riemannian Kernel build"""
+    n_trials, n_channels = 5, 3
+    cov = get_covmats(n_trials, n_channels)
+    K = kernel_riemann(cov, cov, np.eye(n_channels))
+    assert is_sym(K)
```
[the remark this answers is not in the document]
gabelstein: I think not, doesn't even have to be positive semi definite. 
I also stumbled across this, but I think that it's only a misconception from mercer's kernel theorem which says, that for each positive (semi) definite kernel there is a space where that kernel is the scalar product. 
But the other way around does't have to be true, it seems. Not every kernel matrix that comes from an arbitrary scalar product ist positive (semi) definite. Here, we are given the scalar product explicitly.

examples/ERP/plot_embedding_MEG.py
```diff
@@ -78,3 +82,26 @@
...
+###############################################################################
+# with Riemannian Locally Linear Embedding (RLLE)
+
+rlle = RiemannLLE(n_components=2)
+embd = rlle.fit_transform(covs)
+
+###############################################################################
+# Plot the two first components of the RLLE embedded points
+
+fig, ax = plt.subplots(figsize=(7, 8), facecolor='white')
```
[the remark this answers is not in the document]
gabelstein: That function is unfortunately assuming rescaled embedding values between -1 and 1, which I don't provide and which would break the transform function for new data points.

tests/test_utils_kernel.py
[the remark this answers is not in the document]
gabelstein: Okay, there appear to be some numerical errors when computing the kernel matrix and its eigenvalues from the generated spds. There are some eigenvalues very close to 0, but in the negative domain. They should be 0, but aren't. There are only n_channels positive eigenvalues, the rest is (up to numerical errors) zero. Not quite sure how to handle this.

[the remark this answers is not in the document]
gabelstein: So this is not a problem for the Embedding scenario, because it regularizes the kernel matrix anyways in the barycenter_weights function to make it positive definite as before it's only semi-definite. 
For other applications (like the RIemannian SVC that will be my next PR), this might be problematic. I'll have to think about this a bit more. Still unsure if we should implement it as it is or set the eigenvalues to zero somewhat manually.

agramfort:
+1 for renaming Embedding with proper depreciation

tests/test_utils_kernel.py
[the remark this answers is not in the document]
gabelstein: Three suggestions:
1. regularize kernel matrix in the kernel function by adding some scaled identity (I chose 1e-10) and remove regularization in riemann_lle.
2. compute eigenvalues, set eigenvalues with low absolute value (below 1e-10) to zero. Could still induce errors when recalculating the matrix with its eigenvectors and the adjusted eigenvalues.
3. Leave everything as it is and only test for symmetry.

I prefer 1.

qbarthelemy (changes requested):

examples/ERP/plot_embedding_MEG.py
```diff
@@ -50,19 +51,22 @@
...
 nfilter = 4
 xdwn = XdawnCovariances(estimator='scm', nfilter=nfilter)
 split = train_test_split(X, y, train_size=0.25, random_state=42)
 Xtrain, Xtest, ytrain, ytest = split
 covs = xdwn.fit(Xtrain, ytrain).transform(Xtest)
 
+###############################################################################
+# with Spectral Embedding (SE)
+
 lapl = Embedding(metric='riemann', n_components=2)
```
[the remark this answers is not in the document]
gabelstein: Will do. But I have no idea how to do a deprecation warning.

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,107 @@
...
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            First set of SPD matrices.
+        Y : None | ndarray, shape (n_matrices, n_channels, n_channels)
+            Second set of SPD matrices. If None, Y is set to X.
+        Cref : None | ndarray, shape (n_channels, n_channels)
+            Reference point for the tangent space and inner product
+            calculation. If None, Cref is calculated as the mean of X
+            according to the specified metric.
+        metric : {'riemann'}
```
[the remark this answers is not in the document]
gabelstein: Can do, although I think the respective kernels look different when we start from the geodesics and do the derivation of tangent space projection using the logarithmic map and scalar product on tangent space <A,B>_P = tr(P^{-1/2}AP^{-1/2}P^{-1/2}BP^{-1/2}) there. 
If my calculations are correct, these are the resulting kernels:

For euclid, it would be 
K^P_{i,j} = {tr}(P^{-1/2}(X_i - P)P^{-1/2} P^{-1/2}(Y_i - P)P^{-1/2})

For logeuclid it would be 
K^P_{i,j} = {tr}((\log(A) - \log(P)(\log(B) - \log(P)).

It is quite possible that this gives kernel matrices with the same properties for RSVC and RLLE as compared to your kernels, but I'll have to look at that in a bit more detail.

pyriemann/embedding.py
```diff
@@ -92,13 +99,269 @@ def fit_transform(self, X, y=None):
...
+    neighbors = np.array([np.argsort(dist)[1:n_neighbors + 1]
+                          for dist in pairwise_distances])
+
+    B = barycenter_weights(X, X, neighbors, reg=reg)
+
+    indptr = np.arange(0, n_matrices * n_neighbors + 1, n_neighbors)
+    W = csr_matrix((B.ravel(), neighbors.ravel(), indptr), shape=(n_matrices,
+                                                                  n_matrices))
+    # M = (W - I).T * (W - I) = W.T * W - W.T - W + I
+    M = (W.T * W - W.T - W).toarray()
```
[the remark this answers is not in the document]
gabelstein: the line below it adds the identity

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,107 @@
...
+    # for i in range(n_matrices_X):
+    #     for j in range(n_matrices_Y):
+    #         K[i][j] = np.trace(X_[i] @ Y_[j])
+    # einsum does same as that just a looooooot faster
+
+    K = np.einsum('abc,dbc->ad', X_, Y_)
+
+    # regularization due to numerical errors
+    if np.array_equal(X_, Y_):
+        K.flat[:: n_matrices_X + 1] += reg
```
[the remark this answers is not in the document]
gabelstein: To make the individual kernel methods usable (and safe) without calling them from the generic kernel function, I'd prefer to leave it as it is and include regularization in each individual function.

tests/test_embedding.py
```diff
@@ -1,24 +1,71 @@
 from conftest import get_metrics
-from pyriemann.embedding import Embedding
+import numpy as np
+from pyriemann.embedding import (Embedding,
+                                 RiemannLLE,
+                                 barycenter_weights,
+                                 riemann_lle)
```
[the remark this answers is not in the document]
gabelstein: I removed it alltogether. I adjusted the code from sklearn without giving it too much attention, and I think the sklearn code is not very well described. It doesn't really give the null_space of the matrix. So I just moved the code to the `riemann_lle` function since it's not called anywhere else.

agramfort:
can you share a 2D plot of the embedding you get in the example?

gabelstein:
> can you share a 2D plot of the embedding you get in the example?
> […](#)
> Message ID: ***@***.***>



LLE
image

Spectral 
image

pyriemann/utils/kernel.py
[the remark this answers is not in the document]
gabelstein: It gives different kernel matrices leading to different embeddings. The embeddings look a bit better with my kernels so I implemented it like this.

examples/ERP/plot_embedding_MEG.py
```diff
@@ -1,25 +1,26 @@
...
 """
-# Authors: Pedro Rodrigues <pedro.rodrigues01@gmail.com>
-#
+# Authors:  Pedro Rodrigues <pedro.rodrigues01@gmail.com>,
+#           Gabriel Wagner vom Berg <gabriel@bccn-berlin.de>
 # License: BSD (3-clause)
 
 from pyriemann.estimation import XdawnCovariances
-from pyriemann.embedding import Embedding
+from pyriemann.utils.viz import plot_embedding
```
agramfort, line 15: honestly I would not start putting such a simple viz code in the public API only for illustration in one example.

I would keep for now the plot_embedding function defined here.

pyriemann/embedding.py
```diff
@@ -1,12 +1,17 @@
...
+from scipy.sparse import csr_matrix
+
+from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.manifold import spectral_embedding
+
+from .utils.kernel import kernel
 from .utils.distance import pairwise_distance
 
 
-class Embedding(BaseEstimator):
```
agramfort, line 9: Embedding class was released in v0.2.5 so technically you should deprecate here and not break silently public API.

agramfort (commented):
@plcrodrigues can you have a look as it touches your code?

qbarthelemy (changes requested):
Wow, great improvements!

pyriemann/embedding.py
```diff
@@ -92,13 +99,270 @@ def fit_transform(self, X, y=None):
...
+        else:
+            R = reg
+        G.flat[:: n_neighbors + 1] += R
+        w = solve(G, v, sym_pos=True)
+        B[i, :] = w / np.sum(w)
+    return B
+
+
+def riemann_lle(X,
+                n_components=2,
```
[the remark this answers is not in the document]
gabelstein: why the *?

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,105 @@
+import numpy as np
+
+from .mean import mean_riemann
+from .base import invsqrtm, logm
+
+
+def kernel(X, Y=None, Cref=None, metric='riemann', reg=1e-10):
+    r""" Calculates the Kernel matrix K of inner products of two sets
+         X and Y of SPD matrices on tangent space of C according to the
```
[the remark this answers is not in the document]
gabelstein: exactly, but then it's different from other files. maybe keep Cref as parameter names but use C in equations? somewhat inconsistent, but I guess everyone will get it in the equation. Or change it to C everywhere (although I don't know if C and Cref are used somewhere in the same function).

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,239 @@
...
+        Returns
+        ----------
+        K : ndarray, shape (n_matrices, n_matrices)
+            The kernel matrix of X and Y.
+
+        Notes
+        -----
+        .. versionadded:: 0.2.8
+        """
+    try:
```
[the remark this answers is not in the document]
gabelstein: could be, but that again raises the problem of either not being able to use the standalone functions or duplicating the code in the generic function and all others.

tests/test_utils_kernel.py
```diff
@@ -0,0 +1,76 @@
+from pyriemann.utils.kernel import (kernel,
+                                    kernel_euclid,
+                                    kernel_logeuclid,
+                                    kernel_riemann)
+from pyriemann.utils.mean import mean_covariance
+from pyriemann.utils.test import is_pos_semi_def as is_spsd
```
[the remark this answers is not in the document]
gabelstein: spsd is the goal for kernels.

examples/ERP/plot_embedding_MEG.py
gabelstein: sorry, what?

pyriemann/utils/kernel.py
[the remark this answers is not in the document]
gabelstein: I rewrote the kernel module completely to avoid any code duplicates. 
Quite a bit different now, a little less readable I think, but it would allow for passing generic kernel functions if `kernel` is adjusted a bit. 
what do you think?

pyriemann/embedding.py
```diff
@@ -92,13 +99,270 @@ def fit_transform(self, X, y=None):
...
+            R = reg * trace
+        else:
+            R = reg
+        G.flat[:: n_neighbors + 1] += R
+        w = solve(G, v, sym_pos=True)
+        B[i, :] = w / np.sum(w)
+    return B
+
+
+def riemann_lle(X,
```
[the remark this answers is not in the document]
gabelstein: is it correct like this? don't know how to do mixtures of classes and functions in api

pyriemann/embedding.py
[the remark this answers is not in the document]
qbarthelemy: When there are many parameters, it is a better pratise to define them as keywords only.

tests/test_utils_kernel.py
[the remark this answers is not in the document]
qbarthelemy: In this case, you should import `is_sym_pos_semi_def`.

pyriemann/embedding.py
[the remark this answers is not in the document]
gabelstein: Hm, just adding the '*' like this unfortunately breaks anything.
It says
E       TypeError: locally_linear_embedding() takes 1 positional argument but 5 were given
when the function is called with parameters specified.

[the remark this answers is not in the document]
gabelstein: Ah, it forces you to always specify the keyword arguments, right?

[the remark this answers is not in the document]
qbarthelemy: Yes. And, their position / order is not hardcoded.

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,207 @@
...
+    return _apply_matrix_kernel(kernelfct, X, Y, Cref, reg)
+
+
+def kernel_euclid(X, Y=None, Cref=None, reg=1e-10):
+    r""" Calculates the Euclidean Kernel matrix K of inner products of two sets
+     X and Y of SPD matrices on tangent space of C by calculating pairwise
+
+    .. math::
+        K_{i,j} =
+        {tr}(P^{-1/2}(X_i - P)P^{-1/2} P^{-1/2}(Y_j - P)P^{-1/2})
```
qbarthelemy, line 120: I don't know how to chose between `C` and `Cref`. But I am sure that `P` is a bad choice ;-)
For me, we must consitent across an entire module, at least; across that package, ideally.

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,207 @@
...
+    return _apply_matrix_kernel(kernelfct, X, Y, Cref, reg)
+
+
+def kernel_logeuclid(X, Y=None, Cref=None, reg=1e-10):
+    r""" Calculates the Log-Euclidean Kernel matrix K of inner products of two
+    sets X and Y of SPD matrices on tangent space of C by calculating pairwise
+
+    .. math::
+        K_{i,j} =
+        {tr}((\log(X_i) - \log(P))(\log(Y_j) - \log(P)))
```
qbarthelemy, line 162: Same remark.

pyriemann/embedding.py
```diff
@@ -92,13 +111,293 @@ def fit_transform(self, X, y=None):
...
+    embd_, error_ = eigen_vectors[:, index], np.sum(eigen_values)
+
+    return embd_, error_
+
+
+def _check_dimensions(X, Y=None, n_components=None, n_neighbors=None):
+    n_matrices, n_channels, n_channels = X.shape
+
+    if not isinstance(Y, type(None)):
+        assert Y.shape[1:] == (n_channels, n_channels), "Dimension of " \
```
qbarthelemy, line 390: A general remark: in your code, can you complete error messages, with the numerical values?

Example:
`Y does not have the good number of channels. Should be %d but got %d.' % (n_channels, Y.shape[1])`

pyriemann/embedding.py
```diff
@@ -1,12 +1,26 @@
...
+@deprecated(
+    "Embedding is deprecated and will be removed in 0.4.0; "
+    "please use SpectralEmbedding."
+)
 class Embedding(BaseEstimator):
+    pass
+
+
+class SpectralEmbedding(BaseEstimator):
     """Embed SPD matrices into an Euclidean space of smaller dimension.
```
qbarthelemy, line 24:
```suggestion
    """Spectral embedding of SPD matrices into an Euclidean space of smaller dimension.
```

pyriemann/embedding.py
```diff
@@ -92,13 +111,293 @@ def fit_transform(self, X, y=None):
...
+                             n_neighbors=5,
+                             metric='riemann',
+                             reg=1e-3):
+    """Riemannian Locally Linear Embedding (LLE).
+
+    Riemannian LLE as proposed in [1]_. LLE is a non-linear, neighborhood-
+    preserving dimensionality reduction algorithm which consists of three
+    main steps:
+    For each point x,
+    1.  find its k nearest neighbors KNN(x) and
```
qbarthelemy, line 325: Like scikit-learn, docstring of class `LocallyLinearEmbedding` must be different from function `locally_linear_embedding`.
```suggestion
    """Perform a Locally Linear Embedding (LLE) of SPD matrices. 

    As proposed in [1]_, Locally Linear Embedding (LLE) is a non-linear, neighborhood-
    preserving dimensionality reduction algorithm which consists of three
    main steps. For each point x,

    1.  find its k nearest neighbors KNN(x) and
```

pyriemann/embedding.py
```diff
@@ -92,13 +111,293 @@ def fit_transform(self, X, y=None):
...
         return self.embedding_
+
+
+class LocallyLinearEmbedding(BaseEstimator, TransformerMixin):
+    """Riemannian Locally Linear Embedding (LLE).
+
+    Riemannian LLE as proposed in [1]_. LLE is a non-linear,
+    neighborhood-preserving dimensionality reduction algorithm which
+    consists of three main steps:
+    For each point x,
```
qbarthelemy, line 134:
```suggestion
    """Locally Linear Embedding (LLE) of SPD matrices. 

    As proposed in [1]_, Locally Linear Embedding (LLE) is a non-linear,
    neighborhood-preserving dimensionality reduction algorithm which
    consists of three main steps. For each point x,

```

qbarthelemy (changes requested):
Adding these three kernels is a great improvement!

examples/ERP/plot_embedding_MEG.py
plcrodrigues: If I understood correcly, @agramfort is saying that the new function that you've created `plot_embedding` should be defined here in the example script and not as an independent one to be called from the public API. I agree with him.

gabelstein: Ah, but the function already existed before, I just adapted it so it could plot both embeddings.

plcrodrigues (commented):
Thanks for your work, @gabelstein ! It's very nice to see a new embedding technique being added to `pyriemann`, specially one that is so well known in the literature. I'm also a big fan of your new `kernels` module, that will definitely be useful for other methods.

Oh and I'm sorry for embarking so late in this review ^^'

pyriemann/utils/kernel.py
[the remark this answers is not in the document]
qbarthelemy: Have a look to equations 7 and 10 of 
https://hal.archives-ouvertes.fr/file/index/docid/820475/filename/BARACHANT_Neurocomputing_ForHal.pdf

This article should be added as reference for Riemannian and LogEuclidean kernels.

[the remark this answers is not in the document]
gabelstein: For the Riemannian kernel I can do that. 
My log-euclidean kernel is different, though. And I think more appropriate, as it takes as a basis the log-euclidean geodesic and goes from there. The log-euclidean kernel in the paper ist just a specific version of the Riemann kernel with C_ref=I. I can add that as another additional kernel, but would like to stick with my implementation of the log-euclidean kernel to keep it persistent with the geodesic.

tests/test_embedding.py
```diff
@@ -1,24 +1,69 @@
 from conftest import get_metrics
-from pyriemann.embedding import Embedding
+import numpy as np
+from pyriemann.embedding import (SpectralEmbedding,
+                                 LocallyLinearEmbedding,
+                                 barycenter_weights,
+                                 riemann_lle)
 import pytest
```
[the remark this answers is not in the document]
gabelstein: Hmja, I wrote up something pretty dumb which would at least indicate if the embeddings didn't work at all:
embed separated blobs in one dimension and do a nearest centroid classification where training and test data are the same.
But it could be that for some random samples this classification fails and then there's an error in the tests that's not really an error. This should be mitigated by choosing a very high class separation but one can neverr be sure.

examples/ERP/plot_embedding_MEG.py
qbarthelemy: I agree with @gabelstein : this function has been added in `viz` module in pyRiemann 0.2.6, and can not be deleted like that.

pyriemann/utils/kernel.py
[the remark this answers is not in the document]
qbarthelemy: In Arsigny's papers, Log-Euclidean metric is defined as the (affine-invariant) Riemannian metric with Cref=Id.

I agree that it is always possible to add parameters to the equation, but it is not the canonical Log-Euclidean kernel anymore.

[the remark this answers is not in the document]
gabelstein: I beg to differ. I'd say that my kernel is indeed the true canonical Log-Euclidean kernel as it is derived from the log-euclidean geodesic. 
The Log-Euclidean kernel in the paper is just a specific version of the Riemannian kernel which derived from the Riemannian geodesic. It is just a very unfortunate choice of naming.
I can provide the derivation if needed, I'll just have to write it up so it's readable (can do next week).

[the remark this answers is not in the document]
qbarthelemy: Looking forward to your derivation!
Once again: ok for me to use a custom kernel in an example.
But, functions in pyRiemann modules must be canonical kernels, supported by published references.

qbarthelemy (changes requested):
Doc build fails. There are several errors during documentation generation:
https://github.com/pyRiemann/pyRiemann/runs/4827408992?check_suite_focus=true#step:4:606

gabelstein:
Here's the derivation for the kernels.
[kernel_derivation.pdf](https://github.com/pyRiemann/pyRiemann/files/7888556/kernel_derivation.pdf)

qbarthelemy (changes requested):
Doc generation is OK now.

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,219 @@
...
+    return K
+
+
+def kernel_riemann(X, Y=None, Cref=None, reg=1e-10):
+    r""" Calculates the Riemannian Kernel matrix K of inner products of two
+    sets X and Y of SPD matrices on tangent space of C by calculating pairwise
+
+    .. math::
+        K_{i,j} =
+        {tr}(\log(C^{-1/2}X_i C^{-1/2})\log(C^{-1/2}Y_j C^{-1/2}))
```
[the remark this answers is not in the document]
gabelstein: how can I see the doc build for this particular commit (and figure out what the mistake might be)?

qbarthelemy:
> Here's the derivation for the kernels

Thanks for this useful draft!
In Section 2, you use the Riemannian LogMap with the Riemannian scalar product, retrieving the Riemannian kernel of Barachant2013. OK.
Then in Section 3, you use the Log-Euclidean LogMap with the Riemannian scalar product, providing an hybrid kernel.
In the PR, you use the Euclidean LogMap with the Riemannian scalar product, providing an hybrid kernel too.

Indeed, if you use the Log-Euclidean LogMap with the Log-Euclidean scalar product,
you will retrieve the Log-Euclidean kernel of Barachant2013.
Similarly, if you use the Euclidean LogMap with the Euclidean scalar product,
you will retrieve the canonical linear kernel well-used in kernel methods.

IMO, your kernels are not false, but they are hybrid / unusual / non canonical versions,
because it must be the same metric which defines LogMap and scalar product.

pyriemann/utils/kernel.py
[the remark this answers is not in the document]
qbarthelemy: In the "Checks" tab, you will see on the top right part "Artifacts", with a down arrow: click on it and download the doc.

pyriemann/embedding.py
```diff
@@ -92,13 +112,292 @@ def fit_transform(self, X, y=None):
...
+    M = (W.T * W - W.T - W).toarray()
+    M.flat[:: M.shape[0] + 1] += 1
+
+    eigen_values, eigen_vectors = eigh(
+        M, eigvals=(1, n_components), overwrite_a=True
+    )
+    index = np.argsort(np.abs(eigen_values))
+    embd_, error_ = eigen_vectors[:, index], np.sum(eigen_values)
+
+    return embd_, error_
```
qbarthelemy, line 383:
```suggestion
    embd, error = eigen_vectors[:, index], np.sum(eigen_values)

    return embd, error
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,231 @@
...
+    Calculates the kernel matrix K of inner products of two sets
+         X and Y of SPD matrices on tangent space of C according to the
+         specified metric.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            First set of SPD matrices.
+        Y : None | ndarray, shape (n_matrices, n_channels, n_channels), \
+            default: None
```
qbarthelemy, line 19:
```suggestion
                default: None
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,231 @@
...
+from .mean import mean_riemann, mean_euclid, mean_logeuclid
+from .base import invsqrtm, logm
+
+
+def kernel(X, Y=None, Cref=None, metric='riemann', reg=1e-10):
+    r"""  Calculates a kernel matrix according to a specified metric
+
+    Calculates the kernel matrix K of inner products of two sets
+         X and Y of SPD matrices on tangent space of C according to the
+         specified metric.
```
qbarthelemy, line 12:
```suggestion
     X and Y of SPD matrices on tangent space of C according to the
     specified metric.
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,231 @@
...
+        {tr}(\log(C^{-1/2}X_i C^{-1/2})\log(C^{-1/2}Y_j C^{-1/2}))
+
+    as proposed in [1]_.
+
+    Parameters
+    ----------
+    X : ndarray, shape (n_matrices_X, n_channels, n_channels)
+        First set of SPD matrices.
+    Y : None | ndarray, shape (n_matrices_Y, n_channels, n_channels), \
+        default: None
```
qbarthelemy, line 92:
```suggestion
            default: None
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,231 @@
...
+    .. math::
+        K_{i,j} =
+        {tr}(C^{-1/2}(X_i - C)C^{-1/2} C^{-1/2}(Y_j - C)C^{-1/2})
+
+    Parameters
+    ----------
+    X : ndarray, shape (n_matrices_X, n_channels, n_channels)
+        First set of SPD matrices.
+    Y : None | ndarray, shape (n_matrices_Y, n_channels, n_channels), \
+        default: None
```
qbarthelemy, line 143:
```suggestion
            default: None
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,231 @@
...
+    .. math::
+        K_{i,j} =
+        {tr}((\log(X_i) - \log(C))(\log(Y_j) - \log(C)))
+
+    Parameters
+    ----------
+    X : ndarray, shape (n_matrices_X, n_channels, n_channels)
+        First set of SPD matrices.
+    Y : None | ndarray, shape (n_matrices_Y, n_channels, n_channels), \
+        default: None
```
qbarthelemy, line 188:
```suggestion
            default: None
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,231 @@
...
+        return X_
+
+    return _apply_matrix_kernel(kernelfct, X, Y, Cref, reg)
+
+
+def kernel_euclid(X, Y=None, Cref=None, reg=1e-10):
+    r"""  Calculates Euclidean kernel
+
+    Calculates the Euclidean kernel matrix K of inner products of two sets
+     X and Y of SPD matrices on tangent space of C by calculating pairwise
```
qbarthelemy, line 132:
```suggestion
    X and Y of SPD matrices on tangent space of C by calculating pairwise
```

qbarthelemy (changes requested):

pyriemann/utils/kernel.py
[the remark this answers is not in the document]
gabelstein: Ah, thanks! It seems to be a problem with the sphinx configuration. None of the formulas (e.g. in geodesics) are rendered although they are written correctly (as far as I can tell). I know nothing about sphinx, but apparently some package needs to be added to the configuration (https://documentation.help/Sphinx/math.html).

gabelstein:
> Indeed, if you use the Log-Euclidean LogMap with the Log-Euclidean scalar product, you will retrieve the Log-Euclidean kernel of Barachant2013. Similarly, if you use the Euclidean LogMap with the Euclidean scalar product, you will retrieve the canonical linear kernel well-used in kernel methods.


But how does the log-euclidean scalar product look? I haven't found that anywhere. But we can discuss this matter at another time, I'll rewrite my implementation.

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,225 @@
...
+
+        C_invsq = invsqrtm(Cref)
+        X_ = C_invsq @ X @ C_invsq
+        X_ = np.array([logm(x_) for x_ in X_])
+        return X_
+
+    return _apply_matrix_kernel(kernelfct, X, Y, Cref, reg)
+
+
+def kernel_euclid(X, Y=None, Cref=None, reg=1e-10):
```
qbarthelemy, line 127: `Cref` can be deleted from Euclidean and LogEuclidean kernels (parameters and doc).

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,225 @@
...
+    K : ndarray, shape (n_matrices_X, n_matrices_Y)
+        The kernel matrix of X and Y.
+
+    Notes
+    -----
+    .. versionadded:: 0.2.8
+
+    References
+    ----------
+    .. [1] A. Barachant, S. Bonnet, M. Congedo and C. Jutten, Classification
```
qbarthelemy, line 163: This reference should be added to Log-Euclidean, instead of Euclidean.

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,225 @@
...
+def kernel(X, Y=None, Cref=None, metric='riemann', reg=1e-10):
+    r"""  Calculates a kernel matrix according to a specified metric
+
+    Calculates the kernel matrix K of inner products of two sets
+     X and Y of SPD matrices on tangent space of C according to the
+     specified metric.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
```
qbarthelemy, line 16:
```suggestion
        X : ndarray, shape (n_matrices_X, n_channels, n_channels)
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,225 @@
...
+            First set of SPD matrices.
+        Y : None | ndarray, shape (n_matrices, n_channels, n_channels), \
```
qbarthelemy, line 18:
```suggestion
        Y : None | ndarray, shape (n_matrices_Y, n_channels, n_channels), \
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,225 @@
...
+            X according to the specified metric.
+        metric : {'riemann', 'euclid', 'logeuclid'}, default: 'riemann'
+            The type of metric used for tangent space and mean estimation.
+        reg : float, default: 1e-10
+            Regularization parameter to mitigate numerical errors in kernel
+            matrix estimation, to provide a positive-definite kernel matrix.
+
+        Returns
+        ----------
+        K : ndarray, shape (n_matrices, n_matrices)
```
qbarthelemy, line 33:
```suggestion
        K : ndarray, shape (n_matrices_X, n_matrices_Y)
```

pyriemann/utils/kernel.py
```diff
@@ -0,0 +1,225 @@
...
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            First set of SPD matrices.
+        Y : None | ndarray, shape (n_matrices, n_channels, n_channels), \
+                default: None
+            Second set of SPD matrices. If None, Y is set to X.
+        Cref : None | ndarray, shape (n_channels, n_channels), default: None
+            Reference point for the tangent space and inner product
+            calculation. If None, Cref is calculated as the mean of
+            X according to the specified metric.
```
qbarthelemy, line 24: Can you indicate that this parameter is only for Riemannian kernel?

pyriemann/utils/kernel.py
gabelstein: Shouldn't they be kept for compatibility purposes of calling the generic kernel function? I can fix the doc so it states that it does nothing.

qbarthelemy: IMO, for LE and E kernels, you should remove Cref from parameters, and then you return `_apply_matrix_kernel(kernelfct, X, Y, None, reg)`.

pyriemann/utils/kernel.py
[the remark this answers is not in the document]
qbarthelemy: It's ok for me.

pyriemann/utils/kernel.py
gabelstein:
```
def kernel(X, Y=None, Cref=None, metric='riemann', reg=1e-10):
    try:
        return globals()[f'kernel_{metric}'](X, Y, Cref, reg=reg)
    except KeyError:
        raise ValueError("Kernel metric must be 'riemann', 'euclid' or "
                         "'logeuclid'.")
```

it'll break this, though. And once we figure out the proper kernels exactly (I'm still very much in doubt about the current implementations of euclidean and log-euclidean kernels), we might need them again.

pyriemann/embedding.py
```diff
@@ -92,13 +112,292 @@ def fit_transform(self, X, y=None):
...
+    data_ : ndarray, shape (n_matrices, n_channels, n_channels)
+        Training data.
+
+    Notes
+    -----
+    .. versionadded:: 0.2.8
+
+    References
+    ----------
+    .. [1] A. Goh and R. Vidal, "Clustering and dimensionality reduction
```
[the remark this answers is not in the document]
gabelstein: good point.

qbarthelemy (approved):
Only minor points to correct.
Otherwise, it's ok for me.
Great contribution, thx @gabelstein !

plcrodrigues (commented):
Thank you very much @gabelstein for this contribution and for having included some of my comments from my previous review. I am pretty much satisfied with this PR and the nice addition that it makes to `pyriemann`.

If I was going to suggest one minor thing, would be to have an example where we would see how the Riemannian kernel that you've implemented is useful and how the results change (or not) when we choose one metric over the other. Leaving the PR as it is, one might have the impression that your Riemannian kernel object was created only to work with LLE, but in fact it is a whole contribution by itself that can be useful in other situations.

plcrodrigues (approved):

gabelstein:
> If I was going to suggest one minor thing, would be to have an example where we would see how the Riemannian kernel that you've implemented is useful and how the results change (or not) when we choose one metric over the other. Leaving the PR as it is, one might have the impression that your Riemannian kernel object was created only to work with LLE, but in fact it is a whole contribution by itself that can be useful in other situations.

My next PR will the the riemannian support vector machine, there it is easier to do a kernel comparison.

Issue #151, opened 2021-12-02
Python 2 support with pyriemann>=0.2.7


jliersch:
Hello!

According to the setup.cfg, the package in version 0.2.7 is still intended to work with python 2:

```
[bdist_wheel]
[...]
universal = 1
```

Unfortunately the installation with pip fails for me with python 2.7.18 due to this error:

```
ERROR: Command errored out with exit status 1:
     command: <secret path>/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-jClpfU/pyriemann/setup.py'"'"'; __file__='"'"'/tmp/pip-install-jClpfU/pyriemann/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-jClpfU/pyriemann/pip-egg-info
         cwd: /tmp/pip-install-jClpfU/pyriemann/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-jClpfU/pyriemann/setup.py", line 16, in <module>
        with open('README.md', 'r', encoding="utf8") as fid:
    TypeError: 'encoding' is an invalid keyword argument for this function
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

It's fine for me, if the support for python 2 is dropped with 0.2.7. But I think this should be made consistent.

sylvchev:
Thanks, the support for python 2.7 is dropped since its deprecation. Could you open a PR to push this modification?

jliersch:
I assume there is a way to provide an expressive error message when trying to install pyriemann with python < 3 and think adding such a message is the right thing to do if the support is dropped. Unfortnuately I don't know the best practice how to add such a message yet, but will look into it by the end of the year.

jliersch:
I looked a little deeper into it which resulted in pull request #160, which from my perspective resolves the issue.

qbarthelemy:
Closed by #160.

PR #160, opened 2022-01-04
Enabled opening README.md with python2 in setup.py

jliersch:
In setup.py, the required python version is set to >=3.6 which would produce an error message telling someone wanting to install pyRiemann with a lower version this version mismatch. But with python < 3 the handling of this version check will not be reached. Instead setup.py fails at calling `open('README.md', 'r', encoding="utf8")`, since the parameter "encoding" was added to "open" with python3.

I propose to handle the difference in signature of "open" to provide a more precise error message if someone trys to install pyRiemann with python < 3.


agramfort:
I think we should just drop the support for python < 3.6

jliersch:
Sorry, I missed adding the corresponding issue #151.

I concur. But I also think that it should made easy for the users to find out that the python 2 support is dropped. This is usually is done by the error message about the python version mismatch (see below). But this message currently can't be reached if one tries to install pyRiemann with python < 3.

Expected error message:
`ERROR: Package 'pyriemann' requires a different Python: 2.7.18 not in '>=3.6'`

Current error message:
`ERROR: Command errored out with exit status 1:
     command: <secret path>/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-jClpfU/pyriemann/setup.py'"'"'; __file__='"'"'/tmp/pip-install-jClpfU/pyriemann/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-jClpfU/pyriemann/pip-egg-info
         cwd: /tmp/pip-install-jClpfU/pyriemann/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-jClpfU/pyriemann/setup.py", line 16, in <module>
        with open('README.md', 'r', encoding="utf8") as fid:
    TypeError: 'encoding' is an invalid keyword argument for this function
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.`

sylvchev:
The error is easier to understand like that. Ok to merge for me.

setup.py
```diff
@@ -13,7 +14,10 @@
 if version is None:
     raise RuntimeError("Could not determine version")
 
-with open("README.md", "r", encoding="utf8") as fid:
+# "open" does not provide the parameter "encoding" for python<3
```
qbarthelemy, line 17:
```suggestion
```

qbarthelemy (approved):

qbarthelemy:
Thx @jliersch !

PR #161, opened 2022-01-19
Display animated figures in the documentation

qbarthelemy:
#### Problem

Currently, animations on artifacts detection are not plotted by the doc.
See: https://github.com/pyRiemann/pyRiemann/pull/133#issuecomment-890333261 and https://github.com/pyRiemann/pyRiemann/pull/142#issuecomment-932163081 .

Warning during doc generation:
```
UserWarning: Animation was deleted without rendering anything. This is most likely not intended.
To prevent deletion, assign the Animation to a variable, e.g. `anim`, that exists until you have 
outputted the Animation using `plt.show()` or `anim.save()`
```

#### Work in progress:
- [ ] example on RP
- [ ] example on RPF

#### Do not merge now: it does not work.

PR #162, opened 2022-01-24
Display animated figures in the documentation, 2nd part

qbarthelemy:
This PR completes the too quickly merged PR #161.

I have not found how to avoid figure to be plotted twice in the documentation.
However, I think that it is better to have a double plot, with the second one animated, rather than nothing.

### DIFF
--- a/examples/artifacts/plot_detect_riemannian_potato_EEG.py
+++ b/examples/artifacts/plot_detect_riemannian_potato_EEG.py
@@ -252,6 +252,7 @@ def online_detect(t):
     p_ep.set_color(ep_colors_)
     return pl_sig0, pl_sig1, p_rp, p_ep
 
+
 interval_display = 1.0  # can be changed for a slower display
 
 potato = FuncAnimation(fig, online_detect,
--- a/examples/artifacts/plot_detect_riemannian_potato_EEG.py
+++ b/examples/artifacts/plot_detect_riemannian_potato_EEG.py
@@ -19,7 +19,6 @@ import numpy as np
 from matplotlib import pyplot as plt
 from matplotlib.colors import to_rgb
 from matplotlib.animation import FuncAnimation
-from IPython.display import HTML
 
 from mne.datasets import sample
 from mne.io import read_raw_fif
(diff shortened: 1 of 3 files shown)


examples/artifacts/plot_detect_riemannian_potato_EEG.py
```diff
@@ -19,6 +19,7 @@
 from matplotlib import pyplot as plt
 from matplotlib.colors import to_rgb
 from matplotlib.animation import FuncAnimation
+from IPython.display import HTML
```
agramfort, line 22: I would not make a hard dep on IPython on examples. It should not be necessary for sphinx gallery

mne has animations but no import of IPython

See for example this page: https://mne.tools/stable/auto_examples/visualization/evoked_topomap.html?highlight=animation#animating-the-topomap

qbarthelemy: Ok, I see. It requires a consequent refactoring...

qbarthelemy: Why IPython can not be added only in dependencies for documentation? @agramfort

qbarthelemy: What about this modification?
```
try:
    from IPython.display import HTML
    plt.rcParams["animation.embed_limit"] = 10
    HTML(potato.to_jshtml(fps=5, default_mode='loop'))
except ImportError:
    raise ImportError("Install IPython to plot animation in documentation")
```

agramfort: my reasoning is that ipython is not a small dependency and if we can avoid it it's better. But I also hear that it's more work... PyRiemann is a super light package with minimal deps and I think we should fight for this. But here it's indeed just for the doc...

qbarthelemy: Ok, you can merge if you are OK.

agramfort (approved):
let's move on here. Ideally doing just this 

https://sphinx-gallery.github.io/stable/auto_examples/plot_8_animations.html

should work but ...

PR #163, opened 2022-03-17
Correct warnings during doc generation

qbarthelemy:
Correct warnings during doc generation for embedding and kernel modules.


agramfort (approved):

agramfort:
thx @qbarthelemy

PR #164, opened 2022-03-22
Riemann Support Vector Machine

gabelstein:
Making use of the new kernel module, the new support vector machine classification.

Should I extend some example code somewhere to compare e.g. MDM with the RSVC?

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -47,6 +47,7 @@ Classification
     FgMDM
     TSclassifier
     KNearestNeighbor
+    RSVC
 
 Clustering
 ------------------
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -16,6 +16,7 @@ Covariance Estimation
     Covariances
     ERPCovariances
     XdawnCovariances
+    BlockCovariances
     CospCovariances
     Coherences
     HankelCovariances
@@ -153,6 +154,7 @@ Covariance preprocessing
     covariances
     covariances_EP
     covariances_X
+    block_covariances
     cross_spectrum
     cospectrum
     coherence
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -29,10 +29,14 @@ Embedding
 
 .. autosummary::
     :toctree: generated/
+
+    locally_linear_embedding
+    barycenter_weights
+
     :template: class.rst
 
-    Embedding
-    RiemannianLLE
+    SpectralEmbedding
+    LocallyLinearEmbedding
 
 Classification
 --------------
@@ -190,6 +194,8 @@ Kernels
 
     kernel
     kernel_riemann
+    kernel_euclid
+    kernel_logeuclid
 
 Mean
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -32,6 +32,7 @@ Embedding
     :template: class.rst
 
     Embedding
+    RiemannianLLE
 
 Classification
 --------------
@@ -177,6 +178,17 @@ Distances
     distance_kullback
     distance_kullback_sym
     distance_wasserstein
+
+Kernels
+~~~~~~~~~~~~~~~~~~~~~~
+.. _kernel_api:
+.. currentmodule:: pyriemann.utils.kernel
+
+.. autosummary::
+    :toctree: generated/
+
+    kernel
+    kernel_riemann
 
 Mean
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -51,7 +51,7 @@ Classification
     FgMDM
     TSclassifier
     KNearestNeighbor
-    RSVC
+    SupportVectorMachine
 
 Clustering
 ------------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -40,11 +40,11 @@ v0.2.8.dev
 
 - Add block covariance matrix: :class:`pyriemann.estimation.BlockCovariances` and :func:`pyriemann.utils.covariance.block_covariances`
 
-- Add Riemannian Locally Linear Embedding: :class:`pyriemann.embedding.RiemannLLE` and :func:`pyriemann.embedding.riemann_lle`
+- Add Riemannian Locally Linear Embedding: :class:`pyriemann.embedding.LocallyLinearEmbedding` and :func:`pyriemann.embedding.locally_linear_embedding`
 
 - Add Riemannian Kernel Function: :func:`pyriemann.utils.kernel.kernel_riemann`
 
-- Add Riemannian Support Vector Machine: :func:`pyriemann.classification.RSVC`
+- Add Riemannian Support Vector Machine: :class:`pyriemann.classification.SupportVectorMachine`
 
 
 v0.2.7 (June 2021)
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -44,6 +44,8 @@ v0.2.8.dev
 
 - Add Riemannian Kernel Function: :func:`pyriemann.utils.kernel.kernel_riemann`
 
+- Add Riemannian Support Vector Machine: :func:`pyriemann.classification.RSVC`
+
 
 v0.2.7 (June 2021)
 ------------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -38,6 +38,8 @@ v0.2.8.dev
 
 - Add ``corr`` option in :func:`pyriemann.utils.covariance.normalize`, to normalize covariance into correlation matrices
 
+- Add block covariance matrix: :class:`pyriemann.estimation.BlockCovariances` and :func:`pyriemann.utils.covariance.block_covariances`
+
 v0.2.7 (June 2021)
 ------------------
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -40,6 +40,11 @@ v0.2.8.dev
 
 - Add block covariance matrix: :class:`pyriemann.estimation.BlockCovariances` and :func:`pyriemann.utils.covariance.block_covariances`
 
+- Add Riemannian Locally Linear Embedding: :class:`pyriemann.embedding.RiemannLLE` and :func:`pyriemann.embedding.riemann_lle`
+
+- Add Riemannian Kernel Function: :func:`pyriemann.utils.kernel.kernel_riemann`
+
+
 v0.2.7 (June 2021)
 ------------------
 
--- a/pyriemann/classification.py
+++ b/pyriemann/classification.py
@@ -518,17 +518,17 @@ class KNearestNeighbor(MDM):
         return out.ravel()
 
 
-class RSVC(BaseEstimator, ClassifierMixin):
+class SupportVectorMachine(BaseEstimator, ClassifierMixin):
     """Classification by Riemannian Support Vector Machine.
 
     Support vector machine with precomputed Riemannian kernel matrix
-    according to the metric as described in [1]_.
+    according to different metrics as described in [1]_.
 
     Parameters
     ----------
     svc_func : SVC class, (default: sklearn.svm.SVC)
         Class to perform the svc calculations.
-    metric : {'riemann'}
+    metric : {'riemann', 'euclid', 'logeuclid'}
         Metric for kernel matrix computation.
     Cref : None | ndarray, shape (n_channels, n_channels)
         Reference point for kernel matrix computation. If None, the mean of
@@ -583,7 +583,7 @@ class RSVC(BaseEstimator, ClassifierMixin):
         Returns
         -------
         self : Riemannian SVC instance
-            The RSVC instance.
+            The SVC instance.
         """
         kernelmat = kernel(X, X, Cref=self.Cref, metric=self.metric)
         self.data_ = X
@@ -602,7 +602,7 @@ class RSVC(BaseEstimator, ClassifierMixin):
         Returns
         -------
         pred : ndarray of int, shape (n_matrices, 1)
-            the prediction for each trials according to the RSVC.
+            the prediction for each trials according to the SVC.
         """
         test_kernel_mat = kernel(self.data_,
                                  X_test,
--- a/pyriemann/classification.py
+++ b/pyriemann/classification.py
@@ -500,7 +500,7 @@ class KNearestNeighbor(MDM):
         return self
 
     def predict(self, covtest):
-        """get the predictions.
+        """Get the predictions.
 
         Parameters
         ----------
@@ -564,6 +564,7 @@ class SupportVectorMachine(BaseEstimator, ClassifierMixin):
         self.svc_ = self.svc_func(kernel='precomputed', **self.svc_params)
 
     def __setattr__(self, name, value):
+        """Enable setting attributes for SVC subclass."""
         if 'svc_' in self.__dict__.keys():
             if name in self.svc_.__dict__.keys():
                 setattr(self.svc_, name, value)
--- a/pyriemann/classification.py
+++ b/pyriemann/classification.py
@@ -526,8 +526,8 @@ class SupportVectorMachine(BaseEstimator, ClassifierMixin):
 
     Parameters
     ----------
-    svc_func : SVC class, (default: sklearn.svm.SVC)
-        Class to perform the svc calculations.
+    svc_func : class, (default: sklearn.svm.SVC)
+        Class that performs the svc calculations.
     metric : {'riemann', 'euclid', 'logeuclid'}
         Metric for kernel matrix computation.
     Cref : None | ndarray, shape (n_channels, n_channels)
@@ -585,7 +585,7 @@ class SupportVectorMachine(BaseEstimator, ClassifierMixin):
         self : Riemannian SVC instance
             The SVC instance.
         """
-        kernelmat = kernel(X, X, Cref=self.Cref, metric=self.metric)
+        kernelmat = kernel(X, Cref=self.Cref, metric=self.metric)
         self.data_ = X
         self.svc_ = self.svc_.fit(kernelmat, y)
 
@@ -604,8 +604,8 @@ class SupportVectorMachine(BaseEstimator, ClassifierMixin):
         pred : ndarray of int, shape (n_matrices, 1)
             the prediction for each trial according to the SVC.
         """
-        test_kernel_mat = kernel(self.data_,
-                                 X,
+        test_kernel_mat = kernel(X,
+                                 self.data_,
                                  Cref=self.Cref,
                                  metric=self.metric)
         return self.svc_.predict(test_kernel_mat)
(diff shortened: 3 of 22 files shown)


pyriemann/classification.py
```diff
@@ -498,19 +500,113 @@ def fit(self, X, y):
...
+        pred : ndarray of int, shape (n_matrices, 1)
             the prediction for each trials according to the closest centroid.
         """
         dist = self._predict_distances(covtest)
         neighbors_classes = self.classes_[np.argsort(dist)]
         out, _ = stats.mode(neighbors_classes[:, 0:self.n_neighbors], axis=1)
         return out.ravel()
+
+
+class SupportVectorMachine(BaseEstimator, ClassifierMixin):
```
agramfort, line 521: I would create RiemannSVC and RiemannSVR classes

pyriemann/classification.py
```diff
@@ -498,19 +500,113 @@ def fit(self, X, y):
...
+    .. versionadded:: 0.2.8
+
+    References
+    ----------
+    .. [1] A. Barachant, S. Bonnet, M. Congedo, and C. Jutten.
+        Classification of covariance matrices using a Riemannian-based kernel
+        for BCI applications". In: Neurocomputing 112 (July 2013), pp. 172-178.
+    """
+
+    def __init__(self, svc_func=SVC, metric='riemann', Cref=None,
```
agramfort, line 557: SVC is not a func but a "klass" or "factory". In any case I would not expose this flexibility have write 2 new estimators (one for classif and one for regression.

pyriemann/classification.py
```diff
@@ -498,19 +500,113 @@ def fit(self, X, y):
...
+        self.svc_params = kwargs
+        self.svc_ = self.svc_func(kernel='precomputed', **self.svc_params)
+
+    def __setattr__(self, name, value):
+        """Enable setting attributes for SVC subclass."""
+        if 'svc_' in self.__dict__.keys():
+            if name in self.svc_.__dict__.keys():
+                setattr(self.svc_, name, value)
+                return
+        super().__setattr__(name, value)
```
agramfort, line 572: you should rely on the set_params method from sklearn estimators. THis is a bit error prone.

tests/test_classification.py
```diff
@@ -194,3 +203,29 @@ def test_TSclassifier_classifier_error():
     """Test TS if not Classifier"""
     with pytest.raises(TypeError):
         TSclassifier(clf=Covariances())
+
+
+def test_SupportVectorMachine_SVC_params():
```
agramfort, line 208: please stick to lower case function names 🙏

pyriemann/classification.py
gabelstein: oh, didn't know about that. Will change.

pyriemann/classification.py
gabelstein: I implemented it this way to be able to use GPU accelerated SVC calculation (like thundersvm).

pyriemann/classification.py
gabelstein: Good point. I don't like regression with SVMs but it should still be implemented.

pyriemann/classification.py
gabelstein: at least for the second part this doesn't work, because it also uses this method to set the parameters in the init.

pyriemann/classification.py
gabelstein: Although I'm unsure if precomputed kernels can be accelerated much by GPU.

sylvchev: I'll ask for a confirmation, but from what I remember, there no specific gain with GPU for precomputed kernels.

pyriemann/classification.py
sylvchev: Is this still a problem if you separate SVC and SVR class?

sylvchev:
Will you implement Stein kernel or normalization? There are some nice example in this paper: https://doi.org/10.1109/MLSP.2013.6661972 I could send you the PDF if you want.

pyriemann/classification.py
gabelstein: Yes, that doesn't change anything unfortunately.

gabelstein:
This is not planned right now. I'm currently reviewing kernel methods with Quentin, so afterwards we might implement a few more.

pyriemann/classification.py
qbarthelemy: If a class `RiemannSVR` is added, it should be placed in a `regression` module, rather than `classification`.

gabelstein:
While at it, I added a very simple KNN regression that I had flying around in some old code. Obviously only works for interpolation, can't extrapolate anything.
Still lacking tests and (if needed) some example. Any ideas on which one to extend?

qbarthelemy (changes requested):
Thx for this contribution! Yes, you can add a new example in the gallery on a MI task, to compare RiemannSVC to other classical classifiers.

pyriemann/classification.py
```diff
@@ -498,19 +497,111 @@ def fit(self, X, y):
...
+        Returns
+        -------
+        pred : ndarray of int, shape (n_matrices, 1)
+            Predictions for each matrix according to the SVC.
+        """
+        test_kernel_mat = kernel(X,
+                                 self.data_,
+                                 Cref=self.Cref,
+                                 metric=self.metric)
+        return self.svc_.predict(test_kernel_mat)
```
qbarthelemy, line 607: Is it possible to add the function `predict_proba`?

doc/api.rst
```diff
@@ -51,6 +51,19 @@ Classification
...
+--------------
+.. _regression_api:
+.. currentmodule:: pyriemann.regression
+
+.. autosummary::
+    :toctree: generated/
+    :template: class.rst
+
+    SVR
+    KNNRegression
```
qbarthelemy, line 66: The name of the regressor (currently, KNNRegression) should be homogeneous with the classifier one (KNearestNeighbor). 

Suggestion 1 (requiring a deprecation):
KNearestNeighbor -> KNNClassifier
KNNRegression -> KNNRegressor

Suggestion 2:
KNearestNeighbor
KNNRegression -> KNearestNeighborRegressor

Suggestion 3 (use sklearn names, requiring a deprecation):
KNearestNeighbor -> KNeighborsClassifier
KNNRegression -> KNeighborsRegressor

doc/whatsnew.rst
```diff
@@ -44,6 +44,8 @@ v0.2.8.dev
 
 - Add Riemannian Kernel Function: :func:`pyriemann.utils.kernel.kernel_riemann`
 
+- Add Riemannian Support Vector Machine: :class:`pyriemann.classification.SupportVectorMachine`
```
qbarthelemy, line 47:
```suggestion
- Add Riemannian Support Vector Machine: :class:`pyriemann.classification.SVC` and :class:`pyriemann.regression.SVR`
```

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+        the training data according to the metric is used.
+    **kwargs
+        Keyword arguments passed to svc_func.
+
+    Attributes
+    ----------
+    svc_ : svc_func instance
+        Fitted SVC with precomputed Riemannian kernel matrix.
+    data_ : ndarray, shape (n_matrices, n_channels, n_channels)
+        Training data.
```
qbarthelemy, line 35:
```suggestion
        If fitted, training data.
```

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            Set of SPD matrices.
+        y : ndarray, shape (n_matrices, 1)
+            labels corresponding to each matrix.
+
+        Returns
+        -------
+        self : Riemannian SVC instance
+            The SVC instance.
```
qbarthelemy, line 78:
```suggestion
        self : Riemannian SVR instance
            The SVR instance.
```

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+                                 self.data_,
+                                 Cref=self.Cref,
+                                 metric=self.metric)
+        return self.svr_.predict(test_kernel_mat)
+
+
+class KNNRegression(MDM):
+    """Regression by K-Nearest-Neighbors.
+
+    Reression by nearest Neighbors. For each point of the test set, the
```
qbarthelemy, line 109:
```suggestion
    Regression by nearest Neighbors. For each point of the test set, the
```

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+    pairwise distance to each element of the training set is estimated. The
+    value is calculated according to the softmax average w.r.t. distance of
+    the k nearest neighbors.
```
qbarthelemy, line 112: Can you add a reference for this algorithm?

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Returns
+        -------
+        pred : ndarray of int, shape (n_matrices, 1)
+            Predictions for each matrix according to the closest centroid.
+        """
+        dist = self._predict_distances(covtest)
+        dist_sorted = np.sort(dist)
+        neighbors_values = self.values_[np.argsort(dist)]
+        softmax_dist = softmax(-dist_sorted[:, 0:self.n_neighbors])
```
qbarthelemy, line 171: IMO, it should be the softmax of squared distances (see #100):
```suggestion
        softmax_dist = softmax(-dist_sorted[:, 0:self.n_neighbors]**2)
```

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            Set of SPD matrices.
+
+        Returns
+        -------
+        pred : ndarray of int, shape (n_matrices, 1)
+            Predictions for each matrix according to the SVC.
```
qbarthelemy, line 97:
```suggestion
        pred : ndarray, shape (n_matrices, 1)
            Predictions for each matrix according to the SVR.
```

qbarthelemy (changes requested):
Don't forget to add tests for the new regression module.

pyriemann/regression.py
gabelstein: Uhm, I didn't base it on any reference.

gabelstein: It might be entirely stupid.

pyriemann/classification.py
```diff
@@ -498,19 +498,136 @@ def fit(self, X, y):
...
+        for BCI applications". In: Neurocomputing 112 (July 2013), pp. 172-178.
+    """
+
+    def __init__(self, metric='riemann', Cref=None, **kwargs):
+        """Init."""
+        self.svc_func = sklearnSVC
+        self.Cref = Cref
+        self.metric = metric
+        self.svc_params = kwargs
+        self.svc_ = self.svc_func(kernel='precomputed', **self.svc_params)
```
agramfort, line 559: attributes ending with _ should not appear in the init but in the fit method.

pyriemann/classification.py
```diff
@@ -498,19 +498,136 @@ def fit(self, X, y):
...
+        """Get the predictions.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            Set of SPD matrices.
+
+        Returns
+        -------
+        pred : ndarray of int, shape (n_matrices, 1)
```
agramfort, line 600:
```suggestion
        pred : ndarray of int, shape (n_matrices,)
```
it's what it should be

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+
+    def __init__(self, metric='riemann', Cref=None,
+                 **kwargs):
+        """Init."""
+        self.svr_func = sklearnSVR
+        self.Cref = Cref
+        self.metric = metric
+        self.svr_params = kwargs
+        self.svr_ = self.svr_func(kernel='precomputed', **self.svr_params)
```
agramfort, line 55: same remark. Move it to the fit

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+        """Get the predictions.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            Set of SPD matrices.
+
+        Returns
+        -------
+        pred : ndarray, shape (n_matrices, 1)
```
agramfort, line 96:
```suggestion
        pred : ndarray, shape (n_matrices,)
```

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+    classes_ : list
+        list of classes.
+
+    """
+
+    def __init__(self, n_neighbors=5, metric='riemann'):
+        """Init."""
+        # store params for cloning purpose
+        self.n_neighbors = n_neighbors
+        MDM.__init__(self, metric=metric)
```
agramfort, line 133:
```suggestion
        super().__init__(metric=metric)
```

pyriemann/regression.py
```diff
@@ -0,0 +1,174 @@
...
+        """Get the predictions.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            Set of SPD matrices.
+
+        Returns
+        -------
+        pred : ndarray of int, shape (n_matrices, 1)
```
agramfort, line 165:
```suggestion
        pred : ndarray of int, shape (n_matrices,)
```

pyriemann/classification.py
```diff
@@ -498,19 +498,136 @@ def fit(self, X, y):
...
+        self.svc_func = sklearnSVC
+        self.Cref = Cref
+        self.metric = metric
+        self.svc_params = kwargs
+        self.svc_ = self.svc_func(kernel='precomputed', **self.svc_params)
+
+    def __setattr__(self, name, value):
+        """Enable setting attributes for SVC subclass."""
+        if 'svc_' in self.__dict__.keys():
+            if name in self.svc_.__dict__.keys():
```
agramfort, line 564:
```suggestion
            if name in self.svc_.get_params():
```

doc/api.rst
gabelstein: I opted for 2) as it does't require deprecation.

pyriemann/classification.py
gabelstein: I'd rather just remove the _ than move it to the fit, honestly. Otherwise, the svc_func params cannot be changed before fitting.

gabelstein: I rewrote it a bit so it's cleaner.

pyriemann/regression.py
gabelstein: Done.

pyriemann/classification.py
agramfort: to me hacking __setattr__ is to me a code smell. We should try to avoid it. Maybe you can inherit from SVC and rely on `super()`'s constructor?

gabelstein: Ah yes, I had tried it before and it didn't work. But now it works (had to copy all parameters from the SVC/SVR).

gabelstein:
tests for regression module added. I now very little about proper testing so I'm grateful for any remarks

qbarthelemy (changes requested):
Can you rebase your branch onto the last commit of the master branch of pyRiemann?

pyriemann/regression.py
```diff
@@ -0,0 +1,206 @@
...
+        -------
+        pred : ndarray of int, shape (n_matrices,)
+            Predictions for each matrix according to the closest centroid.
+        """
+        dist = self._predict_distances(X)
+        dist_sorted = np.sort(dist)
+        neighbors_values = self.values_[np.argsort(dist)]
+        softmax_dist = softmax(-dist_sorted[:, 0:self.n_neighbors]**2)
+        knn_values = neighbors_values[:, 0:self.n_neighbors]
+        out = np.sum(knn_values*softmax_dist, axis=1)
```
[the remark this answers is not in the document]
gabelstein: I think with the softmax it has to be the sum.

pyriemann/regression.py
```diff
@@ -0,0 +1,206 @@
...
+        y : ndarray shape (n_matrices, 1)
+            labels corresponding to each matrix.
+
+        Returns
+        -------
+        self : NearestNeighbor instance
+            The NearestNeighbor instance.
+        """
+        self.values_ = y
+        self.covmeans_ = X
```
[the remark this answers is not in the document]
gabelstein: It is used in the self._predict_distances function.

agramfort (approved):
much cleaner ! thx @gabelstein

sylvchev (approved):

qbarthelemy (commented):
Code is OK for me, but new classifieurs / regressors should be compared in examples.
To avoid heavy downloads of real EEG, I suggest to create simulated datasets, using the module `simulated`.

A new example should compare all classifiers available in pyRiemann on this simulated dataset, like in scikit-learn:
https://scikit-learn.org/stable/auto_examples/classification/plot_classifier_comparison.html
And an example for regressors.

pyriemann/classification.py
```diff
@@ -179,19 +181,18 @@ def predict_proba(self, X):
...
-        X : ndarray, shape (n_trials, n_channels, n_channels)
-            ndarray of SPD matrices.
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            Set of SPD matrices.
 
         Returns
         -------
-        prob : ndarray, shape (n_trials, n_classes)
+        prob : ndarray, shape (n_matrices, n_classes)
             the softmax probabilities for each class.
```
qbarthelemy, line 190:
```suggestion
            Probabilities for each class.
```

tests/test_classification.py
```diff
@@ -1,22 +1,24 @@
...
     def test_two_classes(self, classif, get_covmats, get_labels):
-        n_classes, n_trials, n_channels = 2, 6, 3
-        covmats = get_covmats(n_trials, n_channels)
-        labels = get_labels(n_trials, n_classes)
+        n_classes, n_matrices, n_channels = 2, 6, 3
+        covmats = get_covmats(n_matrices, n_channels)
+        labels = get_labels(n_matrices, n_classes)
         self.clf_predict(classif, covmats, labels)
         self.clf_fit_independence(classif, covmats, labels)
         if classif is MDM:
```
qbarthelemy, line 24: Currently, SVC is not really tested. It must be added in each if-line below:
```suggestion
        if classif is (MDM, SVC):
```

gabelstein:
I made it all a bit more flexible. Now online SVC/SVR is possible, when a custom kernel is passed.

gabelstein:
> Are you ok to define `probability=True` by default for SVC?

unfortunately, that drastically increases fit times. I will look at the tests and see how to fix them so they run properly.

agramfort:
keep it False by default like in sklearn

qbarthelemy (approved):
Thx @gabelstein for these great contributions!

agramfort:
👏 🙏

PR #167, opened 2022-04-09
Update example on SSVEP and add helpers

qbarthelemy:
This PR:

- updates example on SSVEP classification by MDM;
- ~uses the first four training sessions, in order to classify 128 trials instead of 32~;
- moves helper functions for SSVEP signal processing into a dedicated sub-module (in order to be re-used by future examples on SSVEP).


sylvchev (approved):
Thanks @qbarthelemy 
I'll check if the download location could be compatible with MOABB. It seems that it is not the case now.

PR #168, opened 2022-04-18
Update to Python-3.7

sylvchev:
Test and configuration still support python-3.6, while this version is deprecated. Time to move to 3.7 and try testing pyriemann against python-3.9.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,36 +5,45 @@ from setuptools import setup, find_packages
 
 # get the version (don't import mne here, so dependencies are not needed)
 version = None
-with open(op.join('pyriemann', '_version.py'), 'r') as fid:
+with open(op.join("pyriemann", "_version.py"), "r") as fid:
     for line in (line.strip() for line in fid):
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -5,36 +5,45 @@
...
+        "docs": [
+            "sphinx-gallery",
+            "sphinx-bootstrap_theme",
+            "numpydoc",
+            "mne",
+            "seaborn",
+        ],
+        "tests": ["pytest", "seaborn", "flake8"],
+    },
+    zip_safe=False,
```
agramfort, line 48: @sylvchev if we switch to black formatting we should enforce it in the CI.
Autoformatting creates big diffs that take time to reviewers for no good reason.

sylvchev: Yes, completely agree. For this specific file, it was set up before CI and did not pass the linting test.
I could revert and use autopep8 or equivalent if you prefer.

agramfort: could you make sure that now this file is part of the linting?

PR #169, opened 2022-04-20
Add example on PGA

qbarthelemy:
This PR adds an animated example in the gallery, 
on visualization of SSVEP classification using principal geodesic analysis (PGA).


sylvchev (approved):

sylvchev:
I propose to use a different color scheme for PGA visualization, we could revert to previous one if you don't like it.

qbarthelemy:
Thx for your help on data management @sylvchev !

PR #170, opened 2022-04-20
Add power mean for matrix

qbarthelemy:
This PR:

- implements the power mean for SPD matrices
(original code from https://github.com/plcrodrigues/means-field-classifier/blob/master/power_means.py)

- completes tests.

### DIFF
--- a/pyriemann/utils/mean.py
+++ b/pyriemann/utils/mean.py
@@ -2,7 +2,7 @@
 from copy import deepcopy
 import numpy as np
 
-from .base import sqrtm, invsqrtm, logm, expm
+from .base import sqrtm, invsqrtm, logm, expm, powm
 from .ajd import ajd_pham
 from .distance import distance_riemann
 from .geodesic import geodesic_riemann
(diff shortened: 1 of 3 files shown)


pyriemann/utils/mean.py
```diff
@@ -363,6 +363,66 @@ def mean_alm(covmats, tol=1e-14, maxiter=100,
...
+    .. [1] Lim Y and Palfia M. "Matrix Power means and the Karcher mean", J.
+           Funct. Anal., 2012
+    .. [2] Congedo M, Barachant A and Kharati K E. "Fixed Point Algorithms for
+           Estimating Power Means of Positive Definite Matrices", IEEE Trans.
+           Sig. Process., 2017
+    """
+    if not isinstance(p, (int, float)):
+        raise ValueError("Power mean only defined for a scalar exponent")
+    if p == 0:
+        raise ValueError("Exponent must be non-zero")
```
plcrodrigues, line 391: Maybe indicate that when p -> 0 we are actually looking for the geometric mean.

pyriemann/utils/mean.py
```diff
@@ -363,6 +363,66 @@ def mean_alm(covmats, tol=1e-14, maxiter=100,
         return C_iter.mean(axis=0)
 
 
+def mean_power(covmats, p, *, sample_weight=None, zeta=10e-10):
```
plcrodrigues, line 366: What is the * here? Shouldn't this be explained in the doc?

plcrodrigues (commented):
Cool, thanks for adding this feature @qbarthelemy 

I think it would be very useful to users if we proposed an easy way of estimating a list of different means of an array of SPD matrices. We would input a list of different values of p that we wanted to consider and at the output we would get the power mean corresponding to each one of them.

plcrodrigues:
Also, I would have added an example showing how the power means can be used to make a more robust classifier, like what we did in [here](https://hal.archives-ouvertes.fr/hal-02315131).

pyriemann/utils/mean.py
qbarthelemy: Parameters after * are keywords only.

qbarthelemy:
> Also, I would have added an example showing how the power means can be used to make a more robust classifier, like what we did in [here](https://hal.archives-ouvertes.fr/hal-02315131).

Yes, I do agree: I will add an example, but in the next PR adding `MeansField()` classifier.
You can squash+merge if current branch is ok for you.

plcrodrigues (approved):

sylvchev (approved):
Tests and code are nice, let get this merge to move to MeansFields PR and example!

Marco-Congedo:
Hi, i would allow passing p=0 as argument, calling in this case the function for computing the geometric mean.

PR #171, opened 2022-04-26
Add example to compare classifiers

qbarthelemy:
This PR adds an example to compare several Riemannian classifiers on low-dimensional synthetic datasets, adapted to SPD matrices from https://scikit-learn.org/stable/auto_examples/classification/plot_classifier_comparison.html

@gabelstein

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -46,6 +46,8 @@ v0.2.8.dev
 
 - Fix ``fit`` in :class:`pyriemann.channelselection.ElectrodeSelection`
 
+- Add example to compare classifiers on synthetic datasets
+
 
 v0.2.7 (June 2021)
 ------------------
--- a/examples/simulated/plot_classifier_comparison.py
+++ b/examples/simulated/plot_classifier_comparison.py
@@ -1,0 +1,192 @@
+"""
+===============================================================================
+Classifier comparison
+===============================================================================
+
+A comparison of several Riemannian classifiers on low-dimensional synthetic
+datasets, adapted to SPD matrices from [1]_.
+The point of this example is to illustrate the nature of decision boundaries
+of different classifiers.
+This should be taken with a grain of salt, as the intuition conveyed by
+these examples does not necessarily carry over to real datasets.
+
+The plots show training points in solid colors and testing points
+semi-transparent. The lower right shows the classification accuracy on the test
+set.
+
+"""
+# Authors: Quentin Barthélemy
+#
+# License: BSD (3-clause)
+
+from functools import partial
+
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.colors import ListedColormap
+from sklearn.model_selection import train_test_split
+
+from pyriemann.datasets import make_covariances, make_gaussian_blobs
+from pyriemann.classification import MDM, KNearestNeighbor
+
+
+###############################################################################
+
+
+@partial(np.vectorize, excluded=['cov_11', 'clf'])
+def get_proba(cov_00, cov_01, cov_11, clf):
+    cov = np.array([[cov_00, cov_01], [cov_01, cov_11]])
+    with np.testing.suppress_warnings() as sup:
+        sup.filter(RuntimeWarning)
+        return clf.predict_proba(cov[np.newaxis, ...])[0, 1]
+
+
+###############################################################################
+
+names = [
+    "MDM",
+    "Nearest Neighbors"
+]
+classifiers = [
+    MDM(),
+    KNearestNeighbor(n_neighbors=3)
+]
+
+rs = np.random.RandomState(2022)
+n_matrices, n_channels = 50, 2
+
+datasets = [
+    (
+        np.concatenate([
+            make_covariances(
+                n_matrices, n_channels, rs, evals_mean=10, evals_std=1
+            ),
+            make_covariances(
+                n_matrices, n_channels, rs, evals_mean=15, evals_std=1
+            )
+        ]),
+        np.concatenate([np.zeros(n_matrices), np.ones(n_matrices)])
+    ),
+    (
+        np.concatenate([
+            make_covariances(
+                n_matrices, n_channels, rs, evals_mean=10, evals_std=2
+            ),
+            make_covariances(
+                n_matrices, n_channels, rs, evals_mean=12, evals_std=2
+            )
+        ]),
+        np.concatenate([np.zeros(n_matrices), np.ones(n_matrices)])
+    ),
+    make_gaussian_blobs(
+        2*n_matrices, n_channels, random_state=rs, class_sep=3., class_disp=0.5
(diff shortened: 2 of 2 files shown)


qbarthelemy:
Display of the current example:
comp

agramfort:
decision functions are bit weird no? for example I am surprised that the kNN decision function is not more irregular.

gabelstein:
Thanks, looks good! will include it @qbarthelemy

gabelstein:
On closer inspection, this will probably be very confusing, as we can only display two dimensions of the three dimensional space. it could be alleviated with a 3d-plot, but that plot couldn't show decision boundaries properly. I'll look into another way of including an example.

qbarthelemy:
> decision functions are bit weird no? for example I am surprised that the kNN decision function is not more irregular.

Good catch! Digging into the code, I have discovered that current version of KNN does not implement `predict_proba()` : it only inherites from MDM, which has only one center by class.

> will include it

Ok, but do not merge this branch in yours. Wait the merge, and rebase your branch on the last master.

> On closer inspection, this will probably be very confusing, as we can only display two dimensions of the three dimensional space. it could be alleviated with a 3d-plot, but that plot couldn't show decision boundaries properly. I'll look into another way of including an example.

Previous example plots the decision boundary for the horizontal 2D plane going through the mean value of the third coordinates.
3D decision boundaries are not easy to show, but I try a new display.

qbarthelemy:
I use `RandomState`, but figures are not reproducible from one run to another.
comp_3d

pyriemann/classification.py
```diff
@@ -511,6 +516,36 @@ def predict(self, covtest):
...
+        Returns
+        -------
+        prob : ndarray, shape (n_matrices, n_classes)
+            Probabilities for each class.
+        """
+        n_matrices, _, _ = X.shape
+
+        dist = self._predict_distances(X)
+        idx = np.argsort(dist)
+        dist_sorted = np.take_along_axis(dist, idx, axis=1)
```
agramfort, line 540:
```suggestion
        dist_sorted = dist[:, idx[:, :self.n_neighbors]]
```
Should be enough I think

pyriemann/classification.py
```diff
@@ -511,6 +516,36 @@ def predict(self, covtest):
...
+        neighbors_classes = self.classmeans_[idx]
+        probas = softmax(-dist_sorted[:, 0:self.n_neighbors]**2)
+
+        prob = np.zeros((n_matrices, len(self.classes_)))
+        for m in range(n_matrices):
+            for il, ll in enumerate(self.classes_):
+                prob[m, il] = np.sum(
+                    probas[m, neighbors_classes[m, 0:self.n_neighbors] == ll]
+                )
```
agramfort, line 549: Can you see how it differs from how sklearn does it?

https://github.com/scikit-learn/scikit-learn/blob/baf828ca126bcb2c0ad813226963621cafe38adb/sklearn/neighbors/_classification.py#L240

doc/whatsnew.rst
```diff
@@ -46,6 +46,8 @@ v0.2.8.dev
 
 - Fix ``fit`` in :class:`pyriemann.channelselection.ElectrodeSelection`
 
+- Add example to compare classifiers on synthetic datasets
```
agramfort, line 49: You will need a what’s new entry for the bug fix in kNN too I think. Unless it was so far never released

pyriemann/classification.py
qbarthelemy: Yes, I had checked: methods are very similar.

The difference is the definition of probabilities from distances:
pyRiemann uses a softmax of squared distances (this formula is derived from the Riemannian Gaussian distribution, as explained in #100),
whereas sklearn uses reciprocal of distances, and then divides by the sum (to have the sum of proba equal to one) (but, I don't know the origin of this formula).

pyriemann/classification.py
qbarthelemy: I tried, but it does not work.

doc/whatsnew.rst
qbarthelemy: Done.

examples/simulated/plot_classifier_comparison.py
```diff
@@ -0,0 +1,216 @@
...
+
+###############################################################################
+
+figure = plt.figure(figsize=(8, 10))
+i = 1
+# iterate over datasets
+for ds_cnt, (X, y) in enumerate(datasets):
+    # split dataset into training and test part
+    X_train, X_test, y_train, y_test = train_test_split(
+        X, y, test_size=0.4, random_state=42
```
sylvchev, line 104:
```suggestion
        X, y, test_size=0.4, random_state=rs
```
Does it make the figure reproducible?

sylvchev:
Good catch for the missing RandomState in `make_gaussian_blobs`!
The doc is nicely generated, but is it possible to use a different to have a more interpretable example on the 3rd line? See below on the artifact generated by GH Action, the 3rd dataset seems to have only blue class or indecisive probability for the `contourf` plot. On the 3D example you posted above, the partition of the space seems more conclusive.

sphx_glr_plot_classifier_comparison_001

pyriemann/classification.py
agramfort: before committing and making a choice that differs from sklearn I would run a tiny benchmark on Gaussian data following LDA model to see what approach leads to the best calibrated probability.

qbarthelemy: > I would run a tiny benchmark

Nice idea! But, who is "I"? you or me? ;-)

qbarthelemy: Joking aside, I don't see how to do this benchmark.
Because inputs of sklearn are multivariate vectors (generated by a mixture of multivariate Gaussian distributions, ok), while inputs of pyRiemann are covariance matrices. So, results can't be compared.

agramfort:
I would do it in sklearn using euclidian data. Basically replace the
predict_proba in sklearn and see what works best
in sklearn. Then copy this in PyRiemann

pyriemann/classification.py
qbarthelemy: > I would do it in sklearn using euclidian data. Basically replace the
predict_proba in sklearn and see what works best
in sklearn. Then copy this in PyRiemann

I plot log-probabilities of kNN applied on bivariate Gaussian distributions:
- weights="distance" computing reciprocal of distances (classical sklearn implem);
- weights="max_lik", computing softmax of squared distances, a formula derived by maximum likelihood from a (Euclidean) Gaussian distribution (implem added in this branch).

Results are really close, but I think that new option is better, because derived from MLE.
Moreover, this option it could be added in sklearn.

test_knn_comparison

qbarthelemy: @agramfort , can we merge the branch?

sylvchev:
LGTM
Kudos @qbarthelemy

PR #155, opened 2021-12-10
Adding MDWM for EEG transfer learning

emmanuelkalunga:
WIP on the implementation of MDWM classifier for transfer learning on EEG.

- [x] MDWM class
- [x] Tests
- [x] Example
- [x] Rebase the branch
- [ ] Update "what's new" page, and API (adding module transfer learning)

PR #177, opened 2022-06-18
(WIP) MDWM for transfer learning

sylvchev:
This is a new home for PR #155 to correct my messed up rebase.


emmanuelkalunga:
Much cleaner. Thanks @sylvchev

sylvchev:
@quentin I take into account your code suggestions from previous PR. Added Transfer learning section in  `api.rst` and review the generated doc from GH Action (everything is ok)

@agramfort I don't have more work to submit on this PR.

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __init__(self, transfer_coef, metric='riemann', n_jobs=1):
+        """Init."""
+        self.transfer_coef = transfer_coef
+        self.metric = metric
+        self.n_jobs = n_jobs
+        self.covmeans_ = None
+        self.classes_ = None
+        self._target_means = None
+        self._domain_means = None
```
agramfort, line 74: these attributes should only appear after the fit. This init should just set params as attributes

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def fit(self, X, y, X_source, y_source, sample_weight=None):
```
agramfort, line 76: hum we should discuss this as this violates the API of other models.

cc @plcrodrigues as we discussed what's best to do this

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,175 @@
...
+            raise ValueError("X and y must be for the same number of \
+                matrices i.e. n_matrices")
+
+        if sample_weight is None:
+            sample_weight = np.ones(X_source.shape[0])
+
+        if not (X_source.shape[0] == y_source.shape[0]
+                == sample_weight.shape[0]):
+            raise ValueError("X and y must be for the same number of \
+                matrices i.e. n_matrices")
```
agramfort, line 145: sample_weight can cause a failure and the text does not reflect this

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,175 @@
...
+        else:
+            self.target_means_ = Parallel(n_jobs=self.n_jobs)(
+                delayed(mean_covariance)(X[y == ll], metric=self.metric_mean)
+                for ll in self.classes_)
+            self.source_means_ = Parallel(n_jobs=self.n_jobs)(
+                delayed(mean_covariance)(
+                    X_source[y_source == ll],
+                    metric=self.metric_mean,
+                    sample_weight=sample_weight[y_source == ll])
+                for ll in self.classes_)
```
agramfort, line 169: this branching is useless. When n_jobs =1 then joblib Parallel is a pure list comprehension

tests/test_transferlearning.py
```diff
@@ -0,0 +1,246 @@
...
+from numpy.testing import assert_array_equal
+from pyriemann.transferlearning import MDWM
+import pytest
+
+
+mat_transfer = [MDWM]
+
+
+@pytest.mark.parametrize("transfer", mat_transfer)
+class TransferTestCase:
```
agramfort, line 11: can you avoid using classes for tests and use plain functions and fixtures to share code between tests?

tests/test_transferlearning.py
```diff
@@ -0,0 +1,246 @@
...
+
+        covmats_source = get_covmats(n_matrices_source, n_channels)
+        labels_source = get_labels(n_matrices_source, n_classes_source)
+
+        covmats_target = get_covmats(n_matrices_target, n_channels)
+        labels_target = get_labels(n_matrices_target, n_classes)
+
+        trans = transfer(transfer_coef=0.5)
+        trans.fit(covmats_target, labels_target, covmats_source,
+                  labels_source, sample_weight)
```
agramfort, line 141: there is a lot of code duplication between the tests

examples/transfer-learning/plot_classify_mdwm.py
```diff
@@ -0,0 +1,288 @@
+"""
+=====================================================================
+ Apply MDWM transfer learning to classify a set of matrices
+=====================================================================
+
+Applying MDWM to learn from source matrices and classify target matrices.
```
agramfort, line 6: the example does not explain that MDWM means Minimum Distance to Weighted Mean. The header of the example should explain a few things.

sylvchev:
Thanks for the review @agramfort I could take into account most of your remarks quickly. We definitely need to talk about the API for transfer learning.

pyriemann/transferlearning.py
plcrodrigues: Indeed. It might be interesting to follow the approach from `mvlearn` (https://mvlearn.github.io/) where they pass a list of datasets to be considered. In our case we could say that by convention the last dataset in the list should be considered the target. This would also allow to easily consider the multi-source case.

plcrodrigues:
Thanks for the PR, @sylvchev. 

Small question for which I don't know exactly what to think about. I am not sure what we should be calling "transfer learning" in our package. In methods like RPA [1](https://hal.archives-ouvertes.fr/hal-01971856/document), Parallel transport [2](https://ieeexplore.ieee.org/document/8624413), and Zanini et al. [3](https://ieeexplore.ieee.org/document/8013808) we first do a **dataset matching** step and then choose a classifier if we want to do classification; but we could choose any other statistical task afterwards. In MDWM the dataset matching and classification seem to be done at the same time, right? Do you think we can split the procedure into two steps? 

[1] Rodrigues et al. "Riemannian Procrustes Analysis : Transfer Learning for Brain-Computer Interfaces"
[2] Yair et al. "Parallel Transport on the Cone Manifold of SPD Matrices for Domain Adaptation"
[3] Zanini et al. "Transfer Learning: A Riemannian Geometry Framework With Applications to Brain–Computer Interfaces"

pyriemann/transferlearning.py
plcrodrigues: One important difference with `mvlearn` is that we want to handle datasets that can have different numbers of epochs.

pyriemann/transferlearning.py
sylvchev: Ok, same situation happend in Kmeans and MDM, I'll correct this one here and I'll remove the branching from these models in another PR.

tests/test_transferlearning.py
sylvchev: I agree that in this case it could be plain function as there is only one model to test. The goal is to have a TestCase class to make the tests for the common sklearn API functions (fit, predict, predict_proba, ...), as we have done in [test_classification](https://github.com/pyRiemann/pyRiemann/blob/master/tests/test_classification.py#L26) and other test file. The tests for specific aspects of the model are in plain function and use fixtures.

I'll update the code to make it clearer.

tests/test_transferlearning.py
sylvchev: yes, with source/target data, transfer learning makes lengthy tests, but each test use fixture and specific variable names to make the object of the test more explicit. With @emmanuelkalunga we tried to factorize the tests as much as possible.

sylvchev:
@plcrodrigues I see your point, separate the dataset matching from the classification. This is an interesting idea, along the line of euclidean alignment or other transfer learning techniques. I'll try to see if we could distangle those parts for MDWM. 

Say we have a dataset matching/alignment class, with maybe the API of a scikit transformer, we could plug any classifer afterwards. I completely agree that we should find the correct API to do that and it should definitely be able to handle case that goes beyond the 1-source subject/1-target subject case.

I see two directions, one follows a [mvlearn-like approach](https://mvlearn.github.io/index.html) to use list of ndarrays as input `Xs` (e.g. length of the list is the number of subjects or the number of datasets and each item is a set of matrices as we use in pyriemann). The other possibility is to follow a [MOABB-like approach](https://neurotechx.github.io/moabb/auto_examples/plot_explore_paradigm.html), were we provide a `metadata` DataFrame object that specifies the information associated with each matrix (subject, session & run in MOABB and we could definitely add dataset and paradigm).

mvlearn pros and cons:
[+] minimal modification on scikit API (but still break it)
[+] implicit definition of input, it is just a list of ndarrays
[-] no definition of input, is it subject-level, is it dataset level?
[-] plain data structure, ok for multiview but limited for transfer (if list items are datasets, no information regarding subject within each dataset)

moabb pros and cons:
[+] hierarchical data structure (one matrix could have multiple tag -> subject, session, dataset)
[+] explicit information regarding data
[-] metadata structure should not be modified and its definition is critical
[-] scikit API is modified explicitely

qbarthelemy (changes requested):
Thanks for this new PR, much cleaner!

agramfort:
sleeping over it I would say let's wait to merge this one. We need to have a conversation all together with @plcrodrigues about API

any objection?

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,158 @@
...
+        ----------
+        X : ndarray, shape (n_matrices, n_channels, n_channels)
+            Set of SPD matrices from target domain.
+        y : ndarray, shape (n_matrices,)
+            Labels for each matrix of target domain.
+        X_source : ndarray, shape (n_matrices_source, n_channels, n_channels)
+            Set of SPD matrices from source domain.
+        y_source : ndarray, shape (n_matrices_source,)
+            Labels for each matrix of source domain.
+        sample_weight : None | ndarray, shape (n_matrices_source,), default=None
```
qbarthelemy, line 85: This modification will remove flake8 error.
```suggestion
        sample_weight : None | ndarray, shape (n_matrices_source,), \
            default=None
```

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,158 @@
...
+        elif isinstance(self.metric, dict):
+            # check keys
+            for key in ['mean', 'distance']:
+                if key not in self.metric.keys():
+                    raise KeyError('metric must contain "mean" and "distance"')
+
+            self.metric_mean = self.metric['mean']
+            self.metric_dist = self.metric['distance']
+        else:
+            raise TypeError('metric must be dict or str')
```
qbarthelemy, line 112: You should rebase your branch on the last master and use function `_check_metric` added in #181.
```suggestion
        self.metric_mean, self.metric_dist = _check_metric(self.metric)
```

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,147 @@
+import numpy as np
+from joblib import Parallel, delayed
+
+from .classification import MDM
```
qbarthelemy, line 4: To correct tests:
```suggestion
from .classification import MDM, _check_metric
```

qbarthelemy:
Closed by #189

PR #179, opened 2022-06-21
speedup for make_gaussian_blobs

sylvchev:
`make_gaussian_blobs` is a really nice addition to pyRiemann, but it is computationally expensive. The complexity depends mainly on the number of matrices to generate. The sampling code is a loop that generate one matrix at each iteration. As the samples are indenpendent, we could easily parallelize the process to reduce computation time. This is the object of this PR.

NB: this could reduce a lot the doc building time:
```
computation time summary:
    - ../examples/simulated/plot_classifier_comparison.py:                235.60 sec   0.0 MB
    - ../examples/stats/plot_oneWay_Manova.py:                             67.49 sec   0.0 MB
    - ../examples/artifacts/plot_detect_riemannian_potato_field_EEG.py:    49.95 sec   0.0 MB
    - ../examples/artifacts/plot_detect_riemannian_potato_EEG.py:          37.31 sec   0.0 MB
    - ../examples/simulated/plot_toy_classification.py:                    15.94 sec   0.0 MB
    - ../examples/ERP/plot_classify_MEG_mdm.py:                            11.12 sec   0.0 MB
    - ../examples/ERP/plot_embedding_MEG.py:                               11.03 sec   0.0 MB
...
```

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -64,6 +64,10 @@ v0.2.8.dev
 
 - Add :func:`pyriemann.utils.distance.distance_harmonic`, and sort functions by their names in code, doc and tests
 
+- Parallelize functions for dataset generation: :func:`pyriemann.datasets.make_gaussian_blobs`
+
+- Fix dispersion when generating datasets: :func:`pyriemann.datasets.sample_gaussian_spd`
+
 v0.2.7 (June 2021)
 ------------------
 
--- a/examples/simulated/plot_classifier_comparison.py
+++ b/examples/simulated/plot_classifier_comparison.py
@@ -224,11 +224,11 @@ datasets = [
     ),
     make_gaussian_blobs(
         2*n_matrices, n_channels, random_state=rs, class_sep=1., class_disp=.2,
-        n_jobs=-1
+        n_jobs=4
     ),
     make_gaussian_blobs(
         2*n_matrices, n_channels, random_state=rs, class_sep=.5, class_disp=.5,
-        mat_std=.1, n_jobs=-1
+        mat_std=.1, n_jobs=4
     )
 ]
 n_datasets = len(datasets)
--- a/examples/simulated/plot_classifier_comparison.py
+++ b/examples/simulated/plot_classifier_comparison.py
@@ -223,11 +223,12 @@ datasets = [
         y
     ),
     make_gaussian_blobs(
-        2*n_matrices, n_channels, random_state=rs, class_sep=1., class_disp=.2
+        2*n_matrices, n_channels, random_state=rs, class_sep=1., class_disp=.2,
+        n_jobs=-1
     ),
     make_gaussian_blobs(
         2*n_matrices, n_channels, random_state=rs, class_sep=.5, class_disp=.5,
-        mat_std=.1
+        mat_std=.1, n_jobs=-1
     )
 ]
 n_datasets = len(datasets)
--- a/pyriemann/datasets/sampling.py
+++ b/pyriemann/datasets/sampling.py
@@ -33,7 +33,7 @@ def _pdf_r(r, sigma):
         raise ValueError(f'sigma must be a positive number (Got {sigma})')
 
     n_dim = len(r)
-    partial_1 = -np.sum(r**2) / sigma**2
+    partial_1 = -np.sum(r**2) / (2 * sigma**2)
     partial_2 = 0
     for i in range(n_dim):
         for j in range(i + 1, n_dim):
--- a/pyriemann/datasets/sampling.py
+++ b/pyriemann/datasets/sampling.py
@@ -130,7 +130,7 @@ def _slice_sampling(ptarget, n_samples, x0, n_burnin=20, thin=10,
         expect the whole sampling procedure to take longer.
     random_state : int, RandomState instance or None, default=None
         Pass an int for reproducible output across multiple function calls.
-    n_jobs : int, (default: 1)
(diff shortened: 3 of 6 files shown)


agramfort:
@sylvchev can you say how much n_jobs helps by giving running time you obtain? 🙏

examples/simulated/plot_classifier_comparison.py
```diff
@@ -85,11 +85,12 @@ def get_proba(cov_00, cov_01, cov_11, clf):
         y
     ),
     make_gaussian_blobs(
-        2*n_matrices, n_channels, random_state=rs, class_sep=1., class_disp=.2
+        2*n_matrices, n_channels, random_state=rs, class_sep=1., class_disp=.2,
+        n_jobs=-1
```
agramfort, line 89: don't do this. If someone runs this code a big shared computer without paying attention it will be mess

sylvchev:
> @sylvchev can you say how much n_jobs helps by giving running time you obtain? 🙏

for plot_toy_classification.py, with n_jobs=1:
```13 s ± 438 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)```
and with n_jobs=8:
```5.88 s ± 940 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)```

File: pyriemann/datasets/sampling.py [hunk 2]
```diff
@@ -299,11 +339,13 @@ def sample_gaussian_spd(n_matrices, mean, sigma, random_state=None):
     """
 
     n_dim = mean.shape[0]
+    # dispersion is corrected w.r.t. dimension
     samples_centered = _sample_gaussian_spd_centered(
         n_matrices=n_matrices,
         n_dim=n_dim,
-        sigma=sigma,
-        random_state=random_state
+        sigma=sigma / np.sqrt(n_dim),
```
agramfort, line 346: The bug fixes for released code should be documented in a entry in what’s new

qbarthelemy (approved):

pyriemann/datasets/sampling.py
```diff
@@ -41,8 +42,66 @@ def _pdf_r(r, sigma):
...
+    ptarget : function with one input
+        The target pdf to sample from or a multiple of it.
+    x0 : array
+        Initial state for the MCMC procedure. Note that the shape of this array
+        defines the dimensionality n_dim of the data points to be sampled.
+    w : float
+        Initial bracket width.
+    rs : int, RandomState instance or None
+        Pass an int for reproducible output across multiple function calls.
+
```
agramfort, line 59: Returns?

pyriemann/datasets/sampling.py
```diff
@@ -89,58 +151,19 @@ def _slice_sampling(ptarget, n_samples, x0, n_burnin=20, thin=10,
...
-                    else:
-                        xl_i = xprime_i
-
-            # store coordinate i of new sample
-            xt = np.copy(xt)
-            xt[i] = xprime_i
-
-        samples.append(xt)
+    samples = Parallel(n_jobs=n_jobs)(
+        delayed(_slice_one_sample)(ptarget, x0, w, rs)
```
agramfort, line 158: this can be dangerous to pass a RandomState like this. You will see that when using different processes the RandomState will be copied so each parallel run will effectively be seeded the same way. I suspect this code lacks diversity of produced samples when run in parallel.

actually read this https://joblib.readthedocs.io/en/latest/auto_examples/parallel_random_state.html maybe with the current loly backend is fine.

can you check that sequential or parallel runs give the same outputs when using a seed?

doc/whatsnew.rst
```diff
@@ -62,6 +62,10 @@ v0.2.8.dev
 
 - Add example on principal geodesic analysis (PGA) for SSVEP classification
 
+- Parallelize functions for dataset generation: :func:`pyriemann.datasets.make_gaussian_blobs`
+
+- Fix dispersion when generating datasets: :func:`pyriemann.datasets.sample_gaussian_spd`
```
agramfort, line 67: we should split the what's new between Enhancements / New Features and Bug fixes. And we should refer to the PRs on github. See what we use in mne https://github.com/mne-tools/mne-python/blob/main/doc/sphinxext/gh_substitutions.py

pyriemann/datasets/sampling.py
qbarthelemy, line 59:
```suggestion

    Returns
    -------
    sample : ndarray, shape (n_dim,)
        Sample from the target pdf.
```

pyriemann/datasets/sampling.py
```diff
@@ -41,8 +42,66 @@ def _pdf_r(r, sigma):
...
+        # store coordinate i of new sample
+        xt = np.copy(xt)
+        xt[i] = xprime_i
+
+    return xt
+
+
 def _slice_sampling(ptarget, n_samples, x0, n_burnin=20, thin=10,
-                    random_state=None):
+                     random_state=None, n_jobs=1):
```
qbarthelemy, line 104: To suppress flake8 error:
```suggestion
                    random_state=None, n_jobs=1):
```

doc/whatsnew.rst
sylvchev: Yes, I'll push this in a new PR

qbarthelemy (approved):
Good job!

plcrodrigues (approved):
Thanks Sylvain !

agramfort:
@plcrodrigues merge if happy :)

PR #180, opened 2022-06-21
Correct tests for classifiers using tsupdate

qbarthelemy:
Currently, `clf_tsupdate` is not executed due to the type `is` in `if classif is (FgMDM, TSclassifier)`.

Instead of the correction `if classif in (FgMDM, TSclassifier)`, I use `if hasattr(classif(), 'tsupdate')` which does not require the exhaustive sublist of classifiers to test, and which is more flexible when we will add new classifiers.


qbarthelemy:
Tests are ok, I merge.

Issue #52, opened 2017-03-26
[ENH] Add example with benchark of different mean/distance metric for MDM


alexandrebarachant:
MDM works by default with Riemannian mean and distance, but different combination of metrics can be set

```python
metric = dict(distance='euclid', mean='wasserstein')
mdm = MDM(metric=metric)
```

The choice of metric impact peformance and computational cost. An example with benchmark of different metric for the mean and distance needed.

qbarthelemy:
Accuracies and computational times are compared for the three main metrics (`riemann`, `logeuclid` and `euclid`), since #182.

PR #182, opened 2022-06-25
Update classifier comparison, adding different metrics and computational times

qbarthelemy:
This PR updates the example on classifier comparison:
- classifiers are compared for `riemann`, `logeuclid` and `euclid` metrics;
- training and test times are printed.

Related to #52.

### DIFF
--- a/examples/simulated/plot_classifier_comparison.py
+++ b/examples/simulated/plot_classifier_comparison.py
@@ -260,8 +260,7 @@ plot_classifiers("euclid")
 ###############################################################################
 # References
 # ----------
-# .. [1] https://scikit-learn.org/stable/auto_examples/classification/\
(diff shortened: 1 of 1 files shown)


examples/simulated/plot_classifier_comparison.py
```diff
@@ -99,123 +237,31 @@ def get_proba(cov_00, cov_01, cov_11, clf):
...
+# ---------------------------------
+
+plot_classifiers("euclid")
 
 
 ###############################################################################
 # References
 # ----------
 # .. [1] https://scikit-learn.org/stable/auto_examples/classification/\
 #    plot_classifier_comparison.html
```
agramfort, line 264: check the rendered doc. This link does not render well

PR #183, opened 2022-06-25
Sort metric based functions by alphabetic order

qbarthelemy:
Sort functions related to metrics and covariance estimator in alphabetic order, in code, tests and doc.

Rewrite `distance_logdet`, add `distance_harmonic`, complete tests.

### DIFF
--- a/pyriemann/utils/ajd.py
+++ b/pyriemann/utils/ajd.py
@@ -20,6 +20,91 @@ def _get_normalized_weight(sample_weight, data):
     return normalized_weight
 
 
+def rjd(X, eps=1e-8, n_iter_max=1000):
+    """Approximate joint diagonalization based on Jacobi angles.
+
+    This is a direct implementation of the AJD algorithm by Cardoso and
+    Souloumiac [1]_ used in JADE. The code is a translation of the Matlab code
+    provided in the author website.
+
+    Parameters
+    ----------
+    X : ndarray, shape (n_matrices, n_channels, n_channels)
+        Set of symmetric matrices to diagonalize.
+    eps : float, default=1e-8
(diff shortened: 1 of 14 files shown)


pyriemann/utils/ajd.py
```diff
@@ -198,6 +114,91 @@ def ajd_pham(X, eps=1e-6, n_iter_max=15, sample_weight=None):
...
+                    A[q, :] = c * A[q, :] - s * tmp
+
+                    tmp = V[:, p].copy()
+                    V[:, p] = c * V[:, p] + s * V[:, q]
+                    V[:, q] = c * V[:, q] - s * tmp
+
+    D = np.reshape(A, (m, int(nm / m), m)).transpose(1, 0, 2)
+    return V, D
+
+
```
agramfort, line 201: I would not have moved this. This breaks git blame for little gain

pyriemann/utils/distance.py
```diff
@@ -58,9 +82,11 @@ def distance_euclid(A, B):
     Returns
     -------
     d : float
-        Euclidean distance between A and B.
+        Log-det distance between A and B.
     """
-    return np.linalg.norm(A - B, ord='fro')
+    return np.sqrt(np.log(np.linalg.det(
+        (A + B) / 2.0)) - 0.5 *
+        np.log(np.linalg.det(A)*np.linalg.det(B)))
```
agramfort, line 89: This should use slogdet function 

https://numpy.org/doc/stable/reference/generated/numpy.linalg.slogdet.html

which is more stable

agramfort (approved):

sylvchev (approved):

PR #186, opened 2022-07-08
Base functions able to process ndarray of SPD matrices

qbarthelemy:
This PR improves base functions to process ndarray of SPD matrices, and completes tests.

This modification will allow to remove many for-loops from code,
providing simpler and faster (using numpy broadcasting) functions (WIP).

### DIFF
--- a/pyriemann/utils/base.py
+++ b/pyriemann/utils/base.py
@@ -1,36 +1,46 @@
 """Base functions for SPD matrices."""
 
 import numpy as np
-from scipy.linalg import eigh
-
 from numpy.core.numerictypes import typecodes
+
+
+def _diag(X):
+    """Construct ndarray from diagonal terms.
+
+    Parameters
+    ----------
+    X : ndarray, shape (..., n)
+        Diagonal terms of ndarray.
+
+    Returns
+    -------
+    Y : ndarray, shape (..., n, n)
+        Diagonal ndarray.
+    """
+    dims = X.shape
+    n = dims[-1]
+    Y = np.zeros((*dims, n), X.dtype)
+    Y[..., np.arange(n), np.arange(n)] = X
+    return Y
 
 
 def _matrix_operator(C, operator):
     """Matrix function."""
+    if not isinstance(C, np.ndarray) or C.ndim < 2:
+        raise ValueError('Input must be at least a 2D ndarray')
     if C.dtype.char in typecodes['AllFloat'] and not np.isfinite(C).all():
         raise ValueError(
             "Matrices must be positive definite. Add "
             "regularization to avoid this error.")
-    eigvals, eigvects = eigh(C, check_finite=False)
-    D = (eigvects * operator(eigvals)[np.newaxis, :]) @ eigvects.T
+    eigvals, eigvecs = np.linalg.eigh(C)
+    D = eigvecs @ _diag(operator(eigvals)) @ np.swapaxes(eigvecs, -2, -1)
     return D
-
-
-def _matrices_operator(C, operator):
-    """Recursive matrix function."""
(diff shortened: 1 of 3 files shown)


pyriemann/utils/base.py
```diff
@@ -1,25 +1,37 @@
...
-    """Matrix equivalent of an operator."""
+    """Matrix function."""
     if C.dtype.char in typecodes['AllFloat'] and not np.isfinite(C).all():
         raise ValueError(
-            "Covariance matrices must be positive definite. Add "
+            "Matrices must be positive definite. Add "
             "regularization to avoid this error.")
     eigvals, eigvects = eigh(C, check_finite=False)
     eigvals = np.diag(operator(eigvals))
     D = eigvects @ eigvals @ eigvects.T
```
agramfort, line 17:
```suggestion
    D = (eigvects * operator(eigvals)[np.newaxis, :]) @ eigvects.T
```

pyriemann/utils/base.py
```diff
@@ -1,25 +1,36 @@
...
 
 def _matrix_operator(C, operator):
-    """Matrix equivalent of an operator."""
+    """Matrix function."""
     if C.dtype.char in typecodes['AllFloat'] and not np.isfinite(C).all():
         raise ValueError(
-            "Covariance matrices must be positive definite. Add "
+            "Matrices must be positive definite. Add "
             "regularization to avoid this error.")
     eigvals, eigvects = eigh(C, check_finite=False)
```
agramfort, line 15: you can actually call eigh on a 3D array to avoid the for loops and going through lists and arrays. See:

```
In [12]: size = 30

In [13]: C = np.random.random((3, size, size))

In [14]: evals, evecs = np.linalg.eig(C)

In [15]: evals.shape
Out[15]: (3, 30)

In [16]: evecs.shape
Out[16]: (3, 30, 30)

In [17]: np.testing.assert_allclose(evecs[0], np.linalg.eig(C[0])[1])

```

qbarthelemy: You are right.
[numpy-based eigh](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html) allows to process `(…, M, M) array`, but without `check_finite` option;
[scipy-based eigh](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.eigh.html) treats only `(M, M) array`, but with `check_finite` option.
I don't know what is the best option.

agramfort:
can you time? and since we use here check_finite=False
it suggests numpy as what we want?

qbarthelemy:
Comparison on 10 sets of 100 matrices, repeated 50 times.
pr_186
Thx for this good idea!

pyriemann/utils/base.py
```diff
@@ -1,25 +1,30 @@
...
+            "Matrices must be positive definite. Add "
             "regularization to avoid this error.")
-    eigvals, eigvects = eigh(C, check_finite=False)
-    eigvals = np.diag(operator(eigvals))
-    D = eigvects @ eigvals @ eigvects.T
+    eigvals, eigvecs = np.linalg.eigh(C)
+    eigvals = operator(eigvals)
+    if C.ndim >= 3:
+        eigvals = np.expand_dims(eigvals, -2)
+    D = (eigvecs * eigvals) @ np.swapaxes(eigvecs, -2, -1)
```
agramfort, line 20: @qbarthelemy I use broadcasting now

agramfort (approved):

agramfort:
thx @qbarthelemy !

pyriemann/utils/base.py
qbarthelemy: Great!

PR #187, opened 2022-07-09
Distance functions able to process ndarray of SPD matrices

qbarthelemy:
Using #186, this PR improves distances functions to process ndarray of SPD matrices, and completes tests.

Concerning eigvalsh, [numpy-based eigvalsh](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigvalsh.html) treats `(…, M, M) array` but solves the standard eigenvalue problem, not the generalized one like [scipy-based eigvalsh](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.eigvalsh.html).

Concerning solve, [numpy-based solve](https://numpy.org/doc/stable/reference/generated/numpy.linalg.solve.html) treats `(…, M, M) array` , but without option `assume_a` like [scipy-based solve](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html).

### DIFF
--- a/pyriemann/utils/distance.py
+++ b/pyriemann/utils/distance.py
@@ -6,10 +6,22 @@ from scipy.linalg import eigvalsh, solve
 from .base import logm, sqrtm
 
 
-def distance_euclid(A, B):
-    r"""Euclidean distance between two SPD matrices.
+def _recursive(fun, A, B, *args, **kwargs):
+    """Recursive fuction with two inputs."""
+    if not isinstance(A, np.ndarray) or A.ndim < 2:
+        raise ValueError('Input must be at least a 2D ndarray')
+    elif A.ndim == 2:
+        return fun(A, B, *args, **kwargs)
+    else:
+        return np.asarray(
+            [_recursive(fun, a, b, *args, **kwargs) for a, b in zip(A, B)]
+        )
 
-    Compute the Euclidean distance between two SPD matrices A and B, defined
+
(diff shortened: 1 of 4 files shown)


pyriemann/utils/distance.py
```diff
@@ -67,19 +78,20 @@ def distance_kullback(A, B):
...
 
     Returns
     -------
-    d : float
+    d : ndarray, shape (...,) or float
         Left Kullback-Leibler divergence between A and B.
     """
     n = A.shape[-1]
+    tr = np.trace(_recursive(solve, B, A, assume_a='pos'), axis1=-2, axis2=-1)
     logdet = np.log(np.linalg.det(B) / np.linalg.det(A))
```
agramfort, line 93:
```suggestion
    logdet = np.linalg.slogdet(B)[1] - np.linalg.slogdet(A)[1]
```

pyriemann/utils/distance.py
```diff
@@ -6,59 +6,70 @@
 from .base import logm, sqrtm
 
 
+def _recursive(fun, A, B, *args, **kwargs):
+    """Recursive fuction with two inputs."""
```
agramfort, line 10:
```suggestion
    """Recursive function with two inputs."""
```
I would also check that `A.shape == B.shape` or maybe just `len(A) == len(B)`

agramfort:
thx @qbarthelemy

PR #188, opened 2022-07-11
Speedup for means

qbarthelemy:
Using einsum and new base functions #186, this PR accelerates SPD matrix means.
Also, it uses weights when initializing LogDet, Riemannian, Wasserstein means on Euclidean mean.

Comparison on Riemannian (top fig) and LogEuclidean (bottom fig) means before and after PR, repeated 50 times.

pr_XXX_comp_means_riemann

pr_XXX_comp_means_logeuclid

Hoping that this PR fills a little bit of the gap with geomstats f*** fast implementation. @sylvchev @nguigs
https://github.com/nguigs/geomstats/blob/nguigs-pyriemann/notebooks/SPD%20-%20PyRiemann.ipynb

### DIFF
--- a/pyriemann/utils/mean.py
+++ b/pyriemann/utils/mean.py
@@ -1,5 +1,6 @@
 """Means of SPD matrices."""
 
+import warnings
 from copy import deepcopy
 import numpy as np
 
@@ -60,7 +61,7 @@ def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
     B, _ = ajd_pham(covmats)
 
     crit = np.inf
-    for k in range(maxiter):
+    for _ in range(maxiter):
         J = np.einsum('a,abc->bc', sample_weight, logm(B.T @ covmats @ B))
         update = np.diag(np.diag(expm(J)))
         B = B @ invsqrtm(update)
@@ -68,6 +69,8 @@ def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
         crit = distance_riemann(np.eye(n_channels), update)
         if crit <= tol:
             break
+    else:
+        warnings.warn('Convergence not reached')
 
     A = np.linalg.inv(B)
     J = np.einsum('a,abc->bc', sample_weight, logm(B.T @ covmats @ B))
@@ -75,8 +78,7 @@ def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
     return C
 
 
-def mean_alm(covmats, tol=1e-14, maxiter=100,
-             verbose=False, sample_weight=None):
+def mean_alm(covmats, tol=1e-14, maxiter=100, sample_weight=None):
     r"""Ando-Li-Mathias (ALM) mean of SPD matrices.
 
     Return the geometric mean recursively [1]_, generalizing from:
@@ -99,8 +101,6 @@ def mean_alm(covmats, tol=1e-14, maxiter=100,
         The tolerance to stop the gradient descent.
     maxiter : int, default=100
         The maximum number of iterations.
-    verbose : bool, default=False
-        Indicate when reaching maxiter.
     sample_weight : None | ndarray, shape (n_matrices,), default=None
         Weights for each matrix. If None, it uses equal weights.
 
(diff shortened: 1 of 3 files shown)


agramfort (commented):
Numpy 🔥💪

i just personally don’t like while loops and prefer for loops with max iter and use break when tol is reached. It’s more readable and less error prone I feel

pyriemann/utils/mean.py
```diff
@@ -368,15 +370,18 @@ def mean_power(covmats, p, *, sample_weight=None, zeta=10e-10):
...
             'a,abc->bc',
             sample_weight,
             powm(X @ powm(covmats, np.sign(p)) @ X.T, np.abs(p))
         )
         X = powm(H, -phi) @ X
-        test = np.linalg.norm(H - np.eye(n_channels)) / np.sqrt(n_channels)
+
+        crit = np.linalg.norm(H - np.eye(n_channels)) / np.sqrt(n_channels)
+        if crit <= zeta:
+            break
```
agramfort, line 384: usually I would advise to use an else block after the for
https://book.pythontips.com/en/latest/for_-_else.html#else-clause
and raise a warning if convergence was no reached.

pyriemann/utils/mean.py
```diff
@@ -641,12 +646,10 @@ def maskedmean_riemann(covmats, masks, tol=10e-9, maxiter=50, init=None,
     else:
         C = init
 
-    k = 0
     nu = 1.0
     tau = np.finfo(np.float64).max
     crit = np.finfo(np.float64).max
-    while (crit > tol) and (k < maxiter) and (nu > tol):
-        k = k + 1
+    for k in range(maxiter):
```
agramfort, line 652:
```suggestion
    for _ in range(maxiter):
```
as I think k is not used

pyriemann/utils/mean.py
```diff
@@ -368,15 +370,18 @@ def mean_power(covmats, p, *, sample_weight=None, zeta=10e-10):
     else:
         X = sqrtm(G)
 
-    test = 10 * zeta
-    while test > zeta:
+    crit = 10 * zeta
+    for k in range(maxiter):
```
agramfort, line 374:
```suggestion
    for _ in range(maxiter):
```

pyriemann/utils/mean.py
```diff
@@ -285,26 +270,22 @@ def mean_logdet(covmats, tol=10e-5, maxiter=50, init=None, sample_weight=None):
...
-    while (crit > tol) and (k < maxiter):
-        k = k + 1
-
-        J = np.zeros((n_channels, n_channels))
-
-        for index, Ci in enumerate(covmats):
-            J += sample_weight[index] * np.linalg.inv(0.5 * Ci + 0.5 * C)
 
+    crit = np.finfo(np.float64).max
+    for k in range(maxiter):
```
agramfort, line 279:
```suggestion
    for _ in range(maxiter):
```

pyriemann/utils/mean.py
```diff
@@ -285,26 +270,22 @@ def mean_logdet(covmats, tol=10e-5, maxiter=50, init=None, sample_weight=None):
...
+        icovmats = np.linalg.inv(0.5 * covmats + 0.5 * C)
+        J = np.einsum('a,abc->bc', sample_weight, icovmats)
         Cnew = np.linalg.inv(J)
-        crit = np.linalg.norm(Cnew - C, ord='fro')
 
+        crit = np.linalg.norm(Cnew - C, ord='fro')
         C = Cnew
+        if crit <= tol:
+            break
```
agramfort, line 287: else: with warning?

pyriemann/utils/mean.py
```diff
@@ -434,50 +411,43 @@ def mean_riemann(covmats, tol=10e-9, maxiter=50, init=None,
...
         crit = np.linalg.norm(J, ord='fro')
         h = nu * crit
-        C = C12 @ expm(nu * J) @ C12
         if h < tau:
             nu = 0.95 * nu
             tau = h
         else:
             nu = 0.5 * nu
+        if crit <= tol or nu <= tol:
+            break
```
agramfort, line 450: else with warning?

pyriemann/utils/mean.py
```diff
@@ -520,29 +490,23 @@ def mean_wasserstein(covmats, tol=10e-4, maxiter=50, init=None,
...
-    while (crit > tol) and (k < maxiter):
-        k = k + 1
-
-        J = np.zeros((n_channels, n_channels))
-
-        for index, Ci in enumerate(covmats):
-            J += sample_weight[index] * sqrtm(K @ Ci @ K)
 
+    crit = np.finfo(np.float64).max
+    for k in range(maxiter):
```
agramfort, line 500:
```suggestion
    for _ in range(maxiter):
```

Issue #46, opened 2017-01-27
Add: Spectral Transfer using Information Geometry


aj-ptw:
Please add STIG

http://journal.frontiersin.org/article/10.3389/fnins.2016.00430/full

alexandrebarachant:
Thanks AJ, 
Once you are done, open a pull request.

I can see two usecase :
- Fit a bunch of different pipeline on the same data, and combine preds with STIG. This is typically like the votting classifier from sklearn (your code is based on that)
- Fit different estimators on different subjects, and combine preds with STIG, like in the paper.

so IMO, your code should implement a fit function, but should also be able to take a list of already fitted estimators.

aj-ptw:
I can't understand the first use case, STIG specifically does not train on the same data and the paper surmises that the classifiers must be of the same type. In other words I don't understand why there would be a fit function. I'm thinking of implementing this as just calling predict over and over and never worry about fit.

alexandrebarachant:
Is it a theoritical limitation i.e. the method will not work if the estimators are fitted to the same data or if they are from different type ?

I'm thinking about using this method for ensembling (i.e better than majority voting), like it is done in the original paper: https://arxiv.org/pdf/1303.3257v3.pdf

@nwayt001 any comments ?

aj-ptw:
@alexandrebarachant (or @nwayt001) should i do [follow this advice](https://github.com/nwayt001/Transfer_Learning_Project/blob/master/Online%20Experiment/BCI2000/prog/matlab_RSVP/tools/applySML.m#L4) to:

>These lines remove classifiers who make only positive predictions or only negative predictions; it causes issues with the eigendecomposition if they are not removed; we assign a weighting of zero to these classifiers

alexandrebarachant:
This is a problem when working with prediction instead of classficiation score (like proba). A biased classifier will return always the same class.
This will in return decrease the rank of the prediction matrix and will make the eigendecomposition crash.

you should either throw an error, or remove the line.

aj-ptw:
> you should either throw an error, or remove the line.

@alexandrebarachant i removed the classifiers from the covariance matrix and did essentially [check it out](https://github.com/aj-ptw/pyRiemann/blob/3f183887568cb141385377a52799276895e53cb5/pyriemann/ensemble.py#L243-L268)

At this point i have written all the code that Nick did in his Transfer Learning Project, unfortunately there was no code for the EM step 👎 so going to have to slice and dice what we have to add that.

[...]

PR #189, opened 2022-07-12
(MRG) Transfer Learning in pyRiemann

plcrodrigues:
Hi, everyone

This is my first try to integrate some transfer learning methods to pyRiemann. @sylvchev and Emmanuel Kalunga had already started contributing on this topic with their PR #177 on MDWM, but I think it would be nice for us to discuss more broadly the general lines of development for transfer learning in our beloved package.

@agramfort @qbarthelemy and myself had a Zoom meeting recently in which we sketched some aspects of how an API for transfer learning in pyRiemann should look like. I tried implementing these ideas and had to make some design choices. The most relevant things are:

- Our data is always described by a triplet `(X, y, meta)` where `X` has the SPD matrices from all domains and `y` their corresponding labels. The `meta` structure is a `pandas` dataframe with two columns: `domain` for indicating to which domain each point belongs to and `target` indicating for each sample whether it is part of the Target domain or not.
- I had to create a `TLSplitter` object capable of splitting the data into training/validation partition in cross-validation. We can not use simply the ones from `sklearn` because we need the information from `meta`
- The way I see things, transfer learning procedures should be split into two parts: first we transform the data points and then we run classification. With this in mind, I created two examples of transformers `DCT` and `RCT` and a new object called `TLPipeline` that allows us to run the transfer learning procedure with a given classifier of choice.

I am, of course, open to suggestions and remarks and would like to know what you guys think of this first proposal.

Cheers,
Pedro


pyriemann/transferlearning.py
```diff
@@ -0,0 +1,144 @@
...
+            yield train_idx, test_idx
+
+    def get_n_splits(self, X, y, meta):
+        return self.n_splits
+
+
+class DCT(BaseEstimator, TransformerMixin):
+    '''
+    No transformation of the data points between the domains.
+    This is what we call the direct (DCT) method.
```
qbarthelemy, line 35: Class name `DCT` reminds us of Discrete Cosinus Transform...
What about another name? `DummyAdapter`?

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,144 @@
...
+    '''
+
+    def __init__(self):
+        pass
+
+    def fit(self, X, y, meta):
+        self._Minvsqrt = {}
+        for domain in np.unique(meta['domain']):
+            domain_idx = meta['domain'] == domain
+            M = mean_riemann(X[domain_idx])
```
qbarthelemy, line 63: You could use `mean_covariance` with an argument `metric` set by default to 'riemann'.

pyriemann/transferlearning.py
```diff
@@ -0,0 +1,144 @@
...
+        self.clf.fit(X_train, y_train)
+        return self
+
+    def predict(self, X, meta):
+        X_transf = self.transformer.transform(X, meta=meta)
+        X_test = X_transf
+        return self.clf.predict(X_test)
+
+    def score(self, X, y, meta):
+        from sklearn.metrics import accuracy_score
```
qbarthelemy, line 141: Should be imported at the beginning of the file.

pyriemann/transferlearning.py
qbarthelemy: In fact, `preprocessing.Whitening.fit()` and `transform` could be re-used here.

qbarthelemy (changes requested):
Documentation builds during 6 hours. This must be fixed.

Also, I suggest to re-use `preprocessing.Whitening` in `RCT`, to avoid code redundancy.

qbarthelemy:
Transfer Learning framework should also be able to integrate STIG (see issue #46 and PR #47).

plcrodrigues:
I agree, but I'm afraid that if we start adding RPA, TLMDM and STIG in the same PR we'll get drowned, no? That said, I'm not sure how it would be best to split the contributions.

qbarthelemy:
I agree to add RPA and TLMDM in a first time.

pyriemann/transferlearning.py
sylvchev: I changed the name to `TLDummy` in transferlearning_yenc

plcrodrigues:
Help, I don't know how to fix this `git` problem...

pyriemann/transferlearning_meta.py
```diff
@@ -0,0 +1,122 @@
...
+class RCT(BaseEstimator, TransformerMixin):
+    '''
+    Re-center (RCT) the data points from each domain to the Identity.
+    '''
+
+    def __init__(self):
+        pass
+
+    def fit(self, X, y, meta):
+        self.whitening_ = {}
```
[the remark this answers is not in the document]
plcrodrigues: I think we can leave the `meta` alternative as is for now, since we're not evolving it at the moment...

Also, I am having 2nd thoughts about calling `whitening` the function/attribute related to the re-centering. Yeah, of course, the operation corresponds to a whitening and it is clear for someone coming from a multivariate signal processing or machine learning background. However, for people with a Riemannian geometry flavor (which are suppose to be our target public too) this operation is a re-centering to the Identity (or a parallel transport to be more rigorous) and the word `whitening` does not mean anything in particular... Since RPA is based on geometric transformations of data points in a manifold, I would be more inclined to leave the names of its operations as close as possible to geometric ones and not their signal processing counterparts.

pyriemann/utils/rotate.py
```diff
@@ -0,0 +1,290 @@
...
+def _grad_rie(Q, X, Y, weights=None):
+
+    """
+    Gradient of loss function using Riemannian distances between class means.
+    """
+
+    def _grad_i_rie(Q, Xi, Yi):
+        M = np.linalg.inv(Xi) @ Q @ Yi @ Q.T
+        w, v = np.linalg.eig(M)
+        logM = v @ np.diag(np.log(w)) @ np.linalg.inv(v)
```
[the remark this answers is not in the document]
plcrodrigues: It is a matrix with positive eigenvalues but not symmetric. Using `utils.base.logm` throws an error because it is implemented using `np.linalg.eigh` which gives rubbish when you input a non-hermitian matrix.

pyriemann/utils/rotate.py
```diff
@@ -0,0 +1,290 @@
...
+
+    if len(X) != len(Y):
+        raise ValueError("The number of classes in each domain don't match")
+
+    if weights is None:
+        weights = np.ones(len(X)) / len(X)
+
+    G = np.zeros_like(Q)
+    for i in range(len(X)):
+        G = G + weights[i] * _grad_i_euc(Q, X[i], Y[i])
```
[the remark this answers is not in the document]
plcrodrigues: I find this option a bit too cryptic and rather hard for someone to understand that the expression corresponds to the gradient of the loss function on each of the terms of the sum.

pyriemann/transferlearning_meta.py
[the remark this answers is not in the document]
plcrodrigues: Why not call the private attribute `_recenter` but be sure to use the `Whitening` step? Another option would be to create a `ParallelTransport` class that sends points from one place to another in the SPD manifold. The `Whitening` and `ReCentering` would simply be methods that take points to the Identity, but other works like this one [1](https://arxiv.org/pdf/1807.10479.pdf) would send the dataset elsewhere. Wdyt ?

pyriemann/utils/rotate.py
```diff
@@ -0,0 +1,249 @@
...
+        # test if the step size is small
+        crit = np.linalg.norm(-alpha * direction)
+        if crit <= tol_step:
+            break
+
+        # update variables for next iteration
+        Qk_1 = Qk
+        Fk_1 = Fk
+
+    return Qk
```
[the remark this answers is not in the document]
plcrodrigues: Humm, I'm not sure that adding this `else` would be enough to have a control of whether the algorithm has converged. I have not added this feature to the procedure... it always stops after the `maxiter` iterations.

plcrodrigues:
I don't get why the tests fail for `flake8` at `pyriemann/utils/rotate.py:154:16: W605 invalid escape sequence '\s'
` when it does not fail, for instance, in [here](https://github.com/pyRiemann/pyRiemann/blob/aff4f500a8ca9cd430ef27652983d55da26de18c/pyriemann/utils/mean.py#L149).

pyriemann/transferlearning_meta.py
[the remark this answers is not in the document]
qbarthelemy: Ok for me to use `_recenter` as private attribute.

> Another option would be to create a ParallelTransport class that sends points from one place to another in the SPD manifold. The Whitening and ReCentering would simply be methods that take points to the Identity, but other works like this one [1](https://arxiv.org/pdf/1807.10479.pdf) would send the dataset elsewhere. Wdyt ?

Yes, great idea, but in another PR I suggest.

Btw, do you know function `transport`, never added in the API?
https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/utils/tangentspace.py#L375
It could be great to develop the parallel transport functions / classes of pyRiemann. You seem the good guy to do that ;-)

pyriemann/datasets/simulated.py
```diff
@@ -78,8 +79,8 @@ def make_masks(n_masks, n_dim0, n_dim1_min, rs):
 
 
 def make_gaussian_blobs(n_matrices=100, n_dim=2, class_sep=1.0, class_disp=1.0,
-                        return_centers=False, random_state=None, *,
-                        mat_mean=.0, mat_std=1., n_jobs=1,
+                        return_centers=False, center_dataset=True,
+                        random_state=None, input_centers=None, *, n_jobs=1,
```
qbarthelemy, line 83: This modif breaks API...

pyriemann/datasets/simulated.py
```diff
@@ -98,14 +99,15 @@ def make_gaussian_blobs(n_matrices=100, n_dim=2, class_sep=1.0, class_disp=1.0,
         Parameter controlling the separability of the classes.
     class_disp : float, default=1.0
         Intra dispersion of the points sampled from each class.
+    input_centers : ndarray, shape (2, n_dim, n_dim), default=None
```
qbarthelemy, line 102: My advice is to call this parameter `centers`, like [make_blobs](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.make_blobs.html).

pyriemann/datasets/simulated.py
plcrodrigues: I'm not sure I understand the problem. Yeah, I took the `mat_mean` and `mat_std` out of the code but thought I had done it in a clean way so to not have any conflicts and breaks. You think it would be necessary to keep them?

plcrodrigues: I'm not sure I understand the problem. Yeah, I took the mat_mean and mat_std out of the code but thought I had done it in a clean way so to not have any conflicts and breaks. You think it would be necessary to keep them?

qbarthelemy: These parameters are keywords only, so I think it's ok.
I prefer `centers` argument, because without it, I was unable to generate an appropriate dataset for #200.

pyriemann/transferlearning_meta.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def transform(self, X, y=None, meta=None):
+        return X
+
+    def fit_transform(self, X, y, meta):
+        return self.fit(X, y, meta).transform(X, y, meta)
+
+
+class RCT(BaseEstimator, TransformerMixin):
+    '''
```
agramfort, line 54: pleae use """ and not ''' for docstrings

agramfort (commented):
I pushed a commit I made while reviewing locally. See XXX

now I don't like the new files transferlearning_meta and yenc

should we rather a new submodule called pyriemann.transfer and users will just import stuff from there?

plcrodrigues:
Thanks @agramfort and @qbarthelemy for your inputs! I've included all your reviews and comments.

You will notice now that we only have a submodule called `transfer` and that I have abandoned the option of doing transfer learning with the `meta` dictionary. 

**Important question** to both of you: what do you think of the rather non-canonical way in which we have defined our pipelines for transfer learning? See for instance `TLCenter`, `TLStretch` and `TLRotate`. You will notice that if we do `fit` and then `transform` the results are different from `fit_transform`. This was the sole approach that we have found for making sure that the information about the domains of the data points (which are encoded in the labels) were used for calculating all of the required quantities for each domain (e.g. the means, the dispersions, or the rotation matrices).

examples/transfer/plot_rpa_steps.py
```diff
@@ -0,0 +1,205 @@
+"""
+====================================================================
+Plot the data transformations in the Riemannian Procrustes Analysis
+====================================================================
+
+Use the SpectralEmbedding module to plot in 2D the transformations on the data
+points from source and target domains when applying the Riemannian Procrustes
+Analysis to match their statistics.
```
agramfort, line 8: you should add a ref to the RPA paper here

examples/transfer/plot_rpa_steps.py
```diff
@@ -0,0 +1,205 @@
...
+
+from pyriemann.transfer import (
+    encode_domains,
+    decode_domains,
+    TLCenter,
+    TLRotate,
+)
+
+
+def make_example_transfer_learning(N, class_sep=3.0, class_disp=1.0,
```
agramfort, line 28: this function is really copy pasted (I think) in the two examples. Should have a public function for this in simulated.py ?

pyriemann/utils/rotate.py
```diff
@@ -0,0 +1,233 @@
+import warnings
+import numpy as np
+from ..utils.distance import distance
```
agramfort, line 3: I feel this utils submodule is getting too crowded. I would create an pyriemann/transfer folder and put this file in pyriemann/transfer/_rotate.py
(starting with _ if it's meant to be private functions only).

agramfort (commented):
@plcrodrigues this is taking shape !

now I think we need next to add a test_transfer.py file and start properly testing these new features.

thanks @plcrodrigues for leading this !

plcrodrigues:
> now I think we need next to add a test_transfer.py file and start properly testing these new features.

Indeed, I will propose a few tests in my next commit.

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,512 @@
...
+    ----------
+    target_domain : str
+        Which domain to consider as target
+    dispersion : float, default=1.0
+        Which target value for the dispersion of the data points
+    centered_data : bool, default=False
+        Whether the data has been re-centered to the Identity beforehand
+    """
+
+    def __init__(self, target_domain, final_dispersion=1.0,
```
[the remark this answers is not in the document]
plcrodrigues: Mainly because the operation for "stretch" as we know it from the RPA paper is only defined for the Riemannian distance. With that said, we could define it for the Euclid distance and for the LogEuclid ones. However, I am not sure we can do anything very easy for all of the metrics available in `pyriemann`. Anyways, do you think it is really worth defining these variations for stretch?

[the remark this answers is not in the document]
plcrodrigues: Well, I've added the option of choosing any metric for calculating the dispersion of the SPD matrices. Note, however, that the stretching is still in "Riemannian" terms since an Euclidean stretch would only make sense if we re-centered the data points to zero, which is not SPD...

examples/transfer/plot_comparison_pipelines.py
```diff
@@ -0,0 +1,213 @@
...
+        # the classifier is trained with points from source and target
+        X_enc_train_rpa, y_enc_train_rpa = X_enc_train, y_enc_train
+
+        # fit the pipeline
+        pipeline.fit(X_enc_train_rpa, y_enc_train_rpa)
+
+        # get the accuracy score
+        scores_cv['rpa'].append(pipeline.score(X_enc_test, y_enc_test))
+
+        # (4) tlmdm pipeline -- ??
```
agramfort, line 161:
```suggestion
        # (4) TLMDM pipeline -- ??
```

examples/transfer/plot_rpa_steps.py
```diff
@@ -0,0 +1,109 @@
...
+)
+
+
+# fix seed for reproducible results
+seed = 66
+
+# create source and target datasets
+N = 50
+X_enc, y_enc = make_example_transfer_learning(
+    N=N,
```
agramfort: I don’t like much the N parameter. It should n_matrices ? I don’t remember the naming we agreed

pyriemann/transfer/_rotate.py
```diff
@@ -0,0 +1,191 @@
...
+    Returns
+    -------
+    Q : ndarray, shape (n, n)
+        Orthogonal matrix that minimizes the distance between the class means
+        for the source and target domains.
+
+    References
+    ----------
+    .. [1] P. Rodrigues et al. "Riemannian Procrustes Analysis: Transfer
+        Learning for Brain-Computer Interfaces" (2018)
```
agramfort: Can you write the full ref?

pyriemann/utils/viz.py
```diff
@@ -244,7 +244,7 @@ def plot_waveforms(X, display, *, times=None, color='gray', alpha=0.5,
 
 
 def _add_alpha(colors, alphas):
-    ''' Add alphas to RGB channels '''
+    """ Add alphas to RGB channels """
```
agramfort, line 247:
```suggestion
    """Add alphas to RGB channels."""
```

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,514 @@
...
+        """Init"""
+        self.target_domain = target_domain
+        self.final_dispersion = final_dispersion
+        self.centered_data = centered_data
+
+    def fit(self, X, y):
+        _, _, domains = decode_domains(X, y)
+        m = X[0].shape[1]
+        self._means = {}
+        self._dispersions = {}
```
agramfort, line 118:
```suggestion
        self.dispersions_ = {}
```

To follow sklearn. Also you will need to doucement the attributes

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,514 @@
...
+    follows the presentation from [2]_.
+
+    Important: the data points from each domain must have been re-centered
+    to the identity before calculating the rotation.
+
+    Parameters
+    ----------
+    target_domain : str
+        Domain to consider as target.
+    weights : None | array, shape (n_classes), default=None
```
agramfort, line 199:
```suggestion
    weights : None | array, shape (n_classes,), default=None
```

agramfort (commented):
A next batch of comments

examples/transfer/plot_rpa_steps.py
plcrodrigues: Yeah, it's already corrected to `n_matrices`

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,514 @@
...
+from ..utils.mean import mean_covariance, mean_riemann
+from ..utils.distance import distance
+from ..utils.base import invsqrtm, powm, sqrtm
+from ..utils.geodesic import geodesic
+from ._rotate import get_rotation_matrix
+from ..classification import MDM, _check_metric
+from ..preprocessing import Whitening
+from .methods import decode_domains
+
+base_clf = MDM()
```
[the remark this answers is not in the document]
plcrodrigues: I agree, but it was @sylvchev who included this `base_clf` for some reason that I don't remember anymore. Sylvain, could explain us the problem with just passing `MDM()` as a default parameter in `TLClassifier` ?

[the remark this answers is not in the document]
agramfort: it's bad practice to pass mutable parameters as default parameters in function signatures. That's certainly the reason but this is not great either. I would make the default None in the function and instantiate the default model in the fit when needed.

examples/transfer/plot_rpa_steps.py
```diff
@@ -0,0 +1,109 @@
...
+Plot the data transformations in the Riemannian Procrustes Analysis
+====================================================================
+
+Use the SpectralEmbedding module to plot in 2D the transformations on the data
+points from source and target domains when applying the Riemannian Procrustes
+Analysis [1] to match their statistics.
+
+    .. [1] PLC Rodrigues et al “Riemannian Procrustes analysis: transfer
+    learning for brain-computer interfaces”. IEEE Transactions on Biomedical
+    Engineering, vol. 66, no. 8, pp. 2390-2401, December, 2018
```
agramfort, line 12: See how sklearn does it:

https://github.com/scikit-learn/scikit-learn/blob/main/examples/calibration/plot_compare_calibration.py#L197

```suggestion
.. [1] `Riemannian Procrustes analysis: transfer learning for
       brain-computer interfaces <https://ieeexplore.ieee.org/document/8588384>`_,
       PLC Rodrigues, C Jutten, M Congedo
       IEEE Transactions on Biomedical Engineering, vol. 66, no. 8,
       pp. 2390-2401, December, 2018
```

pyriemann/transfer/pipelines.py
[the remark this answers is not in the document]
sylvchev: Exactly, I got a big warning for passing a classifier in the default parameter, and I agree this is not a good solution to define it in the module. I like your solution, I'll push it.

examples/transfer/plot_rpa_steps.py
plcrodrigues: Well, you activated my OCD... I updated the reference style in all files of `pyRiemann`...

examples/transfer/plot_rpa_steps.py
```diff
@@ -0,0 +1,111 @@
+"""
+====================================================================
+Plot the data transformations in the Riemannian Procrustes Analysis
+====================================================================
+
+Use the SpectralEmbedding module to plot in 2D the transformations on the data
+points from source and target domains when applying the Riemannian Procrustes
+Analysis [1]_ to match their statistics.
+
+    .. [1] `Riemannian Procrustes analysis: transfer learning for
```
agramfort, line 10:
```suggestion
.. [1] `Riemannian Procrustes analysis: transfer learning for
```

examples/transfer/plot_rpa_steps.py
```diff
@@ -0,0 +1,111 @@
...
+    TLRotate,
+)
+
+
+# fix seed for reproducible results
+seed = 66
+
+# create source and target datasets
+n_matrices = 50
+X_enc, y_enc = make_example_transfer_learning(
```
agramfort, line 34:
```suggestion
X_enc, y_enc = make_classification_transfer(
```

I would prefer to avoid _example and it's information to say it's for classification.

pyriemann/clustering.py
```diff
@@ -305,14 +305,18 @@ class Potato(BaseEstimator, TransformerMixin, ClassifierMixin):
...
+    A. Barachant, A Andreev, and M. Congedo. TOBI Workshop lV, Jan 2013, Sion,
+    Switzerland. pp.19-20.
+
+    .. [2] `The Riemannian Potato Field: A Tool for Online Signal Quality Index
+    of EEG
+    <https://hal.archives-ouvertes.fr/hal-02015909>`_
+    Q. Barthélemy, L. Mayaud, D. Ojeda, and M. Congedo. IEEE Transactions
+    on Neural Systems and Rehabilitation Engineering, IEEE Institute of
+    Electrical and Electronics Engineers, 2019, 27 (2), pp.244-255
     """
```
agramfort, line 320: check your indentation. When you write .. in ReStructuredText the next line should be indented with block.

pyriemann/clustering.py
```diff
@@ -550,9 +554,12 @@ class PotatoField(BaseEstimator, TransformerMixin, ClassifierMixin):
...
     ----------
-    .. [1] Q. Barthélemy, L. Mayaud, D. Ojeda, M. Congedo, "The Riemannian
-        potato field: a tool for online signal quality index of EEG", IEEE
-        TNSRE, 2019.
+    .. [1] `The Riemannian Potato Field: A Tool for Online Signal Quality Index
+    of EEG
+    <https://hal.archives-ouvertes.fr/hal-02015909>`_
+    Q. Barthélemy, L. Mayaud, D. Ojeda, and M. Congedo. IEEE Transactions
+    on Neural Systems and Rehabilitation Engineering, IEEE Institute of
+    Electrical and Electronics Engineers, 2019, 27 (2), pp.244-255
```
agramfort, line 562: idem .. -> indentation

pyriemann/datasets/simulated.py
agramfort: +1

agramfort (commented):
next batch of comments :)

pyriemann/spatialfilters.py
```diff
@@ -272,22 +275,26 @@ class CSP(BilinearFilter):
...
-        on Multimedia Signal Processing (MMSP), p. 472-476, 2010.
-
-    .. [4] Grosse-Wentrup, Moritz, and Martin Buss. "Multiclass common spatial
-        patterns and information theoretic feature extraction." Biomedical
-        Engineering, IEEE Transactions on 55, no. 8 (2008): 1991-2000.
+    .. [1] `Spatial Patterns Underlying Population Differences in the
+        Background EEG
+        <https://link.springer.com/article/10.1007/BF01129656>`_
+        Z. Koles, M. Lazar, and S. Zhou. Brain Topography 2(4), 275-284, 1990.
+
```
qbarthelemy, line 282:
```suggestion
```

pyriemann/spatialfilters.py
```diff
@@ -272,22 +275,26 @@ class CSP(BilinearFilter):
...
+    .. [2] `Optimizing Spatial Filters for Robust EEG Single-Trial Analysis
+        <https://ieeexplore.ieee.org/document/4408441>`_
+        B. Blankertz, R. Tomioka, S. Lemm, M. Kawanabe, K-R. Muller. IEEE
+        Signal Processing Magazine 25(1), 41-56, 2008.
+
```
qbarthelemy, line 287:
```suggestion
```

pyriemann/spatialfilters.py
```diff
@@ -272,22 +275,26 @@ class CSP(BilinearFilter):
...
+    .. [3] `Common Spatial Pattern revisited by Riemannian geometry
+        <https://hal.archives-ouvertes.fr/hal-00602686>`_
+        A. Barachant, S. Bonnet, M. Congedo and C. Jutten. IEEE International
+        Workshop on Multimedia Signal Processing (MMSP), p. 472-476, 2010.
+
```
qbarthelemy, line 292:
```suggestion
```

pyriemann/spatialfilters.py
```diff
@@ -565,16 +573,23 @@ class AJDC(BaseEstimator, TransformerMixin):
...
+        <https://hal.archives-ouvertes.fr/hal-00343628>`_
+        M. Congedo, C. Gouy-Pailler, C. Jutten. Clinical Neurophysiology,
+        Elsevier, 2008, 119 (12), pp.2677-2686.
+
+    .. [2] `Group indepedent component analysis of resting state EEG in large
+        normative samples
+        <https://hal.archives-ouvertes.fr/hal-00523200>`_
+        M. Congedo, R. John, D. de Ridder, L. Prichep. International Journal of
+        Psychophysiology, Elsevier, 2010, 78, pp.89-99.
+
```
qbarthelemy, line 587:
```suggestion
    .. [2] `Group indepedent component analysis of resting state EEG in large
        normative samples
        <https://hal.archives-ouvertes.fr/hal-00523200>`_
        M. Congedo, R. John, D. de Ridder, L. Prichep. International Journal of
        Psychophysiology, Elsevier, 2010, 78, pp.89-99.
```

pyriemann/tangentspace.py
```diff
@@ -67,13 +67,17 @@ class TangentSpace(BaseEstimator, TransformerMixin):
...
-
-    .. [2] A. Barachant, S. Bonnet, M. Congedo and C. Jutten, "Classification
-        of covariance matrices using a Riemannian-based kernel for BCI
-        applications", in NeuroComputing, vol. 112, p. 172-178, 2013.
+    .. [1] `Multiclass Brain-Computer Interface Classification by Riemannian
+        Geometry
+        <https://hal.archives-ouvertes.fr/hal-00681328>`_
+        A. Barachant, S. Bonnet, M. Congedo, and C. Jutten. IEEE Transactions
+        on Biomedical Engineering, vol. 59, no. 4, p. 920-928, 2012.
+
```
qbarthelemy, line 75:
```suggestion
```

pyriemann/tangentspace.py
```diff
@@ -254,14 +258,17 @@ class FGDA(BaseEstimator, TransformerMixin):
...
-
-    .. [2] A. Barachant, S. Bonnet, M. Congedo and C. Jutten, "Classification
-        of covariance matrices using a Riemannian-based kernel for BCI
-        applications", in NeuroComputing, vol. 112, p. 172-178, 2013.
+    .. [1] `Riemannian geometry applied to BCI classification
+        <https://hal.archives-ouvertes.fr/hal-00602700/>`_
+        A. Barachant, S. Bonnet, M. Congedo and C. Jutten. 9th International
+        Conference Latent Variable Analysis and Signal Separation
+        (LVA/ICA 2010), LNCS vol. 6365, 2010, p. 629-636.
+
```
qbarthelemy, line 266:
```suggestion
```

qbarthelemy (commented):
Thx for references cleaning. However, indentations must contain four spaces.

pyriemann/datasets/simulated.py
```diff
@@ -135,44 +138,63 @@ def make_gaussian_blobs(n_matrices=100, n_dim=2, class_sep=1.0, class_disp=1.0,
...
+    # concatenate the samples
     X = np.concatenate([X0, X1])
-    y = np.concatenate([y0, y1])
+
+    if center_dataset:
+        # re-center the dataset to the Identity
+        M = mean_riemann(X)
+        M_invsqrt = invsqrtm(M)
+        for i in range(n_matrices):
+            X[i] = M_invsqrt @ X[i] @ M_invsqrt
```
qbarthelemy, line 185:
```suggestion
        X = M_invsqrt @ X @ M_invsqrt
```

pyriemann/transfer/methods.py
```diff
@@ -0,0 +1,127 @@
...
+    domain : ndarray, shape (n_matrices,)
+        Which domain each data point belongs to.
+
+    Returns
+    -------
+    X_enc : ndarray, shape (n_matrices, n_channels, n_channels)
+        The same set of SPD matrices given as input.
+    y_enc : ndarray, shape (n_matrices,)
+        Extended labels for each data point.
+    """
```
qbarthelemy, line 30: Should be added in all functions/classes added in API.
```suggestion
    Notes
    -----
    .. versionadded:: 0.3.1
    """
```

pyriemann/datasets/simulated.py
```diff
@@ -78,8 +80,8 @@ def make_masks(n_masks, n_dim0, n_dim1_min, rs):
 
 
 def make_gaussian_blobs(n_matrices=100, n_dim=2, class_sep=1.0, class_disp=1.0,
-                        return_centers=False, random_state=None, *,
-                        mat_mean=.0, mat_std=1., n_jobs=1,
+                        return_centers=False, center_dataset=True,
+                        random_state=None, centers=None, *, n_jobs=1,
```
qbarthelemy, line 84:
```suggestion
                        return_centers=False, random_state=None, *,
                        centers=None, mat_mean=.0, mat_std=1., center_dataset=False, n_jobs=1,
```
I suggest a modification ensuring full backward compatibility for `make_gaussian_blobs`.
When `centers=None` (default), first center `C0` is not defined as identity, but it is computed as previously using `mat_mean` and `mat_std` parameters.

```
    if centers is None:
        C0 = generate_random_spd_matrix(
            n_dim,
            random_state=random_state,
            mat_mean=mat_mean,
            mat_std=mat_std
        )
        epsilon = np.exp(class_sep / np.sqrt(n_dim))
        C1 = epsilon * C0
```

Moreover, `center_dataset` should be False by default.

plcrodrigues: I see what you mean, but the problem is that with your solution we won't have full control of the distance between `C0` and `C1`. Yeah, of course, the `epsilon` is a proxy... but it is not the *exact* value for it. In my current proposal, the `class_sep` is exactly the distance between the two class. This is only possible because we fix `C0` to the Identity and then modulate the distance of `C1` with a stretching. 

Also, the `mat_mean` and `mat_std` parameters are parameters which are totally fine for generating random matrices, but we don't know exactly what to expect from the distribution of the matrices that will get sampled with them. This is why I thought they could be dropped (and too bad for backward compatibility)

pyriemann/datasets/simulated.py
```diff
@@ -135,44 +138,62 @@ def make_gaussian_blobs(n_matrices=100, n_dim=2, class_sep=1.0, class_disp=1.0,
...
+    # randomly permute the samples of the dataset
     idx = rs.permutation(len(X))
-    X = X[idx]
-    y = y[idx]
+    X, y = X[idx], y[idx]
 
     if return_centers:
+        if centers is None:
+            C0 = mean_riemann(X[y == 0])
+            C1 = mean_riemann(X[y == 1])
```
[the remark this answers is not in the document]
plcrodrigues: The initial matrices `C0` and `C1` defined as above are used to sample the Riemannian Gaussian distribution. However, once we re-center the dataset, the centers of class 0 and class 1 (which we call `C0` and `C1`) will change.

tests/test_transfer.py
```diff
@@ -0,0 +1,71 @@
+import numpy as np
+
+from pyriemann.datasets.simulated import make_classification_transfer
+
+from pyriemann.transfer import (
+    TLCenter,
+    TLStretch,
+    TLRotate,
+    decode_domains
```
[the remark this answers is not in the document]
qbarthelemy: You can have a look to `test_classification`.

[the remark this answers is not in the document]
plcrodrigues: @qbarthelemy thanks for your input in the tests! I had to change a few things to make things pass though. 

Most notably, there is no reason to expect that with `metric = euclid` the Whitening step should make the means on each domain of the dataset exactly Identity. In fact, mathematically it does not make much sense to do such an operation with any other metric than `riemann` (Whitening = Parallel Transport with the AIRM). I am not sure if the tests for `TLCenter` and `TLStretch` will ever pass for other metrics than `riemann`...

agramfort:
I just tried to use this branch with @apmellot for a regression task and even using TLCenter does not work as y is not an int. Following sklearn I really don't see why class should be forced to be an int...

plcrodrigues:
I agree and I don't know why we added this restriction to `int` values... will check it.

agramfort:
ok so I play a bit with this PR. A few remarks:

- to follow sklearn TLStratifiedShuffleSplitter should be renamed TLStratifiedShuffleSplit
- we should add a TLRegressor or at least anticipate if the proposed API prevents us from addressing regression usecases in the future

a few more things are bugging me but it's better to talk about this. @plcrodrigues let me know when you have a moment to talk

tests/test_transfer.py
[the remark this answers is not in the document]
qbarthelemy: Ok! Don't forget to update documentation.

tests/test_transfer.py
```diff
@@ -0,0 +1,191 @@
...
+    cv = TLSplitter(
+        target_domain="target_domain",
+        cv_iterator=KFold(n_splits=5, shuffle=True, random_state=rndstate))
+    train_idx, test_idx = next(cv.split(X, y_enc))
+    assert len(train_idx) == 90  # 50 from source and 4/5*100 from target
+    assert len(test_idx) == 10  # 1/5*100 from target
+    assert cv.get_n_splits() == 5
+
+
+def test_tlestimator(rndstate):
```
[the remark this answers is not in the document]
plcrodrigues: I agree and understand that is better to test with different classifiers, but the actual tests and verification for the parameters of a base classifier vs a pipeline are slightly different, so I don't see how we can put everyone on the same `estimator`

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,822 @@
...
+    Classification by nearest centroid. For each of the given classes, a
+    centroid is estimated, according to the chosen metric, as a weighted mean
+    of SPD matrices from the source domain, combined with the class centroid of
+    the target domain [1]_ [2]_.
+    For classification, a given new matrix is attibuted to the class whose
+    centroid is the nearest according to the chosen metric.
+
+    Parameters
+    ----------
+    transfer_coef : float
```
qbarthelemy, line 671: To highlight the link with `domain_weight` of `TLEstimator`,
`transfer_coef` could be renamed `domain_tradeoff`, or `target_domain_weight`.
@sylvchev

tests/test_transfer.py
```diff
@@ -0,0 +1,250 @@
...
+    assert predicted.shape == (n_matrices,)
+
+    # test predict_proba
+    probabilities = clf.predict_proba(X)
+    assert probabilities.shape == (n_matrices, n_classes)
+    assert probabilities.sum(axis=1) == approx(np.ones(n_matrices))
+
+    # test score
+    clf.score(X, y_enc)
+
```
qbarthelemy, line 250:
```suggestion
    clf.score(X, y_enc)
```

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,824 @@
...
+            for step in self.estimator.steps:
+                step_name = step[0]
+                sample_weight[step_name + '__sample_weight'] = w
+            self.estimator.fit(X_dec, y_dec, **sample_weight)
+        else:
+            self.estimator.fit(X_dec, y_dec, sample_weight=w)
+
+        return self
+
+    def predict(self, X):
```
qbarthelemy, line 618: When `estimator` is a classifier, the function `predict_proba()` should be available; not when it is a regressor.
How to define this function conditionally? @agramfort @plcrodrigues

agramfort: good catch... we can do black magic but subclassing with a TLClassifier and TLRegressor is the cleanest

qbarthelemy: Ok for subclasses, I will commit this solution.

pyriemann/datasets/simulated.py
```diff
@@ -222,3 +258,131 @@ def make_outliers(n_matrices, mean, sigma, outlier_coeff=10,
...
+    n_dim = 2
+
+    # create a source dataset with two classes and global mean at identity
+    M1_source = np.eye(n_dim)  # first class mean at Identity at first
+    X1_source = sample_gaussian_spd(
+        n_matrices=n_matrices,
+        mean=M1_source,
+        sigma=class_disp,
+        random_state=seeds[0])
+    y1_source = (1*np.ones(n_matrices)).astype(int)
```
qbarthelemy, line 319: Names of classes should be defined by a parameter, with default value `class_names = [1, 2]`.

qbarthelemy (approved):
It's ok for me.

plcrodrigues:
Thanks @qbarthelemy for taking the time of doing the final tweaks and corrections on this PR !! 

I'm eager to see it merged into the `master` branch.

Btw, for your information:
- The example showing the RPA steps reproduces the Figure 1 of [1]
- The example with the comparison between pipelines reproduces the results in Figure 2 of [1]

Once the PR is merged, I will update my RPA repository ([link](https://github.com/plcrodrigues/RPA)) to reproduce other figures from [1].

[1] PLC Rodrigues et al "Riemannian Procrustes Analysis : Transfer Learning for
Brain-Computer Interfaces" - https://hal.archives-ouvertes.fr/hal-01971856/document

agramfort:
artifact looks good https://github.com/pyRiemann/pyRiemann/suites/9452453423/artifacts/447277750

let me make a few more comments about API/DOC now

examples/transfer/plot_bci_example.py
```diff
@@ -0,0 +1,199 @@
...
+    # Compute covariance matrices on scaled data
+    covs = Covariances().fit_transform(1e6 * epochs.get_data())
+
+    return covs, labels
+
+
+###############################################################################
+
+# We will consider subjects from the Physionet EEG database for which the
+# intra-subject classification has been checked to be > 0.70
```
agramfort, line 86:
```suggestion
###############################################################################
# We will consider subjects from the Physionet EEG database for which the
# intra-subject classification has been checked to be > 0.70
```

otherwise it does not render as a clean section in the doc

examples/transfer/plot_bci_example.py
```diff
@@ -0,0 +1,199 @@
...
+        pipeline.fit(X_enc_train, y_enc_train)
+        scores_cv['rct'].append(pipeline.score(X_enc_test, y_enc_test))
+
+    for meth in scores.keys():
+        scores[meth].append(np.mean(scores_cv[meth]))
+
+
+###############################################################################
+
+# Plot results
```
agramfort, line 180:
```suggestion
###############################################################################
# Plot results
```

examples/transfer/plot_comparison_pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+    TLClassifier,
+    MDWM
+)
+
+warnings.filterwarnings("ignore")
+
+
+###############################################################################
+
+# Choose seed for reproducible results
```
agramfort, line 36:
```suggestion
###############################################################################
# Choose seed for reproducible results
```

examples/transfer/plot_comparison_pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+    TLSplitter,
+    TLDummy,
+    TLCenter,
+    TLStretch,
+    TLRotate,
+    TLClassifier,
+    MDWM
+)
+
+warnings.filterwarnings("ignore")
```
agramfort, line 31: do we really need this? hiding warnings can be dangerous

examples/transfer/plot_bci_example.py
```diff
@@ -0,0 +1,199 @@
...
+
+# Encode the data for transfer learning purposes
+X_enc, y_enc = encode_domains(X, y, domains)
+
+# Object for splitting the datasets into training and validation partitions
+n_splits = 5  # How many times to split the target domain into train/test
+cv = TLSplitter(
+    target_domain='',
+    cv_iterator=StratifiedShuffleSplit(
+        n_splits=n_splits, train_size=0.10, random_state=42))
```
agramfort, line 108:
```suggestion
cv = TLSplitter(
    target_domain='',
    cv=StratifiedShuffleSplit(
        n_splits=n_splits, train_size=0.10, random_state=42))
```

I don't really like inventing a new param name for what is just a standard sklearn cv object. I would keep the param name `cv` and just do

`tl_cv = TLSplitter(`

if you want to be clear that this is a cv object for transfer

pyriemann/transfer/_rotate.py
```diff
@@ -0,0 +1,221 @@
...
+        Q_1 = Q
+        F_1 = F
+
+    else:
+        warnings.warn('Convergence not reached.')
+
+    return Q, F
+
+
+def get_rotation_matrix(M_source, M_target, weights=None, metric='euclid',
```
agramfort, line 136: this function is not present in the API page. It means to me that it is meant to be private and therefore be called `_get_rotation_matrix`

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,949 @@
+import numpy as np
```
agramfort, line 1: this file does not really implement the pipeline object but more the estimators. I would therefore call it `_estimators.py` with _ as you are not supposed to import anything from there but I agree we are not consistent on this. It is just better practice

pyriemann/transfer/methods.py
```diff
@@ -0,0 +1,163 @@
+import numpy as np
```
agramfort, line 1: it is not really methods you implement in this file. Maybe _tools.py ?

agramfort (commented):
a new batch of comments

we are getting there !

if you have spare cycles I would encourage you to add more text in the examples. they are very dry in narrative so far

🙏

pyriemann/transfer/pipelines.py
```diff
@@ -0,0 +1,949 @@
...
+
+class TLCenter(BaseEstimator, TransformerMixin):
+    """Recenter data for transfer learning.
+
+    Recenter the data points from each domain to the Identity on manifold, ie
+    make the mean of the datasets become the identity. This operation
+    corresponds to a whitening step if the SPD matrices represent the spatial
+    covariance matrices of multivariate signals.
+
+    Important: note that using .fit() and then .transform() will give different
```
agramfort, line 96: you should use the `.. note:` directive to make it look nice in the doc

plcrodrigues:
Thanks for the feedback @agramfort ! I've included the `..note::` directive on different parts of the doc and have given a bit more explanations about each example. Hope this corresponds to what you meant!

agramfort (approved):
👏 🚀 🙏 @plcrodrigues !

PR #198, opened 2022-08-18
Faster sampling 2D Riemannian Gaussian

Artim436:
## Introduction

Hi, everyone.

In this issue, I would like to share an improvement that @plcrodrigues and myself made for `_sample_parameter_r`, which is the main bottleneck for sampling SPD matrices from a Riemannian Gaussian distribution via`sample_gaussian_spd`.

**Disclaimer**: This improvement is only for `n_dim=2` but we think it will be useful for the community, since many toy models can be reduced to the 2D case.

In summary, our idea is to use rejection sampling in `_sample_parameter_r` instead of the current implementation with slice sampling. To have a first look, here is a graph of the speedup with a fixed choice of `sigma` (i.e. the dispersion of the Gaussian distribution).

comparaison

## Theory

**Breaking the problem in two**. We want to sample the vector $r = (r_1, \dots, r_m)$ from the probability distribution

$$
p(r) = \dfrac{1}{Z(m, \sigma)} \exp\left(-\frac{1}{2\sigma^2}\sum_{i = 1}^m r^2_i\right) \times \prod_{i < j}\sinh\left(|r_i - r_j|/2\right)
$$

which has to be done via some computational method. At the moment, `pyriemann` uses a slice sampling approach [[1]](https://projecteuclid.org/journals/annals-of-statistics/volume-31/issue-3/Slice-sampling/10.1214/aos/1056562461.full) to obtain samples from $p(r)$ but this is rather slow. We suggest changing the implementation to a rejection sampling approach [[2]](https://en.wikipedia.org/wiki/Rejection_sampling), exploring certain properties of $p$ and therefore obtaining better performance.

We consider at first the simplest case for our sampling procedure, that of $m = 2$. The pdf for $r = (r_1, r_2)$ simplifies to:

$$
\begin{array}{rcl}
p(r_1, r_2) &=& \dfrac{1}{Z(2, \sigma)}\exp\left(-\dfrac{1}{2\sigma^2}(r_1^2 + r_2^2)\right) \times \sinh\left(|r_1 - r_2|/2\right) 
\end{array}
$$

we can see this pdf as a mixture of two components depending on a binary variable $b$

$$
p(r) = p(r \mid b = 0) \times \mathbb{P}(b = 0) + p(r \mid b = 1) \times \mathbb{P}(b = 1)
$$

with $\mathbb{P}(b = 0) = \mathbb{P}(b = 1) = 1/2$ and

$$
p(r \mid b = 0) = \dfrac{2}{Z(2, \sigma)}\exp\left(-\dfrac{1}{2\sigma^2}(r_1^2 + r_2^2)\right)\sinh\Big((r_1 - r_2)/2\Big) \times \mathbb{I}({\{r_1 - r_2 \geq 0\}})
$$

and

$$
p(r \mid b = 1) = \dfrac{2}{Z(2, \sigma)}\exp\left(-\dfrac{1}{2\sigma^2}(r_1^2 + r_2^2)\right) \sinh\Big((r_2 - r_1)/2\Big) \times \mathbb{I}({\{r_1 - r_2 < 0\}})
$$

where $\mathbb{I}$ is the indicator function.

Great, we see that to generate a sample from $p(r_1, r_2)$ we can first generate a Bernoulli variable $b \sim \mathcal{B}(1/2)$ and then sample from one of the conditional distributions. Now the question is: how do we sample from the conditional distributions?

**Sampling from the conditional distributions.** We can use rejection sampling to sample from $p(r \mid b = 0)$ and first we need to find a nice upper bound to it. Considering the inequality valid for all $x > 0$: 

$$
\sinh(x) = \dfrac{1}{2}\exp(x) - \dfrac{1}{2}\exp(-x) \leq \dfrac{1}{2}\exp(x)
$$

we can write

$$
p(r \mid b = 0) \leq \dfrac{1}{Z(2, \sigma)} \exp\left(-\dfrac{1}{2\sigma^2}(r_1^2 + r_2^2)\right) \exp\Big((r_1 - r_2)/2\Big) \times \mathbb{I}(\{r_1 - r_2 \geq 0\})
$$

and through some rearrangements,

$$
p(r \mid b = 0) \leq \dfrac{2\pi\sigma^2 \exp(\sigma^2/4)}{Z(2, \sigma)} \times \dfrac{1}{2\pi \sigma^2}\exp\left(-\dfrac{1}{2\sigma^2}\Big(\big(r_1-\sigma^2/2\big)^2 + \big(r_2+\sigma^2/2\big)^2\Big)\right)
$$

The expression above indicates that if we want to sample from $p(r \mid b = 0)$ we can do it via rejection sampling using as auxiliary distribution 

$$
g_+(r) = \dfrac{1}{2\pi \sigma^2}\exp\left(-\dfrac{1}{2\sigma^2}\Big(\big(r_1-\sigma^2/2\big)^2 + \big(r_2+\sigma^2/2\big)^2\Big)\right)
$$

the algorithm goes as follows:
1. Sample $u \sim \mathcal{U}(0, 1)$ and $r \sim g_+(r)$

2. Check whether 

$$
u < \exp\left(-\dfrac{1}{2\sigma^2}(r_1^2+r_2^2)\right)\sinh((r_1 - r_2)/2)\times \mathbb{I}({r_1 - r_2 \geq 0})\times\dfrac{1}{M g_+(r)}
$$

- If this holds, then accept $r$ as a sample from $p(r \mid b = 0)$
- If not, reject the sample

Where $M = \pi\sigma^2 \exp(\sigma^2/4)$.

As you can imagine, sampling from $p(r \mid b = 1)$ follows the same logic but with a different auxiliary pdf $g_-(r)$

## The algorithm

Summing up, the new implementation that we propose for `_sample_parameter_r` is based on the following algorithm:

- Sample $b \sim \mathcal{B}(1/2)$
- If $b = 0$, then sample $r \sim p(r \mid b = 0)$
- If $b = 1$, then sample $r \sim p(r \mid b = 1)$

the sampling from the conditional distributions is done following the rejection sampling procedure described above.

## Why not consider more dimensions ?

A natural question to ask is why we have not considered cases with more dimensions than just two dimensions? Well, things can get quite complicated when $m$ increases...

For instance, suppose we have $r = (r_1, r_2, r_3)$ then the pdf of interest can be written as

$$
p(r_1, r_2, r_3) = \dfrac{1}{Z(3,\sigma)}\exp\Big(-\dfrac{1}{2}(r_1^2 + r_2^2 + r_3^2)\Big) \times \sinh\Big(\dfrac{|r_1 - r_2|}{2}\Big) \times \sinh\Big(\dfrac{|r_1 - r_3|}{2}\Big) \times \sinh\Big(\dfrac{|r_2 - r_3|}{2}\Big)
$$

To use the same strategy from our 2D example, we would have to sample three Bernoulli random variables (one for each factor in the product) and consider the $2^3 = 8$ possible combinations of signs inside each of the $\sinh$. We have tried to implement this case, but our first results indicate a the probability of acceptance that is too small and makes the rejection sampling algorithm impractical.

Moreover, we see that the number of conditional distributions increases as $2^m$ where $m$ is the dimensionality of the SPD matrices being considered. Therefore, our algorithm does not look scalable for larger matrix dimensions.

## Final remarks

This is it, we have obtained a much faster implementation for sampling 2D Gaussian SPDs than what was available in `pyriemann` so far. We should mention that our implementation is based on a `while` loop that stops once we have obtained the desired number of samples. However, we can use certain properties of the rejection sampling algorithm to calculate the probability of acceptance of a sample and write code that generates several candidates in the upper hand. Such an algorithm can be even faster than the one we have implemented, but it requires approximating the normalizing constant of $p(r_1, r_2)$, which can be cumbersome. We will leave this extension for a future PR.

### DIFF
--- a/examples/simulated/plot_riemannian_gaussian.py
+++ b/examples/simulated/plot_riemannian_gaussian.py
@@ -34,18 +34,15 @@ mean = generate_random_spd_matrix(n_dim)  # random reference point
 samples_1 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=mean,
                                 sigma=sigma,
-                                random_state=random_state,
-                                sampling_method='rejection')
+                                random_state=random_state)
 samples_2 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=mean,
                                 sigma=sigma/2,
-                                random_state=random_state,
-                                sampling_method='rejection')
+                                random_state=random_state)
 samples_3 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=epsilon*mean,
                                 sigma=sigma,
-                                random_state=random_state,
-                                sampling_method='rejection')
+                                random_state=random_state)
 
 # Stack all of the samples into one data array for the embedding
 samples = np.concatenate([samples_1, samples_2, samples_3])
--- a/examples/simulated/plot_riemannian_gaussian.py
+++ b/examples/simulated/plot_riemannian_gaussian.py
@@ -23,7 +23,7 @@ print(__doc__)
 ###############################################################################
 # Set parameters for sampling from the Riemannian Gaussian distribution
 n_matrices = 100  # how many SPD matrices to generate
-n_dim = 4  # number of dimensions of the SPD matrices
+n_dim = 2  # number of dimensions of the SPD matrices
 sigma = 1.0  # dispersion of the Gaussian distribution
 epsilon = 4.0  # parameter for controlling the distance between centers
 random_state = 42  # ensure reproducibility
@@ -34,15 +34,18 @@ mean = generate_random_spd_matrix(n_dim)  # random reference point
 samples_1 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=mean,
                                 sigma=sigma,
-                                random_state=random_state)
+                                random_state=random_state,
+                                sampling_method='rejection')
 samples_2 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=mean,
                                 sigma=sigma/2,
-                                random_state=random_state)
+                                random_state=random_state,
+                                sampling_method='rejection')
 samples_3 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=epsilon*mean,
                                 sigma=sigma,
-                                random_state=random_state)
+                                random_state=random_state,
+                                sampling_method='rejection')
 
 # Stack all of the samples into one data array for the embedding
 samples = np.concatenate([samples_1, samples_2, samples_3])
--- a/pyriemann/datasets/sampling.py
+++ b/pyriemann/datasets/sampling.py
@@ -103,7 +103,7 @@ def _rejection_sampling_2D(n_samples, sigma, random_state=None):
     Parameters
     ----------
     n_samples : int
-        Number of samples to get from the ptarget distribution.
+        Number of samples to get from the target distribution.
     sigma : float
         Dispersion of the Riemannian Gaussian distribution.
     random_state : int, RandomState instance or None, default=None
--- a/pyriemann/datasets/sampling.py
+++ b/pyriemann/datasets/sampling.py
@@ -61,7 +61,6 @@ def rejection_sampling_a(sigma, r_sample):
     COV_MATRIX = (sigma**2)*np.eye(2)
     M = np.pi*(sigma**2)*np.exp(sigma**2/4)
     if r_sample[0] >= r_sample[1]:
-        M = np.pi*(sigma**2)*np.exp(sigma**2/4)
         num = np.exp(-1/(2*sigma**2) * np.sum(r_sample**2))\
             * np.sinh((r_sample[0] - r_sample[1])/2) * 1/M
         den = multivariate_normal.pdf(r_sample, mean=MU_A, cov=COV_MATRIX)
@@ -87,7 +86,6 @@ def rejection_sampling_b(sigma, r_sample):
     COV_MATRIX = (sigma**2)*np.eye(2)
     M = np.pi*(sigma**2)*np.exp(sigma**2/4)
     if r_sample[0] < r_sample[1]:
-        M = np.pi*(sigma**2)*np.exp(sigma**2/4)
         num = np.exp(-1/(2*sigma**2) * np.sum(r_sample**2))\
             * np.sinh((r_sample[1] - r_sample[0])/2)
         den = multivariate_normal.pdf(r_sample, mean=MU_B, cov=COV_MATRIX)*M
@@ -95,7 +93,7 @@ def rejection_sampling_b(sigma, r_sample):
     return 0
 
 
-def rejection_sampling_v1(n_samples, sigma):
+def rejection_sampling(n_samples, sigma, random_state=None):
     """ rejection sampling algorithm optimized for spatial
     complexity but not for time complexity works very well
     for low sigma values
@@ -117,128 +115,21 @@ def rejection_sampling_v1(n_samples, sigma):
     COV_MATRIX = (sigma**2)*np.eye(2)
     RES = []
     cpt = 0
-    while cpt != n_samples:
-        if np.random.binomial(1, 0.5, 1) == 1:
-            r_sample = multivariate_normal.rvs(MU_A, COV_MATRIX, 1)
-            res = rejection_sampling_a(sigma, r_sample)
-            if np.random.rand(1) < res:
-                RES.append(r_sample)
-                cpt += 1
-        else:
-            r_sample = multivariate_normal.rvs(MU_B, COV_MATRIX, 1)
-            res = rejection_sampling_b(sigma, r_sample)
-            if np.random.rand(1) < res:
-                RES.append(r_sample)
-                cpt += 1
-    return np.array(RES)
-
-
-def rejection_acceptation_probability(t):
-    """polynomial function which approaches the probability
-    of acceptation of the rejection_sampling algorithm depending on sigma
-    Parameters
-    ----------
-    t : float
-    Returns
-    -------
-    rejection_acceptation_probability(t) : float
-    """
-    return 0.01265*t**3 - 0.1648*t**2 + 0.7145*t-0.03374
-
-
-def rejection_acceptation_probability_shifted(t):
-    """polynomial function which approach the probability
-    of acceptation of the rejection_sampling algorithm depending on sigma,
-    the shift is used to estimate with confidence the number
-    of samples we need to get n_samples
-    Parameters
-    ----------
-    t : float
-    Returns
-    -------
-    rejection_acceptation_probability_shifted(t) : float
-    """
-    if t <= 0.3:
-        return 0.02
-    return 0.01265*t**3 - 0.1648*t**2 + 0.7145*t-0.15
-
-
-def rejection_sampling_v2(n_samples, sigma, random_state=None):
-    """ rejection sampling algorithm
-    optimized for time complexity but not for spatial complexity,
-    works very well for any sigma values which is not too low (<0.01)
-    Parameters
-    ----------
-    n_samples : int
-        Number of samples to get from the ptarget distribution.
-    sigma : float
-        Dispersion of the Riemannian Gaussian distribution.
-    random_state : int, RandomState instance or None, default=None
-        Pass an int for reproducible output across multiple function calls.
-    Returns
-    -------
-    r_samples : ndarray, shape (n_samples, n_dim)
-        Samples of the r parameters of the Riemannian Gaussian distribution.
-    """
-    N = int(n_samples/rejection_acceptation_probability_shifted(sigma) + 1000)
-    MU_A = np.array([sigma**2/2, -sigma**2/2])
-    MU_B = np.array([-sigma**2/2, sigma**2/2])
-    COV_MATRIX = (sigma**2)*np.eye(2)
-    RES = []
-    cpt = 0
     rs = check_random_state(random_state)
-    B = rs.binomial(1, 0.5, N)
-    M1 = multivariate_normal.rvs(MU_A, COV_MATRIX, N,
-                                 random_state=random_state)
-    M2 = multivariate_normal.rvs(MU_B, COV_MATRIX, N,
-                                 random_state=random_state)
-    R = rs.rand(N)
-    icount_br = 0
-    kcount_m1 = 0
-    lcount_m2 = 0
     while cpt != n_samples:
-        if B[icount_br] == 1:
-            r_sample = M1[kcount_m1]
-            kcount_m1 += 1
+        if rs.binomial(1, 0.5, 1) == 1:
+            r_sample = multivariate_normal.rvs(MU_A, COV_MATRIX, 1, rs)
             res = rejection_sampling_a(sigma, r_sample)
-            if R[icount_br] < res:
+            if rs.rand(1) < res:
                 RES.append(r_sample)
                 cpt += 1
-            icount_br += 1
-            if (icount_br >= N) and (cpt != n_samples):
-                raise ValueError("sigma value too low")
         else:
-            r_sample = M2[lcount_m2]
-            lcount_m2 += 1
+            r_sample = multivariate_normal.rvs(MU_B, COV_MATRIX, 1, rs)
             res = rejection_sampling_b(sigma, r_sample)
-            if R[icount_br] < res:
+            if rs.rand(1) < res:
                 RES.append(r_sample)
                 cpt += 1
-            icount_br += 1
-            if (icount_br >= N) and (cpt != n_samples):
-                raise ValueError("sigma value too low")
     return np.array(RES)
-
-
-def rejection_sampling(n_samples, sigma, random_state=None):
-    """ final rejection sampling algorithm
-    Parameters
-    ----------
-    n_samples : int
-        Number of samples to get from the ptarget distribution.
-    sigma : float
-        Dispersion of the Riemannian Gaussian distribution.
-    random_state : int, RandomState instance or None, default=None
-        Pass an int for reproducible output across multiple function calls.
-    Returns
-    -------
-    r_samples : ndarray, shape (n_samples, n_dim)
-        Samples of the r parameters of the Riemannian Gaussian distribution.
-    """
-    if sigma < 0.1:
-        return rejection_sampling_v1(n_samples, sigma)
-    return rejection_sampling_v2(n_samples, sigma,
-                                 random_state=random_state)
 
 
 def _slice_one_sample(ptarget, x0, w, rs):
@@ -366,7 +257,9 @@ def _slice_sampling(ptarget, n_samples, x0, n_burnin=20, thin=10,
     return samples
 
 
-def _sample_parameter_r(n_samples, n_dim, sigma, random_state=None, n_jobs=1):
+def _sample_parameter_r(
+        n_samples, n_dim, sigma, random_state=None, n_jobs=1, sampling_method=None
+        ):
     """Sample the r parameters of a Riemannian Gaussian distribution.
 
     Sample the logarithm of the eigenvalues of a SPD matrix following a
@@ -387,15 +280,21 @@ def _sample_parameter_r(n_samples, n_dim, sigma, random_state=None, n_jobs=1):
     n_jobs : int, default=1
         The number of jobs to use for the computation. This works by computing
         each of the class centroid in parallel. If -1 all CPUs are used.
+    sampling_method : str, default=None
+        Name of the sampling method used to sample samples_r. It can be
+        'slice' or 'rejection'.
 
     Returns
     -------
     r_samples : ndarray, shape (n_samples, n_dim)
         Samples of the r parameters of the Riemannian Gaussian distribution.
     """
-    if n_dim == 2:
+    if n_dim == 2 and sampling_method != "slice":
         return rejection_sampling(n_samples, sigma,
                                   random_state=random_state)
+    if n_dim != 2 and sampling_method == "rejection":
+        raise ValueError(
+            f'n_dim={n_dim} is not yet supported with rejection sampling')
     rs = check_random_state(random_state)
     x0 = rs.randn(n_dim)
     ptarget = partial(_pdf_r, sigma=sigma)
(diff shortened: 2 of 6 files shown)


examples/simulated/plot_riemannian_gaussian.py
```diff
@@ -34,15 +34,18 @@
 samples_1 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=mean,
                                 sigma=sigma,
-                                random_state=random_state)
+                                random_state=random_state,
+                                sampling_method='rejection')
```
agramfort, line 38: I would introduce a value called "auto" for the sampling_method method that would default to "rejection" if dim == 2 and slice otherwise so you don't have to expose these options in the tutorial and it will just become faster for users without any code change

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,96 @@ def _pdf_r(r, sigma):
...
+    the function g+.
+    Parameters
+    ----------
+    sigma : float
+        Dispersion of the Riemannian Gaussian distribution.
+    r_samples : ndarray, shape (1, n_dim)
+        Sample of the r parameters of the Riemannian Gaussian distribution.
+    Returns
+    -------
+    probability_of_acceptation : float
```
agramfort, line 58:
```suggestion
    """Auxiliary function for the 2D rejection sampling algorithm.
    
    It is used in the case where r is sampled with the function g+.

    Parameters
    ----------
    sigma : float
        Dispersion of the Riemannian Gaussian distribution.
    r_samples : ndarray, shape (1, n_dim)
        Sample of the r parameters of the Riemannian Gaussian distribution.

    Returns
    -------
    probability_of_acceptation : float
```

see pep257

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,96 @@ def _pdf_r(r, sigma):
...
+    """
+    MU_A = np.array([sigma**2/2, -sigma**2/2])
```
agramfort, line 60: such capital letters are used for constants not for regular variables
```suggestion
    mu_a = np.array([sigma**2/2, -sigma**2/2])
```

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,96 @@ def _pdf_r(r, sigma):
...
+    the function g-.
+    Parameters
+    ----------
+    sigma : float
+        Dispersion of the Riemannian Gaussian distribution.
+    r_samples : ndarray, shape (n_samples, n_dim)
+        Samples of the r parameters of the Riemannian Gaussian distribution.
+    Returns
+    -------
+    probability_of_acceptation : float
```
agramfort, line 83: to reformat for pep257

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,96 @@ def _pdf_r(r, sigma):
...
+    n_samples : int
+        Number of samples to get from the ptarget distribution.
+    sigma : float
+        Dispersion of the Riemannian Gaussian distribution.
+    random_state : int, RandomState instance or None, default=None
+        Pass an int for reproducible output across multiple function calls.
+    Returns
+    -------
+    r_samples : ndarray, shape (n_samples, n_dim)
+        Samples of the r parameters of the Riemannian Gaussian distribution.
```
agramfort, line 112: to reformat

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,96 @@ def _pdf_r(r, sigma):
...
+
+def _rejection_sampling_2D(n_samples, sigma, random_state=None):
+    """ Rejection sampling algorithm for the 2D case. The algorithm is
+    optimized for spatial complexity but not for time complexity.
+    Works very well for low sigma values
+
+    Parameters
+    ----------
+    n_samples : int
+        Number of samples to get from the ptarget distribution.
```
agramfort, line 104:
```suggestion
        Number of samples to get from the target distribution.
```

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,96 @@ def _pdf_r(r, sigma):
...
+        Pass an int for reproducible output across multiple function calls.
+    Returns
+    -------
+    r_samples : ndarray, shape (n_samples, n_dim)
+        Samples of the r parameters of the Riemannian Gaussian distribution.
+    """
+    MU_A = np.array([sigma**2/2, -sigma**2/2])
+    MU_B = np.array([-sigma**2/2, sigma**2/2])
+    COV_MATRIX = (sigma**2)*np.eye(2)
+    RES = []
```
agramfort, line 117: remove the capital letters in variables

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,96 @@ def _pdf_r(r, sigma):
...
+    cpt = 0
+    rs = check_random_state(random_state)
+    while cpt != n_samples:
```
agramfort, line 120: use a for loop not while here

pyriemann/datasets/sampling.py
```diff
@@ -188,13 +280,26 @@ def _sample_parameter_r(n_samples, n_dim, sigma, random_state=None, n_jobs=1):
     n_jobs : int, default=1
         The number of jobs to use for the computation. This works by computing
         each of the class centroid in parallel. If -1 all CPUs are used.
+    sampling_method : str, default=None
```
agramfort, line 283: make it default to "auto" as explained above

pyriemann/datasets/sampling.py
```diff
@@ -188,13 +280,26 @@ def _sample_parameter_r(n_samples, n_dim, sigma, random_state=None, n_jobs=1):
...
 
     Returns
     -------
     r_samples : ndarray, shape (n_samples, n_dim)
         Samples of the r parameters of the Riemannian Gaussian distribution.
     """
-
+    if sampling_method not in ['slice', 'rejection', None]:
+        raise ValueError(f'Unknown sampling method {sampling_method},\
+                         try slice or rejection')
```
agramfort, line 296:
```suggestion
        raise ValueError(f'Unknown sampling method {sampling_method},'
                         'try slice or rejection')
```

pyriemann/datasets/sampling.py
plcrodrigues: I would also prefer to use a for loop, but this would mean that we know the number of steps that we need to get `n_samples`... and this is not really the case, since we're rejecting candidates randomly. Of course, we could estimate a "reasonable" number of passes through the for loop like making it n_samples / prob_acceptance, but the prob_acceptance is hard to calculate and it is just an estimate of the number of passes we will go through anyways. Note that in the `sbi` package they do rejection_sampling with a while loop too; see https://github.com/mackelab/sbi/blob/main/sbi/samplers/rejection/rejection.py

qbarthelemy (changes requested):
Thx for this PR with such a detailed description!
Can you update `test_sampling` too?

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,100 @@ def _pdf_r(r, sigma):
     return np.exp(partial_1 + partial_2)
 
 
+def _rejection_sampling_2D_gfunction_plus(sigma, r_sample):
+    """Auxiliary function for the 2D rejection sampling algorithm.
+    It is used in the case where r is sampled with the function g+.
```
qbarthelemy, line 48:
```suggestion
    """Auxiliary function for the 2D rejection sampling algorithm.

    It is used in the case where r is sampled with the function g+.
```

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,100 @@ def _pdf_r(r, sigma):
...
+        num = np.exp(-1/(2*sigma**2) * np.sum(r_sample**2))\
+            * np.sinh((r_sample[0] - r_sample[1])/2) * 1/m
+        den = multivariate_normal.pdf(r_sample, mean=mu_a, cov=cov_matrix)
+        return num / den
+    return 0
+
+
+def _rejection_sampling_2D_gfunction_minus(sigma, r_sample):
+    """Auxiliary function for the 2D rejection sampling algorithm.
+    It is used in the case where r is sampled with the function g-.
```
qbarthelemy, line 74:
```suggestion
    """Auxiliary function for the 2D rejection sampling algorithm.

    It is used in the case where r is sampled with the function g-.
```

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,100 @@ def _pdf_r(r, sigma):
...
+        num = np.exp(-1/(2*sigma**2) * np.sum(r_sample**2))\
+            * np.sinh((r_sample[1] - r_sample[0])/2)
+        den = multivariate_normal.pdf(r_sample, mean=mu_b, cov=cov_matrix)*m
+        return num/den
+    return 0
+
+
+def _rejection_sampling_2D(n_samples, sigma, random_state=None):
+    """Rejection sampling algorithm for the 2D case.
+    Implementation of a rejection sampling algorithm. The implementation
```
qbarthelemy, line 100:
```suggestion
    """Rejection sampling algorithm for the 2D case.

    Implementation of a rejection sampling algorithm. The implementation
```

pyriemann/datasets/sampling.py
```diff
@@ -42,6 +43,103 @@ def _pdf_r(r, sigma):
...
+    Parameters
+    ----------
+    sigma : float
+        Dispersion of the Riemannian Gaussian distribution.
+    r_samples : ndarray, shape (1, n_dim)
+        Sample of the r parameters of the Riemannian Gaussian distribution.
+
+    Returns
+    -------
+    probability_of_acceptation : float
```
qbarthelemy, line 60:
```suggestion
    p : float
        Probability of acceptation.
```

qbarthelemy (approved):

examples/simulated/plot_riemannian_gaussian.py
agramfort, line 38:
```suggestion
                                random_state=random_state)
```

examples/simulated/plot_riemannian_gaussian.py
```diff
@@ -34,15 +34,18 @@
...
 samples_2 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=mean,
                                 sigma=sigma/2,
-                                random_state=random_state)
+                                random_state=random_state,
+                                sampling_method='rejection')
```
agramfort, line 43:
```suggestion
                                random_state=random_state)
```

examples/simulated/plot_riemannian_gaussian.py
```diff
@@ -34,15 +34,18 @@
...
 samples_3 = sample_gaussian_spd(n_matrices=n_matrices,
                                 mean=epsilon*mean,
                                 sigma=sigma,
-                                random_state=random_state)
+                                random_state=random_state,
+                                sampling_method='rejection')
```
agramfort, line 48:
```suggestion
                                random_state=random_state)
```

agramfort (commented):
I would not expose the `sampling_method` in the example. Just keep the default value of the parameter.

besides LGTM

agramfort (approved):

pyriemann/datasets/simulated.py
```diff
@@ -79,7 +79,8 @@ def make_masks(n_masks, n_dim0, n_dim1_min, rs):
 
 def make_gaussian_blobs(n_matrices=100, n_dim=2, class_sep=1.0, class_disp=1.0,
                         return_centers=False, random_state=None, *,
-                        mat_mean=.0, mat_std=1., n_jobs=1):
+                        mat_mean=.0, mat_std=1., n_jobs=1,
+                        sampling_method=None):
```
sylvchev, line 83:
```suggestion
                        sampling_method="auto"):
```

pyriemann/datasets/sampling.py
```diff
@@ -167,7 +267,8 @@ def _slice_sampling(ptarget, n_samples, x0, n_burnin=20, thin=10,
     return samples
 
 
-def _sample_parameter_r(n_samples, n_dim, sigma, random_state=None, n_jobs=1):
+def _sample_parameter_r(n_samples, n_dim, sigma,
+                        random_state=None, n_jobs=1, sampling_method=None):
```
sylvchev, line 271:
```suggestion
                        random_state=None, n_jobs=1, sampling_method="auto"):
```

pyriemann/datasets/sampling.py
```diff
@@ -188,13 +289,26 @@ def _sample_parameter_r(n_samples, n_dim, sigma, random_state=None, n_jobs=1):
...
+        will be equal to 'slice' for n_dim != 2 and equal to
+        'rejection' for n_dim = 2.
 
     Returns
     -------
     r_samples : ndarray, shape (n_samples, n_dim)
         Samples of the r parameters of the Riemannian Gaussian distribution.
     """
-
+    if sampling_method not in ['slice', 'rejection', None]:
```
sylvchev, line 303:
```suggestion
    if sampling_method not in ['slice', 'rejection', 'auto']:
```

pyriemann/datasets/sampling.py
```diff
@@ -242,9 +356,8 @@ def _sample_parameter_U(n_samples, n_dim, random_state=None):
     return u_samples
 
 
-def _sample_gaussian_spd_centered(
-        n_matrices, n_dim, sigma, random_state=None, n_jobs=1
-        ):
+def _sample_gaussian_spd_centered(n_matrices, n_dim, sigma, random_state=None,
+                                  n_jobs=1, sampling_method=None):
```
sylvchev, line 360:
```suggestion
                                  n_jobs=1, sampling_method="auto"):
```

pyriemann/datasets/sampling.py
```diff
@@ -264,6 +377,11 @@ def _sample_gaussian_spd_centered(
     n_jobs : int, default=1
         The number of jobs to use for the computation. This works by computing
         each of the class centroid in parallel. If -1 all CPUs are used.
+    sampling_method : str, default=None
+        Name of the sampling method used to sample samples_r. It can be
+        'slice' or 'rejection'. If it is None, the sampling_method
```
sylvchev, line 382:
```suggestion
        'auto', 'slice' or 'rejection'. If it is 'auto', the sampling_method
```

pyriemann/datasets/sampling.py
```diff
@@ -301,7 +420,8 @@ def _sample_gaussian_spd_centered(
     return samples
 
 
-def sample_gaussian_spd(n_matrices, mean, sigma, random_state=None, n_jobs=1):
+def sample_gaussian_spd(n_matrices, mean, sigma, random_state=None,
+                        n_jobs=1, sampling_method=None):
```
sylvchev, line 424:
```suggestion
                        n_jobs=1, sampling_method='auto'):
```

pyriemann/datasets/sampling.py
```diff
@@ -324,6 +444,11 @@ def sample_gaussian_spd(n_matrices, mean, sigma, random_state=None, n_jobs=1):
     n_jobs : int, default=1
         The number of jobs to use for the computation. This works by computing
         each of the class centroid in parallel. If -1 all CPUs are used.
+    sampling_method : str, default=None
+        Name of the sampling method used to sample samples_r. It can be
+        'slice' or 'rejection'. If it is None, the sampling_method
```
sylvchev, line 449:
```suggestion
        'auto', 'slice' or 'rejection'. If it is 'auto', the sampling_method
```

pyriemann/datasets/sampling.py
```diff
@@ -264,6 +377,11 @@ def _sample_gaussian_spd_centered(
     n_jobs : int, default=1
         The number of jobs to use for the computation. This works by computing
         each of the class centroid in parallel. If -1 all CPUs are used.
+    sampling_method : str, default=None
```
sylvchev, line 380:
```suggestion
    sampling_method : str, default="auto"
```

pyriemann/datasets/sampling.py
```diff
@@ -324,6 +444,11 @@ def sample_gaussian_spd(n_matrices, mean, sigma, random_state=None, n_jobs=1):
...
```
sylvchev, line 447:
```suggestion
    sampling_method : str, default="auto"
```

tests/test_sampling.py
```diff
@@ -8,20 +8,41 @@
 
 
 @pytest.mark.parametrize("n_jobs", [1, -1])
-def test_sample_gaussian_spd(n_jobs):
-    """Test Riemannian Gaussian sampling."""
-    n_matrices, n_dim, sigma = 10, 8, 1.
+@pytest.mark.parametrize("sampling_method", [None, 'slice', 'rejection'])
```
sylvchev, line 11:
```suggestion
@pytest.mark.parametrize("sampling_method", ['auto', 'slice', 'rejection'])
```

pyriemann/datasets/sampling.py
```diff
@@ -264,6 +385,12 @@ def _sample_gaussian_spd_centered(
     n_jobs : int, default=1
         The number of jobs to use for the computation. This works by computing
         each of the class centroid in parallel. If -1 all CPUs are used.
+    sampling_method : str, default='auto'
+        Name of the sampling method used to sample samples_r. It can be
+        'auto', 'slice' or 'rejection'. If it is 'auto', the sampling_method
+        will be equal to 'slice' for n_dim != 2 and equal to
+        'rejection' for n_dim = 2.
+        .. versionadded:: 0.3.1
```
qbarthelemy, line 393:
```suggestion

        .. versionadded:: 0.3.1
```

pyriemann/datasets/sampling.py
```diff
@@ -188,13 +289,34 @@ def _sample_parameter_r(n_samples, n_dim, sigma, random_state=None, n_jobs=1):
...
```
qbarthelemy, line 297:
```suggestion

        .. versionadded:: 0.3.1
```

plcrodrigues (approved):

pyriemann/datasets/sampling.py
```diff
@@ -324,6 +475,13 @@ def sample_gaussian_spd(n_matrices, mean, sigma, random_state=None, n_jobs=1):
     n_jobs : int, default=1
         The number of jobs to use for the computation. This works by computing
         each of the class centroid in parallel. If -1 all CPUs are used.
+    sampling_method : str, default='auto'
+        Name of the sampling method used to sample samples_r. It can be
```
qbarthelemy, line 479: Not sure that `samples_r` is understandable by any user reading the doc.
```suggestion
        Sampling method to sample eigenvalues. It can be
```

sylvchev:
Thanks @Artim436 and all for this PR!

PR #199, opened 2022-08-26
MAINT correct license

sylvchev:
It seems that the license is not correctly recognized, it is listed as BSD-1 clause in https://github.com/ml-tooling/best-of-ml-python#medical-data  and the wording is not exactly the same as the [reference one](https://tldrlegal.com/license/bsd-3-clause-license-(revised)#fulltext).

I used MNE license file template to update ours. If the issue persists in best-of-ml listing, I'll open an issue on their side.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright © 2015, authors of pyRiemann
+Copyright (c) 2015-2022, authors of pyRiemann
 All rights reserved.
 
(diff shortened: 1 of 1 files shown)


qbarthelemy:
Thx @sylvchev !

qbarthelemy:
It's OK now: https://github.com/ml-tooling/best-of-ml-python#medical-data

PR #209, opened 2022-11-22
hide KMeans from import in clustering module

agramfort:
this avoids my mistake

`from sklearn.clustering import KMeans`

which as the KMeans from sklearn...


qbarthelemy:
Thx Alex!

PR #211, opened 2022-11-25
Use a unique function check_weights 

qbarthelemy:
Many modules check weights: `utils.mean`, `utils.median`, `utils.ajd`, `transfer._rotate`, 
but each one has its own function.

This PR uses a single function and completes tests.


agramfort (approved):

agramfort:
Thx @qbarthelemy

PR #214, opened 2022-12-01
Improve doc, code and tests of module spatial filters

qbarthelemy:
Tests for spatial filters were written only for the special case of n_channels < n_filters.

This PR:
- improves code (remove np.dot and useless reshapes),
- completes documentation of arrays shape, for attributes or outputs,
- completes tests with all possible cases (n_channels, n_filters, baseline_cov),
and tests shape of attributes.


agramfort (approved):
thx @qbarthelemy

PR #216, opened 2022-12-06
Improve doc, code and tests of module estimation

qbarthelemy:
Tests for estimation were written only for the special case of n_filters and svd < n_channels, and attribute P_ was not tested.

This PR:

- improves code (remove np.dot),
- completes documentation of arrays shape (for attributes or outputs) and adds references for Hankel matrices,
- completes tests with all possible cases (delays, n_channels, n_filters, applyfilters), and tests shape of attributes.


agramfort (approved):
🙏

Issue #58, opened 2017-06-11
Kernel Matrix Representation


alexandrebarachant:
From this [paper](http://www.cv-foundation.org/openaccess/content_iccv_2015/papers/Wang_Beyond_Covariance_Feature_ICCV_2015_paper.pdf) 
> Wang, Lei, et al. "Beyond covariance: Feature representation with nonlinear kernel matrices." Proceedings of the IEEE International Conference on Computer Vision. 2015.

we can replace covariance by a kernel matrix. In our case, we could apply a kernel function between any pair of channels to build the kernel matrix.

This could be an interesting functionality to implement. We could rely on scikit-learn [kernel functions](http://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.kernel_metrics.html#sklearn.metrics.pairwise.kernel_metrics)

This could be done as a stand-alone Transformer (i.e. `KernelMatrixRepresentation`) or by extending the parameter `estimator` of the covariance estimation.

The later solution is easier, just extending the dict in `_check_est` from `utils.covariance` with what the `sklearn.metrics.pairwise.kernel_metrics` returns. However, we must add a mechanism to pass parameters to the estimator (which is not possible currently)

This is an easy one. Any Help is welcome

plcrodrigues:
I personally prefer the second option, since passing parameters to the covariance estimator could be useful for doing other stuff in the future. What you think about including a dict `params` to the `__init__` function of the Covariances estimator? I guess it won't break the code, right?

alexandrebarachant:
Yes, we should add a param, like `estimator_param` which would be a dict that can be passed to the estimator.
This would be usefull for existing estimator as well. 
and it would be better than what i currently do (pasing a `lambda x: est(x, param)`  to the estimator)

plcrodrigues:
Cool. I would like to work on this! I've just read the paper and the idea seems indeed interesting.

alexandrebarachant:
It would be great if you can do it !

I already tried very quickly with the moab (by passing a function to the estimator), depending on the kernel it actually improve performances marginally.

This is an easy PR. It would be great to also add an example with performances of various estimation methods.

PR #217, opened 2022-12-21
Add kernel matrices as new features

qbarthelemy:
Solves https://github.com/pyRiemann/pyRiemann/issues/58

@plcrodrigues @gabelstein 

Accuracy of Laplacian kernels look promissing on this small example.
@sylvchev , can you test on MOABB if this improvement is statistically significant on large datasets?
Plot_cov_kernels

### DIFF
--- a/pyriemann/estimation.py
+++ b/pyriemann/estimation.py
@@ -1,7 +1,8 @@
-"""Estimation of covariance matrices."""
+"""Estimation of SPD matrices."""
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.covariance import shrunk_covariance
+from sklearn.metrics.pairwise import pairwise_kernels
 
 from .spatialfilters import Xdawn
 from .utils.covariance import (covariances, covariances_EP, cospectrum,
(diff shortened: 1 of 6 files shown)


pyriemann/estimation.py
```diff
@@ -704,6 +705,107 @@ def transform(self, X):
...
+        <https://www.cv-foundation.org/openaccess/content_iccv_2015/papers/Wang_Beyond_Covariance_Feature_ICCV_2015_paper.pdf>`_
+        L. Wang, J. Zhang, L. Zhou, C. Tang, W Li. ICCV, 2015.
+    .. [2]
+        https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.pairwise_kernels.html
+    """  # noqa
+
+    def __init__(self, metric='linear', n_jobs=None):
+        """Init."""
+        self.metric = metric
+        self.n_jobs = n_jobs
```
agramfort, line 759: some kernels have parameters eg the gamma in RBF. You should expose this option from the **kwds in the sklearn.metrics.pairwise.pairwise_kernels function

qbarthelemy: yep, good idea!

agramfort (approved):

sylvchev (approved):

sylvchev:
Here are some results for larger datasets, on left/right hand motor imagery, with 5-fold CV on 6 datasets, averaging score accuracy for all subjects. I compared a classifier relying on covariance (baseline) with classifiers relying on kernels with different metrics (linear, polynomial, rbf, laplacian, cosine).

For MDM with covariance and MDM with various kernels
mdm-kernels

Same for tangent space projection + logistic regression
tslr-kernels

Finally for tangent space + SVM
tssvm-kernels

The improvement is visible for MDM with linear kernel, but no real improvement for tangent space with + classifier.

sylvchev:
Results for MDM are nice and call for more research work!

LGTM!

qbarthelemy:
Thx @sylvchev for this great validation!

For motor imagery classification, there is no significant differences with respect covariance matrices (equivalent to linear kernel, as explained in Wang2015).

Maybe, other kernels and/or other tasks will show accuracy improvement thanks to non-linear dependencies between temporal samples.

PR #220, opened 2023-01-17
Add fixed point covariance estimator and add **kwds arguments in Covariances 

qbarthelemy:
This PR:
- adds a new covariance estimator: [fixed point covariance matrix](https://hal.science/hal-02495012/document);
- adds **kwds arguments in `Covariances`, for parameters passed directly to the covariance estimator.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -35,7 +35,7 @@ v0.3.1.dev
 
 - Add kernel matrices representation :class:`pyriemann.estimation.Kernels` and complete example comparing estimators. :pr:`217` by :user:`qbarthelemy`
 
-- Add a new covariance estimator, robust fixed point covariance, and add **kwds arguments for all covariance based functions and classes. :pr:`220` by :user:`qbarthelemy`
+- Add a new covariance estimator, robust fixed point covariance, and add kwds arguments for all covariance based functions and classes. :pr:`220` by :user:`qbarthelemy`
 
 v0.3 (July 2022)
 ----------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -35,6 +35,8 @@ v0.3.1.dev
 
 - Add kernel matrices representation :class:`pyriemann.estimation.Kernels` and complete example comparing estimators. :pr:`217` by :user:`qbarthelemy`
 
+- Add a new covariance estimator, fixed point covariance, and add **kwds arguments for all covariance based functions and classes. :pr:`220` by :user:`qbarthelemy`
+
 v0.3 (July 2022)
 ----------------
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -35,7 +35,7 @@ v0.3.1.dev
 
 - Add kernel matrices representation :class:`pyriemann.estimation.Kernels` and complete example comparing estimators. :pr:`217` by :user:`qbarthelemy`
 
-- Add a new covariance estimator, fixed point covariance, and add **kwds arguments for all covariance based functions and classes. :pr:`220` by :user:`qbarthelemy`
+- Add a new covariance estimator, robust fixed point covariance, and add **kwds arguments for all covariance based functions and classes. :pr:`220` by :user:`qbarthelemy`
 
 v0.3 (July 2022)
 ----------------
--- a/pyriemann/utils/covariance.py
+++ b/pyriemann/utils/covariance.py
@@ -52,7 +52,7 @@ def _fpcm(X, *, init=None, tol=10e-3, n_iter_max=50, assume_centered=False):
         <https://hal.science/hal-01816367/document>`_
         F. Pascal, Y. Chitour, J.P. Ovarlez, P. Forster, P. Arzabal. IEEE
         Transactions on Signal Processing, 2008.
-    """  #noqa
+    """  # noqa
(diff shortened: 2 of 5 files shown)


qbarthelemy:
Yep, but `**kwds` arguments must also be added
in functions: `covariances_EP`, `covariances_X`, `block_covariances`;
and in classes: `ERPCovariances`, `XdawnCovariances`, `BlockCovariances`, `HankelCovariances`.

pyriemann/utils/covariance.py
```diff
@@ -60,14 +109,6 @@ def _sch(X):
...
     -----
     .. versionadded:: 0.3
-
-    References
-    ----------
-    .. [1] `A shrinkage approach to large-scale covariance estimation and
-        implications for functional genomics
-        <https://pubmed.ncbi.nlm.nih.gov/16646851/>`_
-        J. Schafer, and K. Strimmer. Statistical Applications in Genetics and
-        Molecular Biology, Volume 4, Issue 1, November 2005.
```
agramfort, line 70: I would have kept the ref here. Otherwise it's maybe less clear where to find the info.

pyriemann/utils/covariance.py
```diff
@@ -6,36 +6,85 @@
...
+        The maximum number of iterations.
+    assume_centered : bool, default=False
+        If `True`, data will not be centered before computation.
+        Useful when working with data whose mean is almost, but not exactly
+        zero. If `False`, data will be centered before computation.
+
+    Returns
+    -------
+    cov : ndarray, shape (n_channels, n_channels)
+        Fixed point covariance matrix.
```
agramfort, line 32: can you add the ref "Theoretical analysis of an improved covariance matrix estimator in non-Gaussian noise" here?

Do I understand correctly that this does some kind of robust estimation with M-estimators?

pyriemann/utils/covariance.py
```diff
@@ -117,27 +159,30 @@ def _check_est(est):
...
-        * 'lwf' for shrunk Ledoit-Wolf covariance matrix [4]_,
-        * 'mcd' for minimum covariance determinant matrix [5]_,
-        * 'oas' for oracle approximating shrunk covariance matrix [6]_,
-        * 'sch' for Schaefer-Strimmer covariance matrix [7]_,
-        * 'scm' for sample covariance matrix [8]_,
+        Covariance matrix estimator [est]_:
+
+        * 'corr' for correlation coefficient matrix [corr]_,
+        * 'cov' for numpy based covariance matrix [cov]_,
+        * 'fpcm' for fixed point covariance matrix [fpcm]_,
```
agramfort, line 175: I find the name not super informative and reading the PR it's not super clear what it is useful for and why I should use it? Can you add a blob of text somewhere to explain this?

🙏

pyriemann/utils/covariance.py
qbarthelemy: Ok, but since these functions are private, there are not displayed in API, and thus users can't see the references.
We can duplicate references. What do you think?

agramfort: yes duplicate them it does not harm.

pyriemann/utils/covariance.py
qbarthelemy: Done.

It is not a robust M-estimator. M is just the name of the covariance matrix.

agramfort:
What am I missing here @qbarthelemy ? See the estimate is very different from Id matrix.

image

qbarthelemy:
Good catch @agramfort !

I had naively translated code from Matlab covariance toolbox https://github.com/alexandrebarachant/covariancetoolbox/blob/master/lib/estimation/FPCov.m
where trace-normalization is missing (Eq(8) in Pascal2005).

It is corrected and I have completed tests.

pyriemann/utils/covariance.py
qbarthelemy: You were right: fixed point covariance estimates the distribution-free robust M-estimator, see [Tyler1987](https://projecteuclid.org/journals/annals-of-statistics/volume-15/issue-1/A-Distribution-Free-M-Estimator-of-Multivariate-Scatter/10.1214/aos/1176350263.full). 
I have updated doc.

agramfort (approved):
victory !

Screenshot 2023-01-27 at 23 15 43

Feel free to just add a line in what's new saying that it's a robust estimator.

+1 for MRG

thx @qbarthelemy

doc/whatsnew.rst
```diff
@@ -35,6 +35,8 @@ v0.3.1.dev
 
 - Add kernel matrices representation :class:`pyriemann.estimation.Kernels` and complete example comparing estimators. :pr:`217` by :user:`qbarthelemy`
 
+- Add a new covariance estimator, robust fixed point covariance, and add **kwds arguments for all covariance based functions and classes. :pr:`220` by :user:`qbarthelemy`
```
sylvchev, line 38:
```suggestion
- Add a new covariance estimator, robust fixed point covariance, and add kwds arguments for all covariance based functions and classes. :pr:`220` by :user:`qbarthelemy`
```
Seems this is generating a warning

sylvchev (approved):

sylvchev:
Thanks @qbarthelemy for the PR and thank you @agramfort for spotting this!

Issue #204, opened 2022-10-15
Allow processing of HPD matrices


qbarthelemy:
## Description

A [Hermitian positive definite (HPD) matrix](https://en.wikipedia.org/wiki/Definite_matrix#Definitions_for_complex_matrices) $M$ is defined as:
$M = A + iB$
where $A$ is a symmetric positive definite (SPD) matrix and $B$ is a skew-symmetric matrix.

pyRiemann should be adapted to process such complex-valued data.

## [1](#1) - Tests

To validate modifications and to complete test, we must also add a new function to generate generic, SPD or HPD matrices.
- [x] `make_matrices(n_matrices, n_channels, kind='spd' | 'hpd', same_eigvecs=True)` will be able to generate SPD or HPD matrices;
- [x] `get_covmats` will be replaced by `get_matrices`.

Then, tests will be:
```
@pytest.mark.parametrize("kind", ['spd', 'hpd'])
def test_distance_harmonic(get_matrices, kind):
    n_matrices, n_channels = 6, 5
    mats = get_matrices(n_matrices, n_channels, kind=kind)
```

## [2](#2) - Utils

The first obvious change is to update matrix operators,
`pyriemann.utils.base._matrix_operator`, adding `.conj()`:
`D = (eigvecs * eigvals) @ np.swapaxes(eigvecs.conj(), -2, -1)`

Other functions which must be adapted too:
- [x] base operators,
- [x] distances,
- [x] geodesics,
- [x] tangent space,
- [x] means,
- [x] medians,
- [x] ajd,
- [ ] kernels (?).

## [3](#3) - Covariance matrices

For [complex inputs](https://en.wikipedia.org/wiki/Covariance_matrix#Complex_random_vectors), like in [radar image processing](https://ammarmian.github.io/pdf/wiley_book_2021.pdf), covariance matrix estimations must use a Hermitian self-multiplication.

For the sample covariance matrix (scm), pseudo-covariance $C = \frac{1}{n} X X^T$ must be replaced by actual covariance $C = \frac{1}{n} X X^H$.

This is OK for `cov` estimator realying on [np.cov](https://numpy.org/doc/stable/reference/generated/numpy.cov.html).
But, most of covariance estimators rely on [scikit-learn](https://scikit-learn.org/stable/modules/covariance.html), processing only real inputs: 
- [x] `scm`,
- [x] `lwf`,
- [x] `mcd`,
- [x] `oas`,
[...]

PR #223, opened 2023-02-03
Add robust M-estimators for covariance matrix

qbarthelemy:
This PR aims to:
- generalize fixed-point covariance estimator (recently added in #220) into a new covariance estimator, supporting three robust M-estimators: Huber, Student-t and Tyler;
(related to https://github.com/scikit-learn/scikit-learn/pull/2081)
- add to public API covariance estimators not imported from scikit-learn;
- highlight covariance estimators supporting complex inputs;
- complete tests;
- add an example on robust covariance estimation for corrupted inputs;
- move example on NaN-mean from "signal" to "simulated" folder.

@AmmarMian, your review is welcome!

Result of example:
plot_covariance_estimation_robust_1


pyriemann/utils/covariance.py
```diff
@@ -28,90 +83,98 @@ def _fpcm(X, *, init=None, tol=10e-3, n_iter_max=50, assume_centered=False):
...
         X -= np.mean(X, axis=1, keepdims=True)
     if init is None:
-        cov = n_channels * _scm(X, assume_centered=True)
+        cov = _scm(X, assume_centered=True)
     else:
         cov = init
+    X_H = X.conj().T
 
     for _ in range(n_iter_max):
-        diag_ = np.diag(X.T @ np.linalg.inv(cov) @ X)
```
AmmarMian, line 65: This line is not time and memory efficient: As it has to compute a matrix of size (n_times,n_times) to then takes its diag part, it doesn't scale well.

Attached here are two implementations, one I used in robuststats and another one from @flbouchard where he uses a Cholesky decomposition and Lapack inversion. This an important issue because we do this computation multiple times through the fixed-point iteration.

```python
import numpy as np
import numpy.linalg as la
from scipy.linalg.lapack import dtrtri
from scipy.stats import multivariate_normal
from scipy.linalg import toeplitz
from sklearn.covariance import empirical_covariance
from pyriemann.utils.base import invsqrtm
from numpy.testing import assert_almost_equal


n_channels = 500
n_samples = 10000
cov = toeplitz(np.power(0.7, np.arange(0, n_channels)))
X = multivariate_normal.rvs(cov=cov, size=n_samples).T
init = empirical_covariance(X.T)

def weight_func(x):  # Example 2, Section V-C in [1]
    return n_channels / np.where(x < 1e-10, 1e-10, x)

def iteration_diag(X, init):
    X_H = X.conj().T
    powers = np.diag(X_H @ np.linalg.inv(init) @ X)
    cov_new = np.einsum('ab,bc->ac', X * weight_func(powers), X_H)
    return cov_new


def iteration_invsqrtm(X, init):
    temp = invsqrtm(init)@X
    quadratic = np.einsum('ij,ji->i', temp.T, temp)
    quadratic = weight_func(quadratic)
    temp = X * np.sqrt(quadratic)
    return temp@temp.T

def iteration_lapack(X, init):
    L = la.cholesky(init)
    iL, info = dtrtri(L, lower=1)
    v = iL @ X
    a = np.einsum('ij,ji->i', v.T, v)
    Y = np.sqrt(weight_func(a)) * X
    return Y @ Y.T

assert_almost_equal(iteration_diag(X, init), iteration_invsqrtm(X, init))
assert_almost_equal(iteration_diag(X, init), iteration_lapack(X, init))
```
image

AmmarMian (changes requested):
Needs more efficient implementation.

pyriemann/utils/covariance.py
qbarthelemy: Thx @AmmarMian for this valuable suggestion!

qbarthelemy: On complex arrays, `dtrtri` raises this warning:
`ComplexWarning: Casting complex values to real discards the imaginary part`

AmmarMian: Ah yes probably only work for real-valued data. I'll circle back with Florent who knows well this library if there is a possibility for complex-valued data.

For now maybe the invsqrtm implementation is the best one for all kind of data since it doesn't requires the diag.

AmmarMian: Actually, we can just use the ctrtri in the complex case:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lapack.ctrtri.html

doc/whatsnew.rst
```diff
@@ -39,6 +39,9 @@ v0.3.1.dev
 
 - Add example in gallery on frequency band selection using class distinctiveness measure. :pr:`219` by :user:`MSYamamoto`
 
+- Add :func:`pyriemann.utils.covariance.covariance_mest` supporting three robust M-estimators (Huber, Student-t and Tyler)
+  and add an example on robust covariance estimation for corrupted data. :pr:`223` by :user:`qbarthelemy`
```
agramfort, line 43: say also that these are available with covariances function and Covariances object

pyriemann/utils/covariance.py
qbarthelemy: `ctrtri` is ok for complex arrays. However, after several iterations, this implementation generates NaN values and breaks many tests.

Consequently, I use `invsqrtm`-based implementation, more stable and faster than naive one.
I put this code in a specific function `distance_mahalanobis`, and I add a test to compare results with SciPy in the case real-valued data.

pyriemann/utils/covariance.py
```diff
@@ -28,90 +83,98 @@ def _fpcm(X, *, init=None, tol=10e-3, n_iter_max=50, assume_centered=False):
         If `True`, data will not be centered before computation.
         Useful when working with data whose mean is almost, but not exactly
         zero. If `False`, data will be centered before computation.
+    q : float, default=0.5
+        Using Huber's M-estimator, percentage of samples deemed uncorrupted, in
+        (0, 1].
+    nu : float, default=5
+        Using Student-t's M-estimator, degree of freedom for t-distribution
+        (strictly positive).
```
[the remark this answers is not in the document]
qbarthelemy: @AmmarMian , do you have a suggestion to complete the explanation? And, is default value OK?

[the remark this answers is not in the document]
AmmarMian: Description is okay. You could add that (1-q) corresponds to the percentage of data treated as outliers w.r.t a Gaussian distribution.

agramfort (approved):

AmmarMian (changes requested):
Line 168: 
`cov *= n_channels / np.trace(cov)`
should only be done when using Tyler's estimator. The other ones do not need normalization since they are consistent. Normalizing means you also lose the information of the signal power level. Which is useful in certain applications.

Moreover, consider the adding of the choice of normalization between normalisation to have trace equal to n_channels or that determinant = 1 (This normalization is much more interesting geoemtry-wise as SPD matrices with unit determinant it is a submanifold of SPD matrices that is well known).

pyriemann/utils/covariance.py
```diff
@@ -28,90 +85,103 @@ def _fpcm(X, *, init=None, tol=10e-3, n_iter_max=50, assume_centered=False):
...
-    .. [3] `Covariance structure maximum-likelihood estimates in compound
-        Gaussian noise: Existence and algorithm analysis
-        <https://hal.science/hal-01816367/document>`_
-        F. Pascal, Y. Chitour, J.P. Ovarlez, P. Forster, P. Arzabal. IEEE
-        Transactions on Signal Processing, 2008.
+        D.E. Tyler. The Annals of Statistics, 1987.
     """  # noqa
     n_channels, n_times = X.shape
+
+    if m_estimator == 'tyl':
```
AmmarMian, line 128: Other M-estimators also exist:
image

Extract from Gordana Draskovich's Thesis which isn't public but which I can send you privately.

qbarthelemy: Thx for this nice summary table!
Other estimators can be added later, if required by pyRiemann users.

agramfort (approved):
@qbarthelemy I let you merge if done/happy

AmmarMian (approved):

qbarthelemy:
First step for #204, generalizing matrix operators to complex-valued inputs.

PR #225, opened 2023-02-15
Small corrections for robust M-estimators

qbarthelemy:
Small corrections for #223, after a call with @AmmarMian:

- use relative error as stopping criterion for fixed point estimation;
- use Huber's estimator in example.


agramfort (approved):

agramfort:
thx @qbarthelemy

Issue #224, opened 2023-02-14
Speed up pairwise_distance


agramfort:
currently the pairwise_distance function does a double for loop to compute all distances. It makes the computation quadratic on all levels. I should be possible to avoid this. Eg certain metrics are symmetric so we can divided by 2 and maybe for the others we can avoid computing more than once the eigenvecs/values of each matrix.

a fun challenge @qbarthelemy :)

qbarthelemy:
When Y is None, only half of the pairwise distance matrix is computed, like [scipy.pdist](https://github.com/scipy/scipy/blob/v1.10.0/scipy/spatial/distance.py#L2592):
https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/utils/distance.py#L331
Remark that it is false for non-symmetric divergences.

Storing all EVDs, and then computing pairwise distances seems a good idea to speed up pairwise AIR distances.
But I do not see how to generalize to any kind of distance.

SciPy uses a double for-loop for [scipy.cdist](https://github.com/scipy/scipy/blob/v1.10.0/scipy/spatial/distance.py#L2598).

qbarthelemy:
When Y is None, only half of the pairwise distance matrix is computed, like [scipy.pdist](https://github.com/scipy/scipy/blob/v1.10.0/scipy/spatial/distance.py#L2592):
https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/utils/distance.py#L331
Remark that it is false for non-symmetric divergences.

Storing all EVDs, and then computing pairwise distances seems a good idea to speed up pairwise AIR distances.
But I do not see how to generalize to any kind of distance.

SciPy uses a double for-loop for [scipy.cdist](https://github.com/scipy/scipy/blob/v1.10.0/scipy/spatial/distance.py#L2598).

agramfort:
I think we can definitely do better

<img width="797" alt="Screenshot 2023-02-18 at 16 05 16" src="https://user-images.githubusercontent.com/161052/219876170-b04fd22b-e04c-4341-8e3f-01297b89643a.png">

agramfort:
I am afraid so to some extent

gabelstein:
I wrote the pairwise functions for euclid, harmonic, logeuclid and riemann. 
The rest does not seem as easy to do as these four. Should we just go with these four for the time being? I would strongly propose so as the speedup is immense and the other metrics are probably not heavily used.
Maybe someone else will do the rest some day.

[...]

PR #229, opened 2023-02-20
Correct pairwise_distance for non symmetric metrics

qbarthelemy:
As noted in https://github.com/pyRiemann/pyRiemann/issues/224#issuecomment-1433562379, `pairwise_distance` is currently false for non-symmetric metrics.


sylvchev (approved):
Thanks for the upgraded test

agramfort (approved):
we should backport this to the maint/0.4 branch with the removable of pandas and make a 0.4.1 release soon

PR #230, opened 2023-02-21
Correct mean_covariance when used with keyword arguments

qbarthelemy:
`mean_covariance(covmats, metric='riemann', maxiter=10)`
raises the following error
`TypeError: mean_covariance() got an unexpected keyword argument 'maxiter'`

### DIFF
--- a/pyriemann/utils/mean.py
+++ b/pyriemann/utils/mean.py
@@ -577,7 +577,7 @@ def mean_covariance(covmats, metric='riemann', sample_weight=None, **kwargs):
         'wasserstein', or a callable function.
     sample_weight : None | ndarray, shape (n_matrices,), default=None
         Weights for each matrix. If None, it uses equal weights.
-    kwargs : list of params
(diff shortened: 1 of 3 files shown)


pyriemann/utils/mean.py
```diff
@@ -577,18 +577,22 @@ def mean_covariance(covmats, metric='riemann', sample_weight=None, *args):
         'wasserstein', or a callable function.
     sample_weight : None | ndarray, shape (n_matrices,), default=None
         Weights for each matrix. If None, it uses equal weights.
-    args : list of params
-        The arguments passed to the sub function.
+    kwargs : list of params
+        The keyword arguments passed to the sub function.
```
agramfort, line 581:
```suggestion
    **kwargs : dict
        The keyword arguments passed to the sub function.
```

it the numpydoc way

See https://numpydoc.readthedocs.io/en/latest/example.html

sylvchev (approved):

PR #232, opened 2023-02-28
Add function make_matrices to generate SPD, SPSD, HPD and HPSD matrices

qbarthelemy:
This PR:
- adds function `make_matrices` to generate SPD, SPSD, HPD and HPSD matrices;
- deprecates function `make_covariances`, limited to SPD matrices;
- updates tests.

Related to #204

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -183,6 +183,7 @@ Datasets
     make_gaussian_blobs
     make_outliers
     make_covariances
+    make_matrices
     make_masks
     sample_gaussian_spd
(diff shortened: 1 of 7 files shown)


qbarthelemy:
I needed to update tests because, before with `make_covariances`, all the tests used covariance matrices sharing the same eigen vectors, which was a too particular case. Using different eigen vectors, `make_matrices` performs more generic tests.

agramfort:
thx @qbarthelemy

PR #233, opened 2023-03-05
reduce code duplication (found with codeclimate)

agramfort:


sylvchev (approved):
Good catch for the code duplication.
The rewrite of the spatialfilter and classification tests made the tests more clear. I was not satisfied with the class-based approach for tests, especially the `test_two_classes` and `test_multi_classes` methods that seems very artificial. It is simpler and more readable with direct test functions, so thank you @agramfort !

PR #235, opened 2023-03-09
Complete doc and tests for geodesics for HPD matrices

qbarthelemy:
Complete doc for geodesics for HPD matrices
Refactor and complete tests for geodesics for HPD matrices


agramfort:
❤️

PR #239, opened 2023-03-23
Use _check_distance in utils.distance

qbarthelemy:
`_check_distance` was not used in `utils.distance`


agramfort (approved):

agramfort:
thx @qbarthelemy

PR #243, opened 2023-04-07
Enhance mean module to process HPD matrices

qbarthelemy:
Use actually `_check_mean` in `mean_covariance`;
Enhance functions `mean_ale` and `mean_alm`, removing np.diag(np.diag());
Enhance all means to process HPD matrices (excepted `mean_ale`, blocked by #238), related to #204;
Complete doc and tests;
Simplify code for medians.

### DIFF
--- a/pyriemann/utils/mean.py
+++ b/pyriemann/utils/mean.py
@@ -86,9 +86,6 @@ def mean_alm(covmats, tol=1e-14, maxiter=100, sample_weight=None):
                      X_1^{-\frac{1}{2}})^{\frac{1}{2}} X_1^{\frac{1}{2}}
 
     and requiring a high number of iterations.
-
-    This is the adaptation of the Matlab code proposed by Dario Bini and
-    Bruno Iannazzo, http://bezout.dm.unipi.it/software/mmtoolbox/ .
     Extremely slow, due to the recursive formulation.
 
     Parameters
--- a/pyriemann/utils/mean.py
+++ b/pyriemann/utils/mean.py
@@ -54,6 +54,7 @@ def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
     # init with AJD
     B = ajd_pham(covmats)[0]
 
+    eye_n = np.eye(n)
     crit = np.inf
     for _ in range(maxiter):
         J = np.einsum(
@@ -64,7 +65,7 @@ def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
         delta = np.real(np.diag(expm(J)))
         B = (np.abs(delta) ** -.5)[:, np.newaxis] * B
 
-        crit = distance_riemann(np.eye(n), np.diag(delta))
+        crit = distance_riemann(eye_n, np.diag(delta))
         if crit <= tol:
             break
     else:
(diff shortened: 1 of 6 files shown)


pyriemann/utils/mean.py
```diff
@@ -14,12 +14,12 @@
 def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
```
agramfort, line 14: to go all the way covmats param could be renamed but that's a deprecation cycle whichi is never fun...

pyriemann/utils/mean.py
```diff
@@ -30,52 +30,60 @@ def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
...
-    r"""Ando-Li-Mathias (ALM) mean of SPD matrices.
+    r"""Ando-Li-Mathias (ALM) mean of SPD/HPD matrices.
 
     Return the geometric mean recursively [1]_, generalizing from:
 
     .. math::
-        \mathbf{C} = A^{\frac{1}{2}}(A^{-\frac{1}{2}}B^{\frac{1}{2}}
-                     A^{-\frac{1}{2}})^{\frac{1}{2}}A^{\frac{1}{2}}
+        \mathbf{C} = X_1^{\frac{1}{2}} (X_1^{-\frac{1}{2}}X_2^{\frac{1}{2}}
+                     X_1^{-\frac{1}{2}})^{\frac{1}{2}} X_1^{\frac{1}{2}}
```
agramfort, line 86: we should make consistent how we document the mean equations. Here you only have 2 matrices but one can have n_matrices > 2.

agramfort (approved):
feel free to take my comments into account in a follow up PR

thx @qbarthelemy

pyriemann/utils/mean.py
qbarthelemy: Yes, my sweetest dream is to rename all `covmats` into `X` !
We could plan to group all deprecations in a release. What do you think? @agramfort @sylvchev

pyriemann/utils/mean.py
qbarthelemy: I am afraid that recursive nature of this mean prevents to write an analytic formula.
@sylvchev , what do you think?

qbarthelemy: By the way, the link http://bezout.dm.unipi.it/software/mmtoolbox/ does not exist.

sylvchev: The ALM is useful for a sanity check as it is a reference in geometry but this is not scale due to the recursion, so this is a special case.

sylvchev (approved):

sylvchev:
Thanks @qbarthelemy !

pyriemann/utils/mean.py
qbarthelemy: @agramfort , what is the best way to do this deprecation for module `utils.mean`?
Renaming argument `covmats` into `X`?
Renaming argument and transforming it into a positional-only argument?

agramfort: example with new signature:

def mean_ale(X, tol=10e-7, maxiter=50, sample_weight=None, covmats=None):

then if covmats is not None you raise a deprecation warning that covmats has been renamed to X and that this break in the next release.

clear?

qbarthelemy: Yes, it's clear and elegant. Thx!

PR #246, opened 2023-05-27
Add squared argument to distances

qbarthelemy:
Many distances need to compute a square root to output final value,
but many algorithms require the squared distances. 

Adding a squared argument to all distances allows to avoid these useless computations.


agramfort (approved):

agramfort:
Thx @qbarthelemy

PR #247, opened 2023-06-05
Correct transform and predict_proba of MeanField

qbarthelemy:
Classifier `MeanField` has been added in https://github.com/pyRiemann/pyRiemann/pull/172 .

Code had been adapted from
https://github.com/plcrodrigues/means-field-classifier/blob/master/power_means.py#L42
where `transform` function used squared distances.
But `predict_proba` has been copied from MDM, using squared distances too.

Finally, in `predict_proba`, before applying softmax, distances are magnified by power 4.

I suggest to remove squared distances from transform to be homogenous to MDM.

@plcrodrigues

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -29,6 +29,8 @@ v0.5.dev
 
 - Add argument ``squared`` to all distances. :pr:`246` by :user:`qbarthelemy`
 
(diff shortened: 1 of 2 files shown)


agramfort (approved):
@plcrodrigues I let you approve and merge

plcrodrigues (approved):
Sounds good.

Issue #248, opened 2023-06-08
Test failing on s390x, aarch64, ppc64le


sanjayankur31:
Hello,

When updating the Fedora package to 0.4, we see this one test failing on s390x and aarch64.

On aarch64:

```
=================================== FAILURES ===================================
_______________________ test_distance_mahalanobis[True] ________________________
rndstate = RandomState(MT19937) at 0xFFFF6BB67940, complex_valued = True
    @pytest.mark.parametrize("complex_valued", [True, False])
    def test_distance_mahalanobis(rndstate, complex_valued):
        n_channels, n_times = 3, 100
        X = rndstate.randn(n_channels, n_times)
        if complex_valued:
            X = X + 1j * rndstate.randn(n_channels, n_times)
        dist = distance_mahalanobis(X, np.cov(X))
        assert dist.shape == (n_times,)
>       assert np.all(np.isreal(dist))
E       assert False
E        +  where False = <function all at 0xffff7aba7a60>(array([False, False, False, False, False, False, False, False, False,\n       False, False, False, False, False, False,...False, False, False, False, False,\n       False, False, False, False, False, False, False, False, False,\n       False]))
E        +    where <function all at 0xffff7aba7a60> = np.all
E        +    and   array([False, False, False, False, False, False, False, False, False,\n       False, False, False, False, False, False,...False, False, False, False, False,\n       False, False, False, False, False, False, False, False, False,\n       False]) = <function isreal at 0xffff7a9f8b80>(array([0.74238958+3.32289838e-18j, 1.56911221+5.89993045e-19j,\n       1.13067829-4.57724315e-18j, 2.23289203+3.4041728...0.86465782+1.17541044e-19j, 1.35724931-4.13169590e-18j,\n       2.24127347-7.31375023e-18j, 2.226908  -1.62186528e-17j]))
E        +      where <function isreal at 0xffff7a9f8b80> = np.isreal
tests/test_utils_distance.py:213: AssertionError
```

on s390x

```
_______________________ test_distance_mahalanobis[True] ________________________
rndstate = RandomState(MT19937) at 0x3FF52FEF940, complex_valued = True
    @pytest.mark.parametrize("complex_valued", [True, False])
    def test_distance_mahalanobis(rndstate, complex_valued):
        n_channels, n_times = 3, 100
        X = rndstate.randn(n_channels, n_times)
        if complex_valued:
            X = X + 1j * rndstate.randn(n_channels, n_times)
[...]

PR #249, opened 2023-06-09
Keep real part of Mahalanobis distance

qbarthelemy:
Solves #248,
keeping real part of Mahalanobis distance, 
and updating tests for test functions.


sylvchev (approved):

sylvchev:
Thanks @qbarthelemy !

PR #250, opened 2023-06-16
Corrected a bug on the rejection_sampling_2D algorithm and updated th…

mhurte:
Hi everyone,

I would like to share a fix for a bug that I found as I was testing some sampling features of this library.

As I tried to compute the acceptance rate for the rejection sampling algorithm in 2D that is implemented in the _sampling.py_ file, I noticed that something was odd  : The following figure shows the acceptance rate as a function of sigma before the fix, compaired to the theoretical one for this method.

<p align="center">

Since we  know what the theoretical acceptance rate is for this algorithm, it was easy to notice that something was wrong when plotting the figure above. The acceptance rate is quickly converging to 1 when it should be converging to zero.

After searching through the code what the matter could be, I eventually found out that the roles of the _mu_a_ and _mu_b_ as defined in the functions `_rejection_sampling_2D_gfunction_plus`, `_rejection_sampling_2D_gfunction_minus` and `_rejection_sampling_2D` were in fact inverted.

After fixing this, the program now behaves properly, the following figure shows the acceptance rate as a function of sigma after the fix.

<p align="center">

Moreover, I added a boolean parameter to `_rejection_sampling_2D`, _return_rate_ (defaulted to False) that allows to get the acceptance rate for the generation along with the sample if necessary.
I also used the `_pdf_r` function that was already implemented in the file in the functions  `_rejection_sampling_2D_gfunction_minus` and  `_rejection_sampling_2D_gfunction_plus` to make the code more readable.

The documentation has been updated accordingly.


pyriemann/datasets/sampling.py
```diff
@@ -129,24 +122,31 @@ def _rejection_sampling_2D(n_samples, sigma, random_state=None):
         Dispersion of the Riemannian Gaussian distribution.
     random_state : int, RandomState instance or None, default=None
         Pass an int for reproducible output across multiple function calls.
+    return_rate : boolean
```
plcrodrigues, line 125: I think that calling it `return_acceptance_rate` would be better, since one could think that it is the `rejection_rate` or any other bizarre rate.

plcrodrigues (commented):
Thanks @mhurte! This is very helpful.

pyriemann/datasets/sampling.py
mhurte: Indeed, I've changed it.

pyriemann/datasets/sampling.py
```diff
@@ -129,36 +122,50 @@ def _rejection_sampling_2D(n_samples, sigma, random_state=None):
...
+        else :
             r_sample = multivariate_normal.rvs(mu_b, cov_matrix, 1, rs)
             res = _rejection_sampling_2D_gfunction_minus(sigma, r_sample)
             if rs.rand(1) < res:
                 r_samples.append(r_sample)
                 cpt += 1
+
+    if (return_acceptance_rate) :
+        acceptance_rate = n_samples / acc
+        return np.array(r_samples), acceptance_rate
```
qbarthelemy, line 167:
```suggestion
    if return_acceptance_rate:
        return np.array(r_samples), n_samples / acc
```

qbarthelemy (changes requested):
Thx @mhurte for this contribution!
Can you complete whatsnew file, and complete tests with the new parameter?

plcrodrigues:
In fact, it seems that to run `pytest` locally one has to do:

> python -m pytest tests/

Source: https://stackoverflow.com/a/34140498/1898001

qbarthelemy (approved):

PR #252, opened 2023-06-16
Deprecate input covmats for mean functions

qbarthelemy:
To imitate median functions,
this PR deprecates inputs `covmats` into `X` for mean functions, and renames outputs `C` into `M`.

### DIFF
--- a/pyriemann/utils/mean.py
+++ b/pyriemann/utils/mean.py
@@ -11,7 +11,15 @@ from .geodesic import geodesic_riemann
 from .utils import check_weights
 
 
-def mean_ale(X, tol=10e-7, maxiter=50, sample_weight=None, covmats=None):
+def _deprecate_covmats(covmats, X):
+    if covmats is not None:
+        print("DeprecationWarning: input covmats has been renamed into X and "
+              "will be removed in 0.6.0.")
+        X = covmats
+    return X
+
+
+def mean_ale(X=None, tol=10e-7, maxiter=50, sample_weight=None, covmats=None):
     """AJD-based log-Euclidean (ALE) mean of SPD matrices.
 
     Return the mean of a set of SPD matrices using the approximate joint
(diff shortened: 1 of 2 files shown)


pyriemann/utils/mean.py
```diff
@@ -48,20 +48,19 @@ def mean_ale(covmats, tol=10e-7, maxiter=50, sample_weight=None):
         <https://arxiv.org/abs/1505.07343>`_
         M. Congedo, B. Afsari, A. Barachant, M. Moakher. PLOS ONE, 2015
     """
-    n_matrices, n, _ = covmats.shape
+    if covmats is not None:
+        warnings.warn("Input covmats has been renamed into X and will be "
+                      "removed in 0.6.0.", category=DeprecationWarning)
```
agramfort, line 53: You miss here X=covmats

agramfort (commented):
I would use a helper function private to avoid the warning code dupe. Function takes X and covmats and returns X

pyriemann/utils/mean.py
qbarthelemy: Is this function ok for you?
```
def _deprecate_covmats(covmats, X):
    if covmats is not None:
        print("DeprecationWarning: input covmats has been renamed into X "
               "and will be removed in 0.6.0.")
        X = covmats
    return X
```

agramfort: yes and the signature should actually be 

mean_func(X=None, ..., covmats=None)

so doing

mean_func(covmats=covmats)

still works

qbarthelemy: Ok. But `mean_power` and `maskedmean_riemann` can not support `X=None`.

agramfort: what do you mean?

qbarthelemy:
```
def maskedmean_riemann(X=None, masks, tol=10e-9, maxiter=100, init=None,
                       sample_weight=None, covmats=None):
```
raises `Non default argument follows defaut argument`. 
To solve that, I must define `X=None, masks=None` and raise an error if masks=None...

Same problem for
```
def mean_power(X, p, *, sample_weight=None, zeta=10e-10, maxiter=100,
               covmats=None):
```

agramfort:
> To solve that, I must define X=None, masks=None and raise an error if
masks=None...

+1

PR #254, opened 2023-06-27
Update to Python-3.8

qbarthelemy:
Update to Python-3.8

Update code to support SciPy 1.11.0 and scikit-learn 1.3.0.

Update readme for version 0.5, see #253.

### DIFF
--- a/pyriemann/classification.py
+++ b/pyriemann/classification.py
@@ -35,6 +35,17 @@ def _check_metric(metric):
     return metric_mean, metric_dist
 
 
+def _mode_1d(X):
+    vals, counts = np.unique(X, return_counts=True)
+    mode = vals[counts.argmax()]
+    return mode
+
+
+def _mode_2d(X, axis=1):
+    mode = np.apply_along_axis(_mode_1d, axis, X)
+    return mode
+
+
 class MDM(BaseEstimator, ClassifierMixin, TransformerMixin):
     """Classification by Minimum Distance to Mean.
 
@@ -523,8 +534,8 @@ class KNearestNeighbor(MDM):
         """
         dist = self._predict_distances(covtest)
         neighbors_classes = self.classmeans_[np.argsort(dist)]
-        out = np.unique(neighbors_classes[:, 0:self.n_neighbors], axis=1)[:, 0]
-        return out
+        pred = _mode_2d(neighbors_classes[:, 0:self.n_neighbors], axis=1)
+        return pred
 
     def predict_proba(self, X):
         """Predict proba using softmax.
--- a/pyriemann/classification.py
+++ b/pyriemann/classification.py
@@ -2,7 +2,6 @@
 import functools
 
 import numpy as np
-from scipy import stats
 from sklearn.base import BaseEstimator, ClassifierMixin, TransformerMixin
 from sklearn.svm import SVC as sklearnSVC
 from sklearn.utils.extmath import softmax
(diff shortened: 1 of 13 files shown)


qbarthelemy:
`scipy.stats.mode` for non numeric data is deprecated from SciPy 1.9.0 and is removed in SciPy 1.11.0.

Tests are broken, because transfer learning module uses non numeric extended labels. 

@plcrodrigues , any idea?

qbarthelemy:
I solve the problem, replacing `scipy.stats.mode` by `np.unique`, which supports non numeric data.

plcrodrigues:
> @plcrodrigues , any idea?

This is indeed annoying but doesn't sound too hard to bypass with your `np.unique` approach.

In any case, we should soon take some time to make use of the new `sample-props` feature from `sklearn` -- [PR #24027](https://github.com/scikit-learn/scikit-learn/pull/24027) -- in the transfer learning module of `pyRiemann`. This would avoid recurring to string in the labels.

pyriemann/classification.py
```diff
@@ -524,8 +523,8 @@ def predict(self, covtest):
         """
         dist = self._predict_distances(covtest)
         neighbors_classes = self.classmeans_[np.argsort(dist)]
-        out, _ = stats.mode(neighbors_classes[:, 0:self.n_neighbors], axis=1)
-        return out.ravel()
+        out = np.unique(neighbors_classes[:, 0:self.n_neighbors], axis=1)[:, 0]
```
agramfort, line 526: I doubt this is correct

agramfort: you would need to use return_counts and get the entry in out that has the biggest count

qbarthelemy: You are right. Solution is a little bit more complicated
https://stackoverflow.com/a/12300214

I will try to find the simplest solution for 2D case.

qbarthelemy:
Last test for Windows Python 3.7 which is OK installed joblib 1.2.0
https://github.com/pyRiemann/pyRiemann/actions/runs/5321719371/jobs/9637189077#step:5:24
[5:10](https://bciparis.slack.com/archives/G010MU5D3UH/p1688137804132769)

Now, it is KO installing joblib 1.3.1
https://github.com/pyRiemann/pyRiemann/actions/runs/5415226192/jobs/9846294273?pr=254#step:5:24

I freeze joblib to 1.2.0, but now, all others tests for Python 3.8 and 3.9 are KO.

Any idea?

qbarthelemy:
Waiting scikit-learn bugfix, I decide to detect when _init_centroids must be used with a non default sample_weight.
I also clean code, removing support for version before scikit-learn 0.24.

lesteve:
FYI I had a look and the joblib 1.3.1 issue is actually due to loky (default joblib parallel backend). The loky issue is tracked in https://github.com/joblib/loky/issues/411.

Not entirely sure this is going to get fixed in loky. FYI scikit-learn requires Python 3.8+ since scikit-learn 1.1 (released May 2022), so a possible work-around for pyRiemann would be to drop Python 3.7 support in the 0.5 release.

qbarthelemy:
Close-reopen for tests.

qbarthelemy:
Documentation fails, it is probably related to #255. @agramfort, what do you think ?

agramfort (approved):
thx @qbarthelemy

qbarthelemy:
For documentation build, I can't solve the dependency problem between sphinx and jinja2.

qbarthelemy:
Doc is now OK!
I let you merge @agramfort , @plcrodrigues , @sylvchev

agramfort (approved):

PR #256, opened 2023-07-26
speedup pairwise_distance

### LINKED ISSUE #224: Speed up pairwise_distance

gabelstein:
closes #224

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,7 +10,7 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.6.dev
 --------
 
-- Speedup pairwise distance function :func:`pyriemann.utils.distance.pairwise_distance` by adding individual functions.
+- Speedup pairwise distance function :func:`pyriemann.utils.distance.pairwise_distance` by adding individual functions. :pr:`256` by :user:`gabelstein`
 
 v0.5 (Jun 2023)
 ---------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,6 +10,7 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.6.dev
 --------
 
+- Speedup pairwise distance function :func:`pyriemann.utils.distance.pairwise_distance` by adding individual functions.
 
 v0.5 (Jun 2023)
 ---------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,7 +10,7 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.6.dev
 --------
 
-- Speedup pairwise distance function :func:`pyriemann.utils.distance.pairwise_distance` by adding individual functions. :pr:`256` by :user:`gabelstein`
+- Speedup pairwise distance function :func:`pyriemann.utils.distance.pairwise_distance` by adding individual functions for 'euclid', 'harmonic', 'logeuclid' and 'riemann' metrics. :pr:`256` by :user:`gabelstein`
 
 v0.5 (Jun 2023)
 ---------------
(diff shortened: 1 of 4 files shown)


gabelstein:
Slight problem as logdet and wasserstein distances don't return 0 for same inputs.

gabelstein:
Probably due to numerical instability of numpy.linalg.slogdet.
Also just caught this because of actually checking correctness of pairwise distance matrix. So not a new issue anyways.
I think there's no good fix for this, slogdet is much faster than anything else. 
Maybe just return 0 after a check if the arrays are equal?
Otherwise I just reduced the required precision in the test for now.

gabelstein:
@agramfort @qbarthelemy

agramfort (commented):
basically I would avoid the new public functions (make them private) and I would not do any magic with "np.array_equal(X, Y)" if Y=X then user should just pass Y = None. You should now what you are doing.
Finally do not set the diag to zero. If it's 1e-15 than it's a numerical zero. Basically don't hack the numerical noise.

agramfort:
it's better to fix LLE then

gabelstein:
> it's better to fix LLE then

@agramfort doesn't need fixing, I remembered incorrectly. 
Can this be merged then?

agramfort:
Can you see why CIs fail ? Now from an API perspective I think we should have only one public function called pairwise_distances that takes a metric parameter. It will make the API lighter.

gabelstein:
> Can you see why CIs fail ?

it says "FileNotFoundError: [Errno 2] No such file or directory: '/tmp/sphinx-log'" which I don't really understand. I thought it was an issue on the server side.

> Now from an API perspective I think we should have only one public function called pairwise_distances that takes a metric parameter. It will make the API lighter.

I removed the individual functions from the api page, I thought that was enough. 
Is there  way to make the functions other functions private?

agramfort:
yes if it does not start with some _ then it's a public function. Can you
prepend some _ the function name for each
function you want to keep private? thx

gabelstein:
> prepend some _ the function name for each

Done.

pyriemann/utils/distance.py
```diff
@@ -446,6 +447,177 @@ def distance(A, B, metric='riemann', squared=False):
...
+    Notes
+    -----
+    .. versionadded:: 0.6
+
+    See Also
+    --------
+    pairwise_distance
+    distance_harmonic
+    """
+    if isinstance(Y, type(None)):
```
agramfort, line 520:
```suggestion
    if Y is None:
```

pyriemann/utils/distance.py
```diff
@@ -446,6 +447,177 @@ def distance(A, B, metric='riemann', squared=False):
...
+
+    Returns
+    -------
+    dist : ndarray, shape (n_matrices_X, n_matrices_X) or (n_matrices_X, \
+            n_matrices_Y)
+        Harmonic Distances between pairs of elements of X if Y is None, or
+        between elements of X and Y.
+    Notes
+    -----
+    .. versionadded:: 0.6
```
agramfort, line 513: You don't need these now that it's private

doc/whatsnew.rst
```diff
@@ -10,6 +10,7 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.6.dev
 --------
 
+- Speedup pairwise distance function :func:`pyriemann.utils.distance.pairwise_distance` by adding individual functions. :pr:`256` by :user:`gabelstein`
```
agramfort, line 13: Say for what metric it's faster

qbarthelemy (approved):
Thx @gabelstein for this nice contribution!
We can merge after #254

agramfort:
Please update the branch so the CI can turn green now

gabelstein:
@qbarthelemy 's changes just gave me an idea. Setting the invY and logY to none when applicable speeds up the internal sklearn pairwise distance. 
image

qbarthelemy:
Thx @gabelstein

PR #259, opened 2023-08-21
Deprecate input covtest for predict function of KNN

qbarthelemy:
This PR deprecates inputs `covtest` into `X` for predict function of KNN,

### DIFF
--- a/pyriemann/classification.py
+++ b/pyriemann/classification.py
@@ -536,6 +536,7 @@ class KNearestNeighbor(MDM):
         if covtest is not None:
             warnings.warn("Input covtest has been renamed into X and will be "
                           "removed in 0.8.0.", category=DeprecationWarning)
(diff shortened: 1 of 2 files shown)


pyriemann/classification.py
```diff
@@ -532,7 +533,12 @@ def predict(self, covtest):
         pred : ndarray of int, shape (n_matrices,)
             Predictions for each matrix according to the closest centroid.
         """
-        dist = self._predict_distances(covtest)
+        if covtest is not None:
+            warnings.warn("Input covtest has been renamed into X and will be "
+                          "removed in 0.8.0.", category=DeprecationWarning)
+            X = covtest
```
agramfort, line 539: you should here assert X is None in this case

agramfort (approved):

PR #270, opened 2023-11-13
Block covariance new

gabelstein:
Due to the rewriting of the _matrix_operator function using numpys eigh, it's possible to do efficient block matrix calculations.
It now finally only does the computations on the blocks (as was already mentioned by @qbarthelemy in the old pull request #154 ) which is 10x faster in my use case. 
However, the BlockMatrix class I needed to implement only works for same size blocks. So this functionality would have to be removed. 
Another problem is that the joblib backend screws with the BlockMatrix structure and it doesn't work (which I don't really find very important but should probably be noted somewhere).
What do you think?

### DIFF
--- a/examples/simulated/plot_classifier_comparison.py
+++ b/examples/simulated/plot_classifier_comparison.py
@@ -42,8 +42,9 @@ from pyriemann.classification import (
 @partial(np.vectorize, excluded=['clf'])
 def get_proba(cov_00, cov_01, cov_11, clf):
     cov = np.array([[cov_00, cov_01], [cov_01, cov_11]])
-    with np.testing.suppress_warnings() as sup:
-        sup.filter(RuntimeWarning)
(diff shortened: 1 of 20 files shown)


qbarthelemy:
Thank you @gabelstein  for this interesting suggestion!

Positive point: code acceleration by performing efficient block matrix calculations.

Negative point: removal of the processing of bocks of different sizes (in any case, the depreciation of this functionality will have to be managed over 2 releases), and modification of the _matrix_operator.

What do you think @agramfort @sylvchev @plcrodrigues ?

gcattan:
Did you reach an agreement on what is going to be implemented?

> It would also be a possibility to keep the current implementation and only use the new implementation when all blocks have the same size. The _matrix_operator would need to be changed anyway.

Making sense to me.

bruAristimunha:
@sylvchev, @sylvchev

Issue #271, opened 2023-11-15
kernel must be 'precomputed' or callable, is precomputed.


nicobrb:
Given the model 
```
model = process.SVC(metric="logeuclid", kernel_fct="precomputed", C=0.1,
                                    shrinking=True, cache_size=1024,
                                    class_weight="balanced")
```

Then calling

`model.fit(X, y)`

returns 
`kernel must be 'precomputed' or callable, is precomputed.`

This is due to this piece of code in `pyriemann.classification.py`:

![image](https://github.com/pyRiemann/pyRiemann/assets/39137998/4b14584c-2840-4d87-8880-5c48b30b707b)

Which clearly doesn't check where `kernel_fct = "precomputed"`, but only if it's callable or None.

qbarthelemy:
Good catch! This problem will be solved with #272.

In the meantime you can use `kernel_fct=None` to achieve the same behavior.

qbarthelemy:
Closed with #272

PR #272, opened 2023-11-15
Correct check for kernel_fct param of SVC

qbarthelemy:
Solves #271

Correct check for `kernel_fct` param of SVC, and complete tests.

@gabelstein , can you confirm that `None` and `"precomputed"` must provide the same behavior?


gabelstein:
Ah, I think the 'precomputed' option is from an older implementation of the SVC and I forgot to remove it. 
But yes, now that it's in there, it should do the same as None.

agramfort (approved):

PR #274, opened 2023-12-14
Enhance 'lwf', 'mcd', 'oas' and 'sch' covariance estimator to process complex inputs

gabelstein:
As noted in #204, this implements general complex covariance estimation described [here](https://ieeexplore.ieee.org/document/4217572). 

I put the transformation in the individual estimators like _oas, _lwf etc., that made most sense to me.

What do you thin, @qbarthelemy?


pyriemann/utils/covariance.py
```diff
@@ -11,41 +11,69 @@
...
 def _lwf(X, **kwds):
     """Wrapper for sklearn ledoit wolf covariance estimator"""
-    if not is_real_type(X):
-        raise ValueError("Input must be real-valued.")
+    iscomplex = np.iscomplexobj(X)
+    if iscomplex:
+        X = np.concatenate((X.real, X.imag), axis=0)
     C, _ = ledoit_wolf(X.T, **kwds)
+    if iscomplex:
+        C = _make_complex_covariance(C)
```
agramfort, line 19: you can do this with a tiny decorator to avoid the copy / paste below in all functions

pyriemann/utils/covariance.py
```diff
@@ -357,12 +394,16 @@ def covariances(X, estimator='cov', **kwds):
...
+        R. Abrahamsson, Y. Selen and P. Stoica. 2007 IEEE International
+        Conference on Acoustics, Speech and Signal Processing, Volume 2, 2007.
     """  # noqa
     est = _check_cov_est_function(estimator)
     n_matrices, n_channels, n_times = X.shape
-    covmats = np.empty((n_matrices, n_channels, n_channels), dtype=X.dtype)
-    for i in range(n_matrices):
-        covmats[i] = est(X[i], **kwds)
+    covmats = np.asarray([est(X[i], **kwds)
+                          for i in range(n_matrices)])
```
agramfort, line 406: this is less efficient

pyriemann/utils/covariance.py
```diff
@@ -393,10 +434,9 @@ def covariances_EP(X, P, estimator='cov', **kwds):
     if n_times_p != n_times:
         raise ValueError(
             f"X and P do not have the same n_times: {n_times} and {n_times_p}")
-    covmats = np.empty((n_matrices, n_channels + n_channels_proto,
-                        n_channels + n_channels_proto))
-    for i in range(n_matrices):
-        covmats[i] = est(np.concatenate((P, X[i]), axis=0), **kwds)
+    covmats = np.asarray([est(np.concatenate((P, X[i]), axis=0), **kwds)
+                          for i in range(n_matrices)])
```
agramfort, line 438: idem it's more efficient as done before

qbarthelemy (changes requested):
Many thanks for this contribution which solves a crucial issue!

pyriemann/utils/covariance.py
gabelstein: ah sorry, I always forget about python decorators. will look into it.

gabelstein: Not sure if decorators need documentations, but I added a few lines.

gabelstein:
Thank you for the suggestions, @qbarthelemy and @agramfort. 
I wrote a small decorator, but this is the first time I used this so someone should probably check if it's correctly written (and documented). 
Also, I added a test for the decorator (which also tests the method from the paper for np.cov). Not sure if it makes sense like that so someone who knows more about testing please look into this.

.gitignore
```diff
@@ -10,3 +10,4 @@
 *.egg-info/
 .vscode/
 sandbox/
+.idea/
```
qbarthelemy, line 13: Can you revert this modification?

agramfort (approved):
@qbarthelemy I let you merge if happy

thx @gabelstein

qbarthelemy (approved):
Thx @gabelstein for this great improvement !

PR #275, opened 2024-01-02
Rename HankelCovariances into TimeDelayCovariances

qbarthelemy:
I wanted to explicitly test if `HankelCovariances`  were Hankel matrices, ie with constant anti-diagonals. And that's not the case.

I therefore suggest to deprecate the name to `TimeDelayCovariances`.

### DIFF
--- a/pyriemann/estimation.py
+++ b/pyriemann/estimation.py
@@ -640,14 +640,6 @@ class Coherences(CospCovariances):
         return np.array(out)
 
 
-@deprecated(
-    "HankelCovariances is deprecated and will be removed in 0.8.0; "
-    "please use TimeDelayCovariances."
-)
-class HankelCovariances(BaseEstimator, TransformerMixin):
-    pass
-
-
 class TimeDelayCovariances(BaseEstimator, TransformerMixin):
     """Estimation of covariance matrix with time delay matrices.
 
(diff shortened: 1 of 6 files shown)


agramfort (commented):
can you check with a test that HankelCovariances is properly deprecated? meaning it can still be called and a warning is raised? thx

pyriemann/estimation.py
```diff
@@ -638,11 +640,19 @@ def transform(self, X):
         return np.array(out)
 
 
+@deprecated(
+    "HankelCovariances is deprecated and will be removed in 0.8.0; "
+    "please use TimeDelayCovariances."
+)
 class HankelCovariances(BaseEstimator, TransformerMixin):
```
agramfort, line 647: I would say you need to inherit from TimeDelayCovariances as otherwise code that used to work will break. If I now import HankelCovariances in my code nothing works exept the instance creation without any parameter.

pyriemann/utils/test.py
```diff
@@ -55,6 +55,36 @@ def is_skew_sym(X):
...
+    """Check if matrix is an Hankel matrix.
+
+    Parameters
+    ----------
+    X : ndarray, shape (n, n)
+        Square matrix.
+
+    Returns
+    -------
+    ret : boolean
```
agramfort, line 68:
```suggestion
    ret : bool
```

pyriemann/utils/test.py
```diff
@@ -55,6 +55,36 @@ def is_skew_sym(X):
     return is_square(X) and np.allclose(X, -np.swapaxes(X, -2, -1))
 
 
+def is_hankel(X):
```
agramfort, line 58: can you add tiny tests for this function? I know it's used elsewhere but my rule of thumb is that any public function in foo.py should be tested in tests/test_foo.py

qbarthelemy:
@agramfort , I'm going to merge if you don't have any other comments.

PR #276, opened 2024-01-16
Example of ACM

carraraig:
Here an example of ACM (https://arxiv.org/abs/2302.04508). 
@qbarthelemy can you review this example?


agramfort:
@carraraig can you look into the CIs issues?

qbarthelemy (changes requested):
Thanks Igor for this contribution!

In your example, I suggest to:
- split data into training and test sets, in order to display test scores (instead of training scores);
- display optimal parameters `lag` and `order` found by cross-validation;
- evaluate all combinations of features (ie, CM vs ACM(grid)) and classifiers (MDM vs TS+VSM), so that we can understand if difference in results comes from features or from classifiers.
You can test: CM+TS+SVM, ACM(grid)+TS+SVM; CM+MDM, ACM(grid)+MDM (and optionally, CM+FgMDM, ACM(grid)+FgMDM).

qbarthelemy:
It is possible to split the example in two steps? @carraraig 
Step1: select optimal ACM hyper-parameters on training set and display them
Step2: train classifier using selected hyper-parameters on training set and evaluate performance using test set

See this example
https://pyriemann.readthedocs.io/en/latest/auto_examples/motor-imagery/plot_frequency_band_selection.html

carraraig:
Hi @qbarthelemy !

I have created a new version using your suggestion. In any case I think that the new procedure create results that are too much dependent on the data split in train and test. Honestly I would prefer to use a the Nested Cross validation scheme in order to get some statistic for the results and to avoid the dependence on which data points end up in the training set and which in the test set. 

Let me know what do you think on this.

qbarthelemy:
You are right, there is an arbitrary choice in cutting the data between training and test sets.

But using the same set for training (with or without nested CV) and test leads to over-optimistic results: the scores are always excellent but do not necessarily generalize to new data.

https://en.wikipedia.org/wiki/Training,_validation,_and_test_data_sets

The evaluation on several test sets from different databases takes place in the publications, not in pyRiemann (we prefer minimalist examples for a quick documentation build).

Is it possible to fix build errors?
https://github.com/pyRiemann/pyRiemann/actions/runs/7654176300/job/20865153954?pr=276

In the pipeline names, is it possible to rename the "ACM+*(Grid)" to "ACM(Grid)+*" to highlight that the grid is applied to the ACM feature, not to classifiers?

carraraig:
Hi @qbarthelemy !

Ok I agree with you that for simplicity could be better the train test split idea. I have removed the Nested example and solve the error.

examples/motor-imagery/plot_ACM.py
```diff
@@ -0,0 +1,282 @@
...
+    results.append(res)
+
+results = pd.DataFrame(results)
+
+# Calculate the mean score for each pipelines
+pipeline_stats = results.groupby('pipeline')['score'].agg(
+    ['mean']).sort_values(by='mean', ascending=False)
+
+for pipeline, stats in pipeline_stats.iterrows():
+    print(f"{pipeline}, score: {stats['mean']:.4f}")
```
qbarthelemy, line 256: There is no longer any need to calculate the mean. Code can be simplified into:
```suggestion
for _, row in results.sort_values(by='score', ascending=False).iterrows():
    print(f"{row.pipeline}, score: {row.score:.4f}")
```

qbarthelemy:
I modified the figure to better distinguish the impact of features (CM vs ACM) from the impact of classifiers (so, you have to git pull).

qbarthelemy (approved):
I simplified the AugmentedDataset code in last commit. Is it ok for you @carraraig?

carraraig:
Thanks @qbarthelemy , is ok for me!

qbarthelemy:
Thanks @carraraig!

When your article has been accepted, it would be interesting to see how to include this method into `TimeDelayCovariances`
(see https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/estimation.py#L643)
adding a `padding` argument:
- 'circular' by default (for backward compatibility),
- 'valid' for your new method.


```
    def transform(self, X):

        # ...

        if self.padding not in ["circular", "valid"]:
            raise ValueError("padding must be 'circular' or 'valid', got "
                             f"{self.padding}")

        if self.padding == "circular":
            # current code
        else:
            self.Xtd_ = np.concatenate(
                [
                    X[:, :, p * self.lag: -(self.order - p) * self.lag]
                    for p in range(0, self.order)
                ],
                axis=1,
            )

        covmats = ...
```

PR #279, opened 2024-01-31
Enhance MDWM

qbarthelemy:
This PR:
- converts MDM.covmeans_ attribute from a list of 2d ndarrays into a 3d ndarray,
- uses `check_weights` function,
- completes tests with weights.


agramfort:
Thx @qbarthelemy

PR #281, opened 2024-02-05
Speed up TimeDelayCovariances

qbarthelemy:
Speedup TimeDelayCovariances removing a for loop


pyriemann/estimation.py
```diff
@@ -727,13 +727,10 @@ def transform(self, X):
...
-        for x in X:
-            tmp = x
-            for d in delays:
-                tmp = np.r_[tmp, np.roll(x, d, axis=-1)]
-            Xtd.append(tmp)
-        self.Xtd_ = np.array(Xtd)
+        Xtd = X
+        for d in delays:
+            Xtd = np.concatenate((Xtd, np.roll(X, d, axis=-1)), axis=-2)
+        self.Xtd_ = Xtd
```
agramfort, line 733:
```suggestion
        Xtd = [X]
        for d in delays:
            Xtd.append(np.roll(X, d, axis=-1))
        self.Xtd_ = np.concatenate(Xtd, axis=-2)
```
this should be a bit faster as you only do one malloc with one np.concatenate

qbarthelemy: nice!

agramfort (approved):

PR #291, opened 2024-04-03
prepare release 0.6

qbarthelemy:


gabelstein:
@qbarthelemy should I prepare the extended kernel module for this or would you like to add it to the next release? I could probably finish it by the end of April.

agramfort (approved):

qbarthelemy:
@gabelstein , no, improvement of kernel module could be done in next release.

PR #292, opened 2024-04-05
Adding Euclidean Alignment implementation

brunaafl:
Hello!

I am adding an implementation of the Euclidean Alignment (He H. et al - 2019) used on Junqueira, B et al. (2024). The methods depend directly on the covariance estimators present on Pyriemann, and Sylvain thought about what would be a nice addition for Pyriemann.

In this re-implementation, I defined it as a scikit-learn transformation, and in a way that is not necessary to define a ‘target’ domain before. 

It would be easier to add this transformation as a preprocessing step in a pipeline, and also to integrate with MOABB, being able to use the pre-defined evaluation objects. 

This class is also adapted to align both raw trials and covariance matrices and to define the type of covariance estimator, in the case the input is raw data. Therefore, it can be used with different classifiers.

Since we were working with this transformation, we thought adding an implementation and a tutorial here in PyRiemann would be nice, but I would like to check first if the tutorial would be welcome.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015-2023, authors of pyRiemann
+Copyright (c) 2015-2024, authors of pyRiemann
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -94,7 +94,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'pyRiemann'
-copyright = u'2015-2023, pyRiemann Contributors'
+copyright = u'2015-2024, pyRiemann Contributors'
 author = u'Alexandre Barachant'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -7,8 +7,8 @@ What's new in the package
 
 A catalog of new features, improvements, and bug-fixes in each release.
 
-v0.6.dev
---------
+v0.6 (April 2024)
+-----------------
 
 - Update pyRiemann from Python 3.7 to 3.8. :pr:`254` by :user:`qbarthelemy`
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -7,6 +7,10 @@ What's new in the package
 
 A catalog of new features, improvements, and bug-fixes in each release.
 
+v0.7.dev
+--------
+
+
 v0.6 (April 2024)
 -----------------
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,7 +10,9 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.7.dev
 --------
 
-- Add kernel option to :class:`pyriemann.embedding.LocallyLinearEmbedding`. :pr:`293` by :user:`gabelstein`
+- Add ``kernel`` parameter to :class:`pyriemann.embedding.LocallyLinearEmbedding`. :pr:`293` by :user:`gabelstein`
+
+- Add possibility for ``target_domain`` parameter of :class:`pyriemann.transfer._estimators.TLCenter` to be empty, forcing ``transform()`` to recenter matrices to the last fitted domain. :pr:`292` by :user:`brunaafl`
 
 v0.6 (April 2024)
 -----------------
@@ -51,8 +53,6 @@ v0.6 (April 2024)
   by :user:`qbarthelemy` and :user:`gcattan`
 
 - Add :class:`pyriemann.estimation.CrossSpectra` and deprecate `CospCovariances` renamed into :class:`pyriemann.estimation.CoSpectra`. :pr:`288` by :user:`qbarthelemy`
-
-- Adapt :class:`pyriemann.transfer._estimators.TLCenter` to allow integration with MOABB pipelines and evaluations. :pr:`292` by :user:`brunaafl`
 
 v0.5 (Jun 2023)
 ---------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,6 +10,7 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.7.dev
 --------
 
+- Add kernel option to :class:`pyriemann.embedding.LocallyLinearEmbedding`. :pr:`293` by :user:`gabelstein`
 
 v0.6 (April 2024)
 -----------------
--- a/pyriemann/_version.py
+++ b/pyriemann/_version.py
@@ -1,1 +1,1 @@
-__version__ = '0.6'
+__version__ = '0.7.dev'
--- a/pyriemann/_version.py
+++ b/pyriemann/_version.py
@@ -1,1 +1,1 @@
-__version__ = '0.6.dev'
+__version__ = '0.6'
--- a/pyriemann/embedding.py
+++ b/pyriemann/embedding.py
@@ -1,14 +1,15 @@
 """Embedding SPD matrices via manifold learning techniques."""
 
+import warnings
+
 import numpy as np
 from scipy.linalg import solve, eigh
 from scipy.sparse import csr_matrix
-
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.manifold import spectral_embedding
 
-from .utils.kernel import kernel
 from .utils.distance import pairwise_distance
+from .utils.kernel import kernel as kernel_fct
 
 
 class SpectralEmbedding(BaseEstimator):
@@ -133,12 +134,20 @@ class LocallyLinearEmbedding(BaseEstimator, TransformerMixin):
 
     Parameters
     ----------
-    n_components : int, default=2
+    n_components : int | None, default=2
         Dimensionality of projected space.
-    n_neighbors : int, default=5
-        Number of neighbors for reconstruction of each point.
+        If None, ``n_components`` is set to ``n_matrices - 1``.
+    n_neighbors : int | None, default=5
+        Number of neighbors for reconstruction of each matrix.
+        If None, all available matrices are used.
+        If ``n_neighbors > n_matrices``, ``n_neighbors`` is set to
+        ``n_matrices - 1``.
     metric : {"euclid", "logeuclid", "riemann"}, default: "riemann"
         Metric used for KNN and Kernel estimation.
+    kernel : callable | None, default=None
+        Kernel function to use for the embedding. If None, the canonical
+        kernel specified by the metric is used. Must be a function that
+        takes the arguments (X, Cref, metric).
     reg : float, default=1e-3
         Regularization parameter.
 
@@ -169,12 +178,17 @@ class LocallyLinearEmbedding(BaseEstimator, TransformerMixin):
         Pattern Recognition
     """
 
-    def __init__(self, n_components=2, n_neighbors=5, metric="riemann",
-                 reg=1e-3):
+    def __init__(self, n_components=2,
+                 n_neighbors=5,
+                 metric="riemann",
+                 kernel=None,
+                 reg=1e-3
+                 ):
         self.n_components = n_components
         self.n_neighbors = n_neighbors
         self.metric = metric
         self.reg = reg
+        self.kernel = kernel
 
     def fit(self, X, y=None):
         """Fit the model from data in X.
@@ -193,7 +207,7 @@ class LocallyLinearEmbedding(BaseEstimator, TransformerMixin):
 
         """
         self.data_ = X
-        _check_dimensions(
+        self.n_components, self.n_neighbors = _check_dimensions(
             X,
             n_components=self.n_components,
             n_neighbors=self.n_neighbors,
@@ -205,6 +219,7 @@ class LocallyLinearEmbedding(BaseEstimator, TransformerMixin):
             n_neighbors=self.n_neighbors,
             metric=self.metric,
             reg=self.reg,
+            kernel=self.kernel
         )
 
         self.embedding_, self.error_ = embd, err
@@ -240,6 +255,7 @@ class LocallyLinearEmbedding(BaseEstimator, TransformerMixin):
             ind,
             metric=self.metric,
             reg=self.reg,
+            kernel=self.kernel
         )
 
         X_new = np.empty((X.shape[0], self.n_components))
@@ -266,7 +282,7 @@ class LocallyLinearEmbedding(BaseEstimator, TransformerMixin):
         return self.embedding_
 
 
-def barycenter_weights(X, Y, indices, metric="riemann", reg=1e-3):
+def barycenter_weights(X, Y, indices, metric="riemann", kernel=None, reg=1e-3):
     """Compute Riemannian barycenter weights of X from Y along the first axis.
 
     Estimates the weights to assign to each point in Y[indices] to recover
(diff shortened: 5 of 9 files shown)


brunaafl:
@qbarthelemy
Fyi: @sylvchev, @bruAristimunha

I'm also preparing some test scrips to add for testing the transformation.

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+from scipy.linalg import sqrtm, inv
+from numpy import iscomplexobj, real, any, isfinite
+from sklearn.base import BaseEstimator, TransformerMixin
+from pyriemann.estimation import covariances
+
+from pyriemann.utils.mean import mean_covariance
+from pyriemann.transfer import decode_domains
+
+
+def compute_ref_euclidean(data):
```
gcattan, line 12: I think this method should be private.

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+    data: numpy array.
+    It can be either raw trials with shape (n, c, t) or covariances (n, c, c)
+    Reference trials to compute the alignment matrix
+
+    Returns
+    -------
+    r_ea: numpy array with shape (c, c)
+    Reference matrix
+
+    '''
```
gcattan, line 27: You can remove this, if your method is private.

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+
+    mean = mean_covariance(data, metric='euclid')
+
+    compare = np.allclose(mean, np.identity(mean.shape[0]))
+
+    if not compare:
+        if iscomplexobj(mean):
+            print("covariance matrix problem")
```
gcattan, line 35: Here and below, I suggest to use `warns` instead of `print`.

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+
+    This transformation accept both raw signals and covariance matrices as
+    data inputs. For raw trials, you can determine the estimator using the
+    'estimtor' parameter.
+
+    .. note::
+       Using .fit() and then .transform() will give different results than
+       .fit_transform(). In fact, .fit_transform() should be applied on the
+       training dataset (target and source) and .transform() on the test
+       partition of the target dataset.
```
gcattan, line 71: I think this is the default behavior for transformer_mixin.
https://github.com/scikit-learn/scikit-learn/blob/f07e0138bfee41cd2c0a5d0251dc3fe03e6e1084/sklearn/base.py#L1043

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Parameters
+    ----------
+
+    dtype: str
+    Type of input data, it can be 'covmat' for covariance or 'raw' for raw trials
```
gcattan, line 77: I think it should be an indentation here.

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+
+    estimator: str
+    If dtype=='raw', it represents the covariance estimator
```
gcattan, line 80: And if it is "covmat" or something else?

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+        Dictionary containing the covariance matrices for each domain.
+        Order of the list is the same as 'domains'
+
+    References
+    ----------
+    .. [1] `Transfer Learning for Brain-Computer Interfaces:
+        A Euclidean Space Data Alignment Approach
+        <https://arxiv.org/abs/1808.05464>`_
+        He He and Dongrui Wu, IEEE Transactions on Biomedical Engineering, 2019
+
```
gcattan, line 98: Can you add:
```
Notes
-------
..versionadded:: 0.6.0
```

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+        return X_align
+
+    def _transform_cov(self, X, d):
+        """ Compute covariance matrices of trials in array X and save"""
+
+        cov = covariances(X, estimator=self.estimator)
+        self._cov[d] = cov
+        return cov
+
+    def fit(self, X, y_enc):
```
gcattan, line 150: `fit` method is public, it should be documented.
Something along these lines (except you accept both epochs and covariance matrices):

https://github.com/pyRiemann/pyRiemann/blob/c44b924573dc2f4b4a9efb328d18636b8baa31d5/pyriemann/tangentspace.py#L83

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+
+        _, _, domains = decode_domains(X, y_enc)
+
+        self._fit_sw(X, domains)
+
+        return self
+
+    def transform(self, X, y_enc=None):
+        """
+        Used for transforming the target data after fitting the calibration
```
gcattan, line 162: Should be also documented, as it is a public method.

gcattan (changes requested):
Hi @brunaafl !
Nice work :) IMO, I think it could be a great idea to add a tutorial too. 

Below, some minor comments on the form.

pyriemann/transfer/_euclidean.py
gcattan: Can you also add the reference to your paper?

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def compute_ref_euclidean(data):
+    '''
+    Calculate the reference matrix for computing the Euclidean alignment,
+
+    Parameters
+    ----------
+    data: numpy array.
+    It can be either raw trials with shape (n, c, t) or covariances (n, c, c)
```
VascoSch92, line 19:
```suggestion
    It can be either raw trials with shape (n, c, t) or covariances with shape (n, c, c)
```

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+            print("covariance matrix problem")
+        if iscomplexobj(sqrtm(mean)):
+            print("covariance matrix problem sqrt")
+
+        r_ea = inv(sqrtm(mean))
+
+        if iscomplexobj(r_ea):
+            print("WARNING! Covariance matrix was not SPD somehow. " +
+                  "Can be caused by running ICA-EOG rejection, if " +
+                  "not, check data!!")
```
VascoSch92, line 44:
```suggestion
            warns(
                "Covariance matrix was not SPD somehow. \n"
                 "Can be caused by running ICA-EOG rejection, if " 
                 "not, check the data!"
             )
```

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+    .. [1] `Transfer Learning for Brain-Computer Interfaces:
+        A Euclidean Space Data Alignment Approach
+        <https://arxiv.org/abs/1808.05464>`_
+        He He and Dongrui Wu, IEEE Transactions on Biomedical Engineering, 2019
+
+    """
+    def __init__(self, estimator='lwf', dtype='covmat'):
+        self.estimator = estimator
+        self.dtype = dtype
+        self._ref = list()
```
VascoSch92, line 103:
```suggestion
        self._ref = []
```
just to be coherent with `self._cov` ;-)

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+        self._cov = {}
+
+    def _compute_ref(self, X):
+        ref = compute_ref_euclidean(X)
+
+        self._ref.append(ref)
+
+    def _fit_sw(self, X, domains):
+
+        for i in range(len(np.unique(domains))):
```
VascoSch92, line 113: I think here you can just loop on` np.unique(domains)`, i.e.,
```python
for d in np.unique(domains):
   ...
```

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+            if self.dtype != 'covmat':
+                X_d = self._transform_cov(X_d, d)
+
+            self._compute_ref(X_d)
+
+    def _transform_sw(self, X, domains):
+        """Align each domain"""
+
+        X_align = []
+        for i in range(len(np.unique(domains))):
```
VascoSch92, line 128: same as above

pyriemann/transfer/_euclidean.py
brunaafl: You might be right. I was not sure if it was the case, so I put this comment. Since I was following [https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/transfer/_estimators.py#L89](https://github.com/pyRiemann/pyRiemann/blob/c44b924573dc2f4b4a9efb328d18636b8baa31d5/pyriemann/transfer/_estimators.py#L89) as a reference to my code, I put the same note.

pyriemann/transfer/_euclidean.py
```diff
@@ -0,0 +1,190 @@
...
+        Used for fitting and transforming 'train' data.
+        """
+
+        self.fit(X, y_enc)
+
+        _, _, domains = decode_domains(X, y_enc)
+
+        X_align = self._transform_sw(X, domains)
+
+        return X_align
```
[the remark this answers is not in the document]
brunaafl: I did this so I can could separate `fit_transform` as the method to be applied during the training/fitting, while `transform` during inference in the target set.

pyriemann/transfer/_euclidean.py
brunaafl: I ended up looping on the range so it would be easier to get the correspondent reference matrix, saved on a list, for each domain, although it may not be the best way to do it. I was thinking first on saving the reference matrices on dictionaries, indexed by the domain, but then I would need to set before the number of the test domain.

pyriemann/transfer/euclidean.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self.fit(X, y_enc)
+
+        _, _, domains = decode_domains(X, y_enc)
+
+        X_align = self._transform_sw(X, domains)
+
+        return X_align
+
+    def __sklearn_is_fitted__(self):
```
gcattan, line 256: Your transformer always needs fitting. I don't think you can call transform without fit first. I suggest to remove this override.

brunaafl: Yes, I agree! Supposing that you have data from different 'train' subjects (that will be used to train a classifier, for example) + calibration data from 'target' subject, and you want to use calibration data to compute reference for aligning the test data from 'target', the pipeline I was thinking was:

'Train' subjects --> `fit_transform` --> aligned 'Train'

'Calibration' --> `fit` calibration (compute reference)  or `fit_transform` if you want aligned calibration data

'Test' --> `transform` 'test' **after fitting calibration** --> aligned 'Test'

In this way, the last reference matrix computed/saved would be from calibration, and it would be used for transforming the test! I'm not sure if the way I implemented would be the best to do it, but that is what I was thinking

pyriemann/transfer/euclidean.py
```diff
@@ -0,0 +1,258 @@
...
+
+    This transformation accept both raw signals and covariance matrices as
+    data inputs. For raw trials, you can determine the estimator using the
+    'estimtor' parameter.
+
+    .. note::
+       Using .fit() and then .transform() will give different results than
+       .fit_transform(). In fact, .fit_transform() should be applied on the
+       training dataset (target and source) and .transform() on the test
+       partition of the target dataset.
```
sylvchev, line 61:
```suggestion
       If applied offline, that is to transform the training data, use .fit_transform().
       For pseudo-online or calibration setup, with training data from one domain
       (one subject) and unseen tested data, use .fit() and .transform(). You should
       ensure that there is only one domain in training data.
       
       Using .fit() and then .transform() will give different results than
       .fit_transform(). In fact, .fit_transform() should be applied on the
       training dataset (target and source) and .transform() on the test
       partition of the target dataset.
```

pyriemann/transfer/euclidean.py
```diff
@@ -0,0 +1,258 @@
+import numpy as np
+
+from warnings import warn
+
+from scipy.linalg import sqrtm, inv
+from numpy import iscomplexobj, real, any, isfinite
+from sklearn.base import BaseEstimator, TransformerMixin
+from pyriemann.estimation import covariances
```
bruAristimunha, line 8:
```suggestion
from ..estimation import covariances
```

qbarthelemy (changes requested):
Thx @brunaafl for this contribution!

Due to their very high similarity, `EuclideanAlignment(dtype="covmat")` should be simply implemented using `TLCenter(target_domain="", metric="euclid")`, ie allowing an empty str for `target_domain` and modifying `transform` for this particular case (see https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/transfer/_estimators.py#L215).

sylvchev:
Good call @qbarthelemy 
I discussed with @brunaafl , she will try to to rely on TLCenter for euclidean alignement as much as possible.

qbarthelemy:
`transform` method of class `TLCenter` can be simply generalized into:
```
    def transform(self, X, y_enc):
        """Re-center matrices to Identity.

        If ``target_domain`` is not empty, recenter matrices to this
        specified target domain.
        Else, recenter matrices of each domain to Identity.

        Parameters
        ----------
            Set of SPD matrices.
        y_enc : ndarray, shape (n_matrices,)
            Extended labels for each matrix.
            Used only when ``target_domain`` is empty.

        Returns
        -------
        X : ndarray, shape (n_matrices, n_classes)
            Set of SPD matrices with mean in the Identity.
        """
        if self.target_domain != "":
            return self.recenter_[self.target_domain].transform(X)

        _, _, domains = decode_domains(X, y_enc)

        X_rct = np.zeros_like(X)
        for d in np.unique(domains):
            idx = domains == d
            X_rct[idx] = self.recenter_[d].transform(X[idx])
        return X_rct
```

brunaafl:
Hi!

I made some changes to the TLCenter code, as @qbarthelemy  suggested.

I changed the parameter “target_domain” to be optional, allowing the same functionality as before to be maintained. 

```
    def transform(self, X, y_enc=None):
        """Re-center the data points in the target domain to Identity.

        Parameters
        ----------
            Set of SPD matrices.
        y_enc : None
            Not used, here for compatibility with sklearn API.

        Returns
        -------
        X : ndarray, shape (n_matrices, n_classes)
            Set of SPD matrices with mean in the Identity.
        """
        # Used during inference, apply recenter from specified target domain.

        # Therefore, if you specified before the target
        if self.target_domain != "":
            recenter = self.recenter_[self.target_domain].transform(X)

        # If you didn't specified, you must have fitted before on calibration
        else:
            keys = list(self.recenter_.keys())

            # Get last reference matrix (after fitting on the calibration)
            last = keys[-1]
            recenter = self.recenter_[last].transform(X)

        return recenter
```

However, if used as a transformation in a pipeline in a case where “target_domain” should be left empty, such as in an online cross-domain evaluation, you should fit the transformation to a calibration set of the target before transforming the test data. Doing that, the reference matrix computed using calibration will be the last one saved, and you will use it to recenter the test. 

With this structure, however, I’m not solving the problem of overriding the fit_transform, and applying `fit` and then `transform` will continue to be different from applying `fit_transform`, since this supposes that data passed in `transform` is from target subject. A way I imagined of solving this would be using the fact that y_enc==None in inference for the target, and not None otherwise.

Another thing I was imagining is if it was worth it to add a possibility of aligning “raw” trials as well, so it can be used in a more general way.

Please let me know what are your thoughts, and thank you all for all the comments until now!

qbarthelemy:
Thank you @brunaafl  for this improvement of `TLCenter` class!
I just updated the doc and code. Don't forget to pull before continuing to code.

It's ok for me to use the last calibrated domain if `target_domain` is not defined.
That said, it seems that this does not cover the use case of He & Wu 2019...
On the other hand, to be sure that the user makes an explicit choice, `target_domain` must not be optional.

For recentering raw signals, I think it should rather be added to MOABB, because pyRiemann is more dedicated to SPD matrices.

qbarthelemy (approved):
Thx @brunaafl !

PR #293, opened 2024-04-11
Add LLE kernel option

gabelstein:
This adds a kernel option to LLE to make it more flexible. 
Also, it allows for new parameter 'None' for n_neighbors and n_components to be more similar to e.g. KernelPCA from scikit-learn.


pyriemann/embedding.py
```diff
@@ -282,6 +298,10 @@ def barycenter_weights(X, Y, indices, metric="riemann", reg=1e-3):
         Indices of the points in Y used to compute the barycenter
     metric : {"euclid", "logeuclid", "riemann"}, default="riemann"
         Kernel metric.
+    kernel : callable, default=None
+        Kernel function to use for the embedding. If None, the canonical
+        kernel specified by the metric is used. Must be a function that
+        takes the arguments (X, Cref, metric).
```
qbarthelemy, line 304: Since function `pyriemann.utils.kernel.kernel` allows a callable `metric`,
we can simply update the parameter `metric` as:
```suggestion
    metric : string | callable, default="riemann"
        Metric of kernel used for embedding, can be:
        "euclid", "logeuclid", "riemann", or a callable function.
```
Same remark for all new `kernel` parameters.

Would that cover your use case?

qbarthelemy (changes requested):
Thanks @gabelstein for this improvement!

My major comment is that it is not necessarily necessary to add a new parameter `kernel`: extending `metric` should be enough.

pyriemann/embedding.py
gabelstein: @qbarthelemy with the new extension of the kernel module I have planned, this would not work. It will also cover distance kernel etc so specifying a metric should then only affect the distance metric.

pyriemann/embedding.py
```diff
@@ -396,18 +428,25 @@ def locally_linear_embedding(X,
 def _check_dimensions(X, Y=None, n_components=None, n_neighbors=None):
     n_matrices, n_channels, n_channels = X.shape
 
-    if not isinstance(Y, type(None)):
+    if Y is not None:
         msg = f"Dimension of matrices in data to be transformed must match " \
               f"dimension of data used for fitting. Expected " \
               f"{(n_channels, n_channels)}, got {Y.shape[1:]}."
         assert Y.shape[1:] == (n_channels, n_channels), msg
```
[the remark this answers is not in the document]
gabelstein: Can do, but it will throw an error anyways when multiplication is done. Should it still only be a warning?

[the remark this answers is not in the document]
qbarthelemy: Ok, you can raise an error.

pyriemann/embedding.py
qbarthelemy: I understand. 

But it's difficult to judge which is the best solution since I don't see the code of the future kernel module.

gabelstein: @qbarthelemy Here is the preliminary code https://github.com/gabelstein/pyRiemann/tree/new_kernels

gabelstein: In the new module it would make most sense to make ktype be able to be the callable kernel function.

qbarthelemy: In this case, I prefer to rename `kernel` in `ktype`.

pyriemann/embedding.py
```diff
@@ -340,11 +362,18 @@ def locally_linear_embedding(X,
     X : ndarray, shape (n_matrices, n_channels, n_channels)
         Set of SPD matrices.
     n_components : int, default=2
-        Dimensionality of projected space.
+        Dimensionality of projected space. If None, n_components is set to
+        n_matrices - 1.
```
qbarthelemy, line 366:
```suggestion
        Dimensionality of projected space.
        If None, ``n_components`` is set to ``n_matrices - 1``.
```

pyriemann/embedding.py
```diff
@@ -340,11 +362,18 @@ def locally_linear_embedding(X,
...
     n_neighbors : int, default=5
-        Number of neighbors for reconstruction of each point.
+        Number of neighbors for reconstruction of each point. If None, all
+        available points are used. If n_neighbors > n_matrices, n_neighbors
+        is set to n_matrices - 1.
```
qbarthelemy, line 370:
```suggestion
        Number of neighbors for reconstruction of each point.
        If None, all available matrices are used.
        If ``n_neighbors > n_matrices``, ``n_neighbors`` is set to
        ``n_matrices - 1``.
```

pyriemann/embedding.py
gabelstein: Wouldn't that be confusing if the ktype in the kernel module means something else than the ktype in the embedding? 
In the kernel module, ktype is either a string or a callable that does something with the metric (string).
Here, we need a kernel method to be passed to LLE.

qbarthelemy: You are right, let's keep `kernel`.

pyriemann/embedding.py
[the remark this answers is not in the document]
gabelstein: So it is fine as it is?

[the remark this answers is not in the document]
qbarthelemy: No, can you replace asserts by raise ValueError?

qbarthelemy (approved):

qbarthelemy:
Thx @gabelstein !

PR #294, opened 2024-04-23
Update testing.yml

gcattan:
Python 3.8 and 3.9 are not available on macos-latest.
See this comment https://github.com/actions/setup-python/issues/850#issuecomment-2072657610

@qbarthelemy FYI

PR #295, opened 2024-04-24
Fix LLE warning mistake

gabelstein:
Stupid error, obviously always printed     `n_neighbors = n_matrices - 1` in the warning.

Also, this warning clogs my console extremely on large comparisons. Maybe we could just not proint a warning @qbarthelemy ? 

Or introduce some variable that suppresses it? But this would be more of a global thing, I guess.

### DIFF
--- a/pyriemann/embedding.py
+++ b/pyriemann/embedding.py
@@ -445,9 +445,9 @@ def _check_dimensions(X, Y=None, n_components=None, n_neighbors=None):
     if n_neighbors is None:
         n_neighbors = n_matrices - 1
     elif n_matrices <= n_neighbors:
-        n_neighbors = n_matrices - 1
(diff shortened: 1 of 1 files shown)


qbarthelemy:
>Also, this warning clogs my console extremely on large comparisons. Maybe we could just not proint a warning

Maybe, check on `n_neighbors` could be carried out only during `fit` functions, not during `transform`.

qbarthelemy:
Can you rebase your branch on the last master, to have the corrections of #294 ?

gabelstein:
> Maybe, check on `n_neighbors` could be carried out only during `fit` functions, not during `transform`.

It is already only performed during fit. But warnings can be easily suppressed globally, so it's ok as it is I guess.

PR #296, opened 2024-05-14
Improve clustering module

qbarthelemy:
This PR:

- corrects `PotatoField`, raising is an error when ` partial_fit ` is called without having called `fit`
```AttributeError: 'PotatoField' object has no attribute '_potatoes' ```
Corrected adding an explicit message.

- corrects `Kmeans`: `mdm.predict` in `_fit_single` raises an error when the number of classes is not equal to the number of clusters: `IndexError: index 2 is out of bounds for axis 0 with size 2`
Corrected modifying initialization of `mdm.classes_ `.

- completes doc and tests of `KmeansPerClassTransform`.

- simplifies code for `partial_fit` of `Potato`, merging batch and online updates.
This requires that the `mean` function can be called on a single matrix, which was not the case for `mean_alm`. Corrected.

- completes tests for attributes of different classes.

- completes documentation.

### DIFF
--- a/pyriemann/clustering.py
+++ b/pyriemann/clustering.py
@@ -64,7 +64,8 @@ def _fit_single(X, y=None, n_clusters=2, init="random", random_state=None,
 class Kmeans(BaseEstimator, ClassifierMixin, ClusterMixin, TransformerMixin):
     """Clustering by k-means with SPD matrices as inputs.
 
-    Find clusters that minimize the sum of squared distance to their centroids.
+    The k-means is a clustering method used to find clusters that minimize the
+    sum of squared distances to their centroids.
     This is a direct implementation of the k-means algorithm with a Riemannian
     metric [1]_.
 
@@ -324,8 +325,9 @@ class Potato(BaseEstimator, TransformerMixin, ClassifierMixin):
     """Artefact detection with the Riemannian Potato.
 
     The Riemannian Potato [1]_ is a clustering method used to detect artifact
-    in EEG signals. The algorithm iteratively estimates the centroid of clean
-    signal by rejecting every trial that is too far from it.
+    in multichannel signals. Processing SPD matrices,
+    the algorithm iteratively estimates the centroid of clean
+    matrices by rejecting every matrix that is too far from it.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pyriemann/clustering.py [hunk 1]
```diff
@@ -37,20 +37,17 @@ def _init_centroids(X, n_clusters, init, random_state, x_squared_norms):
...
-def _fit_single(X, y=None, n_clusters=2, init='random', random_state=None,
-                metric='riemann', max_iter=100, tol=1e-4, n_jobs=1):
+def _fit_single(X, y=None, n_clusters=2, init="random", random_state=None,
+                metric="riemann", max_iter=100, tol=1e-4, n_jobs=1):
     """helper to fit a single run of centroid."""
     # init random state if provided
     mdm = MDM(metric=metric, n_jobs=n_jobs)
     mdm.metric_mean, mdm.metric_dist = check_metric(metric)
-    squared_norms = [np.linalg.norm(x, ord='fro')**2 for x in X]
+    squared_norms = [np.linalg.norm(x, ord="fro")**2 for x in X]
```
agramfort, line 46: you could use: `np.linalg.norm(A, ord="fro", axis=(1, 2))**2`

pyriemann/clustering.py
```diff
@@ -421,34 +451,31 @@ def partial_fit(self, X, y=None, alpha=0.1):
...
+            return self
+
+        Xm = mean_covariance(X[(y == self.pos_label)], metric=self.metric)
+        self._mdm.covmeans_[0] = geodesic(
+            self._mdm.covmeans_[0], Xm, alpha, metric=self.metric)
+
+        d = np.squeeze(np.log(self._mdm.transform(Xm[np.newaxis, ...])))
+        self._mean = (1 - alpha) * self._mean + alpha * d
+        self._std = np.sqrt(
+            (1 - alpha) * self._std**2 + alpha * (d - self._mean)**2)
```
agramfort, line 478:
```suggestion
        self._mdm.covmeans_[0] = geodesic(
            self._mdm.covmeans_[0], Xm, alpha, metric=self.metric
        )

        d = np.squeeze(np.log(self._mdm.transform(Xm[np.newaxis, ...])))
        self._mean = (1 - alpha) * self._mean + alpha * d
        self._std = np.sqrt(
            (1 - alpha) * self._std**2 + alpha * (d - self._mean)**2
        )
```

pyriemann/clustering.py
qbarthelemy: Yes, good idea!

agramfort (approved):

PR #298, opened 2024-05-25
MeanField with TangentSpace

gcattan:
A proposal about mixing TS with MeanField, for MI application mainly.
@qbarthelemy @toncho11


qbarthelemy (commented):
I'm sorry @gcattan , but we only add algorithms in pyRiemann that have been validated on real datasets and published.

gcattan:
Oki

PR #299, opened 2024-05-27
Enhance ajd_pham to process HPD matrices

qbarthelemy:
This PR enhances `ajd_pham` to process HPD matrices,
following code https://github.com/Marco-Congedo/Diagonalizations.jl/blob/master/src/optim/LogLike.jl#L46
Note: `tau` in pyRiemann is the transpose of `Gamma` in Diagonalizations.jl

This is the continuation of  #238. Related to #204.

A consequence is that `mean_ale` supports HPD matrices.


agramfort:
let me know when it's good to go.

agramfort (approved):

PR #302, opened 2024-06-05
Improve embedding module

qbarthelemy:


pyriemann/embedding.py
```diff
@@ -338,20 +343,22 @@ def barycenter_weights(X, Y, indices, metric="riemann", kernel=None, reg=1e-3):
...
     """Perform a Locally Linear Embedding (LLE) of SPD matrices.
 
     As proposed in [1]_, Locally Linear Embedding (LLE) is a non-linear,
     neighborhood-preserving dimensionality reduction algorithm which consists
-    of three main steps. For each point xi,
+    of three main steps. For each point X[i],
 
-    1.  find its k nearest neighbors KNN(xi),
+    1.  find its k nearest neighbors KNN(X[i]),
     2.  calculate the best reconstruction of xi based on its
```
gcattan, line 362:
```suggestion
    2.  calculate the best reconstruction of X[i] based on its
```

gcattan (approved):

pyriemann/embedding.py
```diff
@@ -51,18 +56,18 @@ def __init__(self, n_components=2, metric="riemann", eps=None):
 
     def _get_affinity_matrix(self, X, eps):
 
-        # make matrix with pairwise distances between points
+        # make matrix with pairwise distances between matrices
         distmatrix = pairwise_distance(X, metric=self.metric)
 
         # determine which scale for the gaussian kernel
-        if self.eps is None:
+        if eps is None:
```
gcattan, line 63: There is no need to pass `eps` as a parameter, as it is already defined as class member.

qbarthelemy: Yes, you are right.
Function `_get_affinity_matrix` mixes the treatment of the class member `self.eps` and the argument `eps`.
The idea is to harmonize this: as I'm not a fan of overwriting the `self.eps` parameter, I prefer to use `eps` everywhere in `_get_affinity_matrix`. What do you think?

gcattan: Ok. I see your point. I think it deserves a comment. May be something like this:

"We do not want to overwrite the `self.eps` attribute, to keep consistency in the output of the `get_params` method defined by sklearn. Therefore `self.eps` is passed by copy in `_get_affinity_matrix` parameters"

qbarthelemy: I decided to simplify the code, by removing the `eps` from the `_get_affinity_matrix` parameters.

gcattan (approved):

PR #306, opened 2024-06-13
Remote sensing examples

AmmarMian:
A basic example of using pyRiemann to cluster both Hyperspectral images (SPD matrices) or PolSAR images (HPD matrices).
Any input about the coding style is welcomed.

I had to rebase to get all the changes from pyRiemann master branch so I have lots of commits. Please let me know if you'd prefer I make a cleaner version starting from current pyRiemann.


examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
...
+class SlidingWindowVectorize(BaseEstimator, TransformerMixin):
+    """Slinding window over an image of shape (n_rows, n_columns, n_features).
+
+    Parameters
+    ----------
+    window_size : int
+        Size of the window.
+
+    overlap : int
+        Overlap between windows. Default is 0.
```
qbarthelemy, line 83: In pyRiemann, we don't jump a line between parameters.
```suggestion
        Size of the window.
    overlap : int, default=0
        Overlap between windows.
```

examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
...
+
+    Parameters
+    ----------
+    height : int
+        Height of the original image.
+
+    width : int
+        Width of the original image.
+
+    window_size : int
```
qbarthelemy, line 129:
```suggestion
        Height of the original image.
    width : int
        Width of the original image.
    window_size : int
```

examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
...
+def pca_image(image: ArrayLike, nb_components: int):
+    """ A function that centers data and applies PCA on an image.
+
+    Parameters
+    ----------
+        image : ArrayLike
+            An image of shape (n_rows, n_columns, n_features).
+
+        nb_components : int
+            Number of components to keep.
```
qbarthelemy, line 172: Shape of array is indicated in the first line.
```suggestion
    image : ndarray, shape (n_rows, n_columns, n_features) 
        An image.
    n_components : int
        Number of principal components to keep.
```

examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
...
+        assert overlap >= 0, 'Overlap must be positive.'
+        assert overlap <= window_size//2,\
+                'Overlap must be smaller or equal than int(window_size/2).'
+        self.window_size = window_size
+        self.overlap = overlap
+
+    def fit(self, X: ArrayLike, y = None):
+        return self
+
+    def transform(self, X: ArrayLike):
```
qbarthelemy, line 97: Can you add the docstring for this function?

examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
+"""
+====================================================================
+Remote sensing: Clustering on hyperspectral/sar images with Riemannian
+geometry
+====================================================================
+
+This example compares clustering pipelines based on covariance matrices for
+hyperspectral/sar data custering.
```
qbarthelemy, line 8:
```suggestion
Clustering on hyperspectral and SAR images with Riemannian geometry
====================================================================

This example compares pipelines based on covariance matrices for
hyperspectral and synthetic-aperture radar (SAR) images custering.
```

examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
...
+        os.makedirs(data_path, exist_ok=True)
+    if not os.path.exists(os.path.join(data_path, filename)):
+        print('Downloading UAVSAR dataset...')
+        urllib.request.urlretrieve(url, os.path.join(data_path, filename))
+        print('Done.')
+    else:
+        print('UAVSAR dataset already downloaded.')
+
+
+if __name__ == "__main__":
```
qbarthelemy, line 297: `if __name__ == "__main__"` is not necessary. You can use a #-line to separate class definition from the actual example execution.
```suggestion
###############################################################################
```
Don't forget to remove indent for the code below.

examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
...
+from pyriemann.clustering import Kmeans
+from scipy.io import loadmat
+from typing import Tuple, Dict
+
+from numpy.lib.stride_tricks import sliding_window_view
+from math import ceil 
+
+from sklearn.base import BaseEstimator, TransformerMixin
+from numpy.typing import ArrayLike
+
```
qbarthelemy, line 31: You can use a #-line to separate imports from class definition.
```suggestion

###############################################################################
```

qbarthelemy (changes requested):
Thx @AmmarMian for this valuable contribution!

Can you fix flake8 errors?
https://github.com/pyRiemann/pyRiemann/pull/306/checks#step:6:8

Like `SSVEP` or `motor-imagery` folders, can you create a subfolder `helpers` to put all functions that can be re-use by (future) other examples on remote sensing?
(`download_salinas`, `read_salinas`, `dowload_uavsar`)

And don't forget to complete the whatsnew file.

examples/remote-sensing/plot_clustering.py
AmmarMian: features is quite broad depending on the application. It can be:
1.  Spectral dimension if several radar were used (very few datasets have this in radar because it require different sensors)
2. Polarimetric information (several modes of polarisation of emitted waves exist and often are used)
3. Interferometric information (sometimes there are two or more acquisition at the same time but with a slightly different looking angle)
4. Time dimension when several acquisitions are done at different times
5. Also depending on the direction the satellite saw the scene there are usually two modes (ascending: satellite going to north direction, descending the reverse). The same satellite can usually do both at slightly different times due to how the orbits work 

In all these cases, the covariance descriptor is useful for classification and the framework isn't so different for the basic processings: computing covariance over spatial coordinates on a sliding window approach (the goal is to classify pixels over a feature dimension. Depending on the used data, the goals and result of classification may differ).

Fo a basics on SAR, see for example: https://elib.dlr.de/82313/1/SAR-Tutorial-March-2013.pdf

Also besides SAR, there are also non-radar optical images (such as the hyper spectral case in the example), where `features`  refers to often spectral dimensions but there can also be temporal.

To conclude, the applications in remote sensing are so large that the data dimensionalities differ vastly: it can go from a simple (n_rows, n_columns) to (n_rows, n_columns, n_looking_angles, n_polarisations, n_spectral_bands, 2 (asc and desc), n_times) to the most extreme case.
Depending on the application, one can consider what they want as feature and the rest can be seen as batch dimensions.

examples/remote-sensing/plot_clustering.py
```diff
@@ -0,0 +1,373 @@
...
+
+class RemoveMeanImage(BaseEstimator, TransformerMixin):
+    """ Remove the global mean of an three-dimensional image.
+    """
+    def __init__(self):
```
VascoSch92, line 36: Are you re-writing the init because you want to override the init of the parent classes? If it is not the case, you can just delete it ;)

AmmarMian:
@qbarthelemy Can you check now ?
Normally I resolved flake8 errors, added your suggestions and references.

I also split SAR and hyper spectral in two files because they are two types of images. Next I'll add an example of change detection rather than clustering so that will also be a new file.

Ammar

qbarthelemy:
Thank you @AmmarMian for all this work!
I made some small modifs, so don't forget to pull before committing and pushing.

My concern is the building time of the documentation: it went from 20 / 30 min to almost 1hour
(see https://github.com/pyRiemann/pyRiemann/actions/workflows/deploy_ghpages.yml).
Is it possible to process only a sub-image of the SAR image?

Moreover, currently, you apply a raw downsampling on images, `data = data[::7, ::7]`, but it generates spectral aliasing.
It is better to apply a decimation, ie an anti-aliasing low-pass filtering followed by the downsampling.
We could use `resize` function from Pillow, but it adds a dependencies to docs requirement. Ok for you @agramfort ?

agramfort:
I would say it's ok to add Pillow to the doc build reqs

qbarthelemy:
Thanks for all the great advice about radar image processing!
Building time of documentation is now OK.

examples/remote-sensing/helpers/processing_helpers.py
```diff
@@ -0,0 +1,203 @@
...
+            "Overlap must be smaller or equal than int(window_size/2)."
+        self.height = height
+        self.width = width
+        self.overlap = overlap
+        self.window_size = window_size
+
+    def fit(self, X: ArrayLike, y=None):
+        return self
+
+    def transform(self, X: ArrayLike):
```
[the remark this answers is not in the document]
AmmarMian: Yes kinda, but it's very specific to one-dimensional output (1 label per pixel), whilst in my view we could use sliding windows to do any number of processing (for example just stopping at covariances: 1 covariance per window which is sometimes useful in PolSAR or InSAR applications, in which case recasting as one covariance per pixel would be a bit different).

I guess we could generalize the process to any number of dimensions (and we could apply that to sliding windows as well). But I didn't want to add more complexity to those examples since we only want some labels at the end. 

Also if I remember correctly, but might be wrong, I wanted to add it to the pipeline but the way Kmeans works doesn't allow to pass a `TransformerMixin` after to call `fit_transform` with this as the end of the pipeline.

examples/remote-sensing/helpers/processing_helpers.py
```diff
@@ -0,0 +1,203 @@
...
+    # check pixels are centered
+    assert (np.abs(np.mean(X, axis=0)) < 1e-8).all()
+
+    # apply PCA
+    SCM = (1/len(X))*X.conj().T@X
+    d, Q = la.eigh(SCM)
+    reverse_idx = np.arange(len(d)-1, -1, step=-1)
+    Q = Q[:, reverse_idx]
+    Q = Q[:, :n_components]
+    image = image@Q
```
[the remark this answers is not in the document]
AmmarMian: I believe @antoinecollas used this because it's faster to apply over the multidimensional image.

In my case, after the sliding windows, the problem is that there are many redundant data because of the overlap. And even if there weren't the way the data is shaped is not exactly `(n_features, n_samples)'. One could always do something like vectorising image at first then pcq then shaping it back to an image with less features then sliding windows (it would be tricky to a sliding windows over vectorised version where we converted indexes of rows, columns to a 1D index) but since Antoine already coded the PCA over images I simply used that.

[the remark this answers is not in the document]
antoinecollas: I think it's better to use sklearn. Indeed, it means doing reshape -> PCA -> reshape.

examples/remote-sensing/helpers/processing_helpers.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def __init__(self, n_components: int):
+        assert n_components > 0, "Number of components must be positive."
+        self.n_components = n_components
+
+    def fit(self, X: ArrayLike, y=None):
+        return self
+
+    def transform(self, X: ArrayLike):
+        """TODO.
```
antoinecollas, line 83: Some TODO left 👀

examples/remote-sensing/helpers/processing_helpers.py
```diff
@@ -0,0 +1,203 @@
...
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_rows, n_columns, n_features)
+            Input image
+
+        Returns
+        -------
+        Xnew : ndarray, shape TODO
```
antoinecollas, line 92: Some TODO left 👀

antoinecollas (commented):
Well done @AmmarMian 👍🏻  ! I really like that you can do everything with a sklearn Pipeline. It opens many perspectives: better evaluation, comparison of estimators, ...

Overall, you should replace `pca_image` by `sklearn.decomposition.PCA` and put it directly  in `PCAImage`.

AmmarMian:
Ok so from the input of @antoinecollas, I removed `pca_image`and used `sklearn.decomposiiton.PCA` with a reshape. I also added the doctoring for `transform` of `LabelsToImage`.

examples/remote-sensing/helpers/processing_helpers.py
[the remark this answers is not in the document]
qbarthelemy: I am sorry, I just remembered that scikit-learn doesn't support complex data.
So, it is better to keep this implementationif you want to process complex images.

examples/remote-sensing/helpers/processing_helpers.py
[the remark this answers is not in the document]
qbarthelemy: Ok. Let's leave generalization aside.

However, I suggest to move `LabelsToImage.transform()` into a new function `SlidingWindowVectorize.inverse_predict()`.

In this way, `LabelsToImage` can be deleted and we only use one class to process data.

Current code:
```
labels_pred = LabelsToImage(height, width, window_size).fit_transform(
    pipeline.named_steps["kmeans"].labels_
)
results[pipeline_name] = labels_pred
```
can be replaced by
```
preds = pipeline.named_steps["kmeans"].labels_
results[pipeline_name] = pipeline.named_steps["sliding_window"].inverse_predict(preds)
```

For me, we can merge the branch now, but I think it's worth finalizing the image processing classes (code and documentation), in order to quickly be able to add them to the pyRiemann API.

examples/remote-sensing/helpers/processing_helpers.py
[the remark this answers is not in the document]
AmmarMian: Ok So I propose to use `sklearn.decomposition.PCA` if data is real and `pca_image` if complex. Would that work?

examples/remote-sensing/helpers/processing_helpers.py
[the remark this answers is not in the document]
AmmarMian: Ok I change that in the way you propose. It is indeed simpler!

examples/remote-sensing/helpers/processing_helpers.py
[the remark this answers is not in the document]
antoinecollas: Do you want to deal with complex-valued data? In this case, you'll also have problems with every clustering/classifier from sklearn. Does Pyriemann handle complex-valued data well?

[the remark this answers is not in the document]
qbarthelemy: Yes, pyRiemann can process complex-valued data, see #204

For the PCA code, it's up to you whether you want to process complex radar images.

> So I propose to use `sklearn.decomposition.PCA` if data is real and `pca_image` if complex. Would that work?

Yes, you can do that.

AmmarMian:
Thanks for the changes @qbarthelemy. Let me know if anything more is needed before merging!

qbarthelemy:
I made some simplifications on PCA.

My last question is about `RemoveMeanImage`: this mean removal seems redundant with the centering applied into `PCAImage`.
If so, we could delete `RemoveMeanImage`.

AmmarMian:
Indeed you are quite right, I used stuff I had lying around but in this case no need indeed ! We can remove this from the pipeline. Let me do that now and commit again.

PR #310, opened 2024-06-26
Update pyRiemann from Python 3.8 - 3.10 to 3.9 - 3.11

qbarthelemy:


agramfort:
need help?

gcattan:
Probably ubuntu-latest in your the yaml file is still on python 3.8. 
But looks like the error is something about your joblib.
We can quickly check if this is the case by adding it to your requirement file.

qbarthelemy:
I have updated yaml, but doc build is still KO. 
So, yes, I need help. @agramfort @gcattan 

`joblib` raises an error, because `pyriemann` is not actually installed.

gcattan:
Hm. Actually it remembers me something. Can you try to update the sphinx action to `josh146/sphinx-action@master`

qbarthelemy:
Wow, build is now OK!
@gcattan thank you for your help!

Issue #319, opened 2024-08-24
Add TSA to transfer learning module


qbarthelemy:
In the transfer learning module, add Tangent Space Alignement (TSA), see
https://www.frontiersin.org/articles/10.3389/fnhum.2022.1049985/pdf

qbarthelemy:
Code shared by first author
```
import numpy as np
from sklearn.base import BaseEstimator, TransformerMixin
from sklearn.decomposition import PCA
from pyriemann.estimation import Covariances
from pyriemann.transfer import decode_domains
from sklearn.preprocessing import LabelEncoder

class TSAxPCA(BaseEstimator, TransformerMixin):
    def __init__(self, target_domain, eVar=0.999, n_clusters=3, n_components=1):
        self.target_domain = target_domain
        self.eVar = eVar
        self.source_means = None
        self.target_means = None
        self.u = None
        self.vh = None

        self.incomplete_source = None
        self.incomplete_target = None

        self.n_clusters = n_clusters
        self.n_components = n_components

        self.name = "TSAxPCA"

    def fit(self, X, y_enc):
        X, y, domains = decode_domains(X, y_enc)
        X_source = X[domains != self.target_domain]
        y_source = y[domains != self.target_domain]
        X_target = X[domains == self.target_domain]
        y_target = y[domains == self.target_domain]
        self.fit_source(X_source, y_source)
        self.fit_target(X_target, y_target)

    def fit_source(self, X, y):
        y = LabelEncoder().fit_transform(y=y)
        self.incomplete_source = False
        self.source_pcas = [PCA(n_components=(self.n_components if self.n_components != "max" else X.shape[1])) for i in range(len(np.unique(y)))]
        self.source_means = np.array([np.ndarray.mean(X[y == i, :], axis=0) for i in np.unique(y)])
        for i in np.unique(y):
            Xi = X[y == i]
            ni = len(Xi)
            if ni < self.n_clusters:
                self.incomplete_source = True
                break
            self.source_pcas[i].fit(Xi)
            X_pca = self.source_pcas[i].transform(Xi)
            for j in range(self.n_components):
                args = np.argsort(X_pca[:, j], axis=0)
                self.source_means = np.vstack([self.source_means,
                                               [np.ndarray.mean(Xi[args[round(k*ni/self.n_clusters):round((k+1)*ni/self.n_clusters)]], axis=0) for k in range(self.n_clusters)]])
        if self.target_means is not None:
[...]

PR #320, opened 2024-08-24
Add Tangent Space Alignment

qbarthelemy:
Deep refacto of code shared in #319.


pyriemann/transfer/_estimators.py
```diff
@@ -592,43 +601,474 @@ def fit_transform(self, X, y_enc, sample_weight=None):
...
+        return X_new
+
+
+###############################################################################
+
+
+class TlTsCenter(BaseEstimator, TransformerMixin):
+    """Center vectors in tangent space for transfer learning.
+
+    Center the vectors from each domain to zero on tangent space.
```
plcrodrigues, line 628: To the origin of tangent space

pyriemann/transfer/_estimators.py
```diff
@@ -592,43 +601,474 @@ def fit_transform(self, X, y_enc, sample_weight=None):
...
+
+        X_new = np.zeros_like(X)
+        for d in np.unique(domains):
+            idx = domains == d
+            X_new[idx] = X[idx] - self.centers_[d]
+        return X_new
+
+
+class TlTsNormalize(BaseEstimator, TransformerMixin):
+    """Normalize vectors in tangent space for transfer learning.
```
plcrodrigues, line 729: Give more precision on what you mean by normalizing. Maybe say z-score ? Or explicitly say that you put mean zero and std to one.

plcrodrigues (commented):
Thanks @qbarthelemy for this new feature on `pyriemann` (and sorry for the delay for the review).

The code and implementation of TSA looks good to me. Did you use the same code as the one provided by Alexandre Bleuzé? Maybe add a comment about it on the docstring?

I must admit that I would have preferred to keep the names of the classes with `TL` in caps at the beginning. And I find it a bit confusing to add `Ts` and `Ma` on the name of the class. Would it make sense (and doable) to always use the `TLCenter` class and have an option to say whether we're doing stuff on the manifold or the tangent space? (same thing for stretch and rotate). This would allow people to easily switch between operations and even create pipelines that start on manifold and finish on tangent space.

qbarthelemy:
Thanks @plcrodrigues for your feedback. It's a good idea not to duplicate classes.

In the last version, there is a single class `TlCenter` which makes the centering in the manifold if the dimension of input is 3 (like `TLCenter`), and in the tangent space if dimension is 2 (like `TlTsCenter`). Same for `TlScale`, merging `TLStretch` and `TlTsNormalize`; and for `TlRotate`, merging `TLRotate` and `TlTsRotate`.

plcrodrigues (commented):
I think it would be better to force the user to **explicitly** say whether he wants to do the centering (and stretching and rotation) on the manifold or the tangent space. Of course, the `fit` function can detect what kind of data is given depending of its dimensions, but I find the current implementation a bit too implicit and it could be annoying to debug or explain the code later. Why not instantiate the classes saying where the transformations are done? 

Also, I still prefere calling things as `TLCenter` instead of `TlCenter`, etc. Since it is an acronym, it makes more sense to me to keep things as caps lock.

qbarthelemy:
Ok for me to keep prefix `TL`, instead of `Tl`.

sylvchev (approved):
Thanks @qbarthelemy for your implementation.
I'm aligned with your point of view to implicitly use the same class to handle the elements on the manifold or on the tangent space. It is the responsibility of the user to ensure that the pipelines are consistent, and we could provide some examples. I think it shows the link between the transformations, which is a good thing IMHO.

plcrodrigues (approved):
Looks good to me now! :-)

plcrodrigues:
@agramfort could you squash and merge?

sylvchev:
Thanks @qbarthelemy !

PR #322, opened 2024-09-10
Add Log-Cholesky and Cholesky metrics

gabelstein:
Implementation of full framework of log-cholesky metric from [Lin (2019)](https://arxiv.org/pdf/1908.09326).
Kernel is unpublished, so maybe we want to cut it out until it is published? Or just keep it in there with a note to the paper in progress.
I also extended it to work with HPD matrices which the original paper does not include. Not sure if that needs a disclaimer.

The logchol distance that was previously implemented does not directly relate to this paper but was introduced in a [different context](https://doi.org/10.1214/09-AOAS249) and should much rather be termed the cholesky distance.
Not sure if that need a deprecation first.
It would be quite simple to extend the simple cholesky distance to a full RIemannian metric constructing it as a pullback metric using chol as the pullback from the space of lower triangular matrices with positive diagonals with the euclidean metric. But it would roughly be about as useful as the Euclidean metric for SPD matrices, so probably not worth the struggle.

I hope I extended all the test and docs well, but someone should thoroughly check please.


qbarthelemy (changes requested):
Thanks for spotting this confusion in the Cholesky distance implementation!
Don't forget to complete the whatsnew file.

Regarding the log-Cholesky kernel, I am not in favor of adding unpublished methods into pyRiemann.

gabelstein:
@qbarthelemy Alright, I will remove the kernel. 
I personally would prefer to always write out the name "cholesky" as opposed to using "chol".
Some users might not intuitively know what logchol means.
But my heart does not cling to the function names, so if you insist we can call it logchol.

pyriemann/utils/mean.py
```diff
@@ -275,6 +275,59 @@ def mean_kullback_sym(X=None, sample_weight=None, covmats=None):
...
+    sample_weight = check_weights(sample_weight, n_matrices)
+
+    X_chol = np.linalg.cholesky(X)
+
+    tr0, tr1 = np.tril_indices(n_channels, -1)
+    diag0, diag1 = np.diag_indices(n_channels)
+
+    mean = np.zeros(X.shape[-2:], dtype=X.dtype)
+
+    mean[..., tr0, tr1] = np.average(X_chol[..., tr0, tr1], axis=0,
```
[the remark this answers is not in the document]
gabelstein: Yes, you are right.

qbarthelemy:
To avoid long function names, `*_logeuclidean` was shortened to `*_logeuclid`. 
Similarly, `*_logcholesky` is too long, that's why I proposed to shorten it to `*_logchol`.
I don't see what intermediate version to find.

doc/api.rst
```diff
@@ -296,6 +300,7 @@ Kernels
 
     kernel
     kernel_euclid
+    kernel_logcholesky
```
qbarthelemy, line 303:
```suggestion
```

pyriemann/utils/mean.py
```diff
@@ -275,6 +275,60 @@ def mean_kullback_sym(X=None, sample_weight=None, covmats=None):
...
+                                     weights=sample_weight)
+    mean[diag0, diag1] = np.exp(
+        np.average(np.log(X_chol[..., diag0, diag1]),
+                   axis=0, weights=sample_weight))
+    return mean @ mean.conj().swapaxes(-2, -1)
```
qbarthelemy, line 329:
```suggestion
    mean[tr0, tr1] = np.average(
        X_chol[:, tr0, tr1],
        axis=0,
        weights=sample_weight,
    )
    mean[diag0, diag1] = np.exp(np.average(
        np.log(X_chol[:, diag0, diag1]),
        axis=0,
        weights=sample_weight,
    ))
    return mean @ mean.conj().T
```

qbarthelemy:
@gabelstein , can you check if the latest version of the code suits you?

gabelstein:
@qbarthelemy very good, thank you!

qbarthelemy:
Are there any specific properties we could check for Cholesky or log-Cholesky metrics?

qbarthelemy (approved):

qbarthelemy:
Are there any specific properties we could check for Cholesky or log-Cholesky metrics?
Or, can I merge? @gabelstein

gabelstein:
On vacation right now. Can check next week.Am 27.09.2024 19:55 schrieb Quentin Barthélemy ***@***.***>:
Are there any specific properties we could check for Cholesky or log-Cholesky metrics?
Or, can I merge? @gabelstein

qbarthelemy:
I merge. We will make another pull request to complete the tests if necessary.

PR #323, opened 2024-09-25
Add an example of fNIRS data using HybridBlocks

timnaher:
Hi,

We recently applied the package to classify fNIRS mental imagery data and found that Riemannian-based models significantly outperformed traditional fNIRS methods relying on single-channel statistics. To help others in the fNIRS field, I have added an example for classifying fNIRS data.

In our preprint, we introduce a modified version of BlockCovariances, allowing each block to represent a distinct SPD matrix—whether a covariance or kernel. Additionally, we implemented separate regularization for each block before combining them into a block-diagonal matrix.

You can check out the preprint here:
https://www.biorxiv.org/content/10.1101/2024.09.06.611347v1

All the best, Tim


examples/fNIRS/classify_fNIRS.py
```diff
@@ -0,0 +1,348 @@
...
+            start = end
+        return result
+
+###############################################################################
+# Load data
+# ---------
+
+print("Loading preprocessed fNIRS data")
+# Adjust the path to your data
+data_path = './data'
```
agramfort, line 263: I doubt this will work as our doc is rendered automatically and data should not be commited. You need the data to be fetched from some website on run.

examples/fNIRS/classify_fNIRS.py
```diff
@@ -0,0 +1,348 @@
...
+# which can improve classification performance.
+
+###############################################################################
+# References
+# ----------
+# .. [1] `Riemannian Geometry for the classification of brain states with fNIRS
+#    <https://www.biorxiv.org/content/10.1101/2024.09.06.611347v1>`_
+#    T. Näher, L. Bastian, A. Vorreuther, P. Fries, R. Goebel, B. Sorger.
+
+#%%
```
agramfort, line 348: you can remove this line

agramfort (commented):
can you make it so that the data is fetchable so one can play with the example? 

thx a lot @timnaher !

timnaher:
Thank you for the feedback!

I have made the following updates as requested:
- Removed the unnecessary line as per your suggestion.
- Hosted the example data on Zenodo, and updated the script to automatically download the data at runtime

examples/fNIRS/classify_fNIRS.py
```diff
@@ -0,0 +1,364 @@
...
+        ]
+
+        covariance_estimators = [
+            'corr', 'cov', 'hub', 'lwf', 'mcd', 'oas', 'sch', 'scm', 'stu', 'tyl'
+        ]
+
+        M_matrices = []
+
+
+        for i in range(n_samples):
```
agramfort, line 196: I am a bit concerned by this for loop. Our pairwise_kernels functions are vectorized. Technically what you do here boils down to concatenating the tangent space features of each block. What you could do is used a sklearn ColumnTransformer where you pass the list of block indices to the ColumnTransformer and then you can apply a pipeline to each subgroup of signals. You could look at the design of https://coffeine-labs.github.io/coffeine/ 

Also have you compared the performance if you don't bother about the blocks? basically you compute the full matrix? which you can then decide to set to zero in the offdiagonal blocks.

timnaher: Thanks for the feedback! I don't know if you're referring to classification performance or computational efficiency in the second part of your comment.

If it's about classification performance: Yes, we have compared the block-based approach to a full matrix approach, and the block version consistently performed better in our tests. We have also tried adding a "background matrix" to the zero elements of the off-diagonal matrix. This allows for even another kernel function for the HbO and HbR interactions but comes of course at the cost of higher dimensionality of the manifold...

If it's about computational efficiency: No, we haven't specifically optimized for that yet, but I agree there's room for improvement in the code. I'll definitely look into using a ColumnTransformer and review the example from the Coffeine project. Thanks for the suggestion, I will implement this in the coming days.

examples/fNIRS/classify_fNIRS.py
```diff
@@ -0,0 +1,364 @@
...
+        start = 0
+        for m in matrices:
+            end = start + m.shape[0]
+            result[start:end, start:end] = m
+            start = end
+        return result
+
+###############################################################################
+# Download example data
+# ---------
```
qbarthelemy, line 256:
```suggestion
# Download example data
# ---------------------
```

examples/fNIRS/classify_fNIRS.py
```diff
@@ -0,0 +1,364 @@
...
+# values for our fNIRS classification of mental imagery. By using block diagonal matrices
+# for HbO and HbR signals, we can tune our classifier to HbO and HbR signals separately,
+# which can improve classification performance.
+
+###############################################################################
+# References
+# ----------
+# .. [1] `Riemannian Geometry for the classification of brain states with fNIRS
+#    <https://www.biorxiv.org/content/10.1101/2024.09.06.611347v1>`_
+#    T. Näher, L. Bastian, A. Vorreuther, P. Fries, R. Goebel, B. Sorger.
```
qbarthelemy, line 364: Can you add a line at the end?

examples/fNIRS/classify_fNIRS.py
```diff
@@ -0,0 +1,364 @@
+#%%
```
qbarthelemy, line 1: Can you rename the folder `fNIRS` into `fnirs`?

qbarthelemy (changes requested):
Thanks Tim for this example on a new type of data!

Can you correct all errors raised by flake8?
https://github.com/pyRiemann/pyRiemann/pull/323/checks#step:6:8

Can you complete the whatsnew file?

examples/fNIRS/classify_fNIRS.py
timnaher: @agramfort, I have just a question for you to clarify: You suggest changing the logic for creating the blocks before stacking them to block diagonal matrices for each trial, right? Is the loop over trials still okay? Because this is the way it's implemented in `block_covariances`. Thanks!

agramfort: yes I think you can avoid the 2 double for loop and just do one loop over blocks and then let the vectorized version of pairwise / covariance functions do the rest over samples.

qbarthelemy (changes requested):
Can you add a plot to display fNIRS signals?

In addition to the best acccuracy selected by cross-validation, can you print accuracy obtained on a test set. See examples/motor-imagery/plot_augmented.py

Can you compare your pipeline "BlockKernel+SVC" to the baseline pipeline "SVC"?

timnaher:
@agramfort Sorry, I can’t see how the pairwise_kernels from sklearn are vectorized. I also looked for another implementation in pyRiemann, but I can’t find any. The only vectorized kernel function I found are kernel between SPD matrices (in pyriemann/utils/kernel.py), but we need pairwise channel kernels here. Of course one could use the feature dimension to compute all possible channel (and therefore also trial) kernels and then subselect only the ones needed. However, I think this would likely drastically increase computation.

Please let me know if I am making a mistake and missing something obvious.

However, I have implemented separate pipelines for each block to make the code cleaner. The loop over trials, however, is still present.

I have also added an example plot. 

@qbarthelemy  Regarding the train-test split. For these data, there are only about 12 trials per class. We have also decided against using a dedicated train-test split in the preprint. If you like I can still add one in the example, but I am afraid it won’t be representable at all.

Regarding pipeline comparison, do you mean the basic SVC model from the preprint?

agramfort:
@timnaher I see. What I had in mind is this:

```
from pyriemann.utils import distance
import numpy as np

A = np.random.randn(100, 10, 10)
B = np.random.randn(100, 10, 10)

dists = distance.distance_euclid(A, B)
assert len(dists) == 100
```

Basically it should be easy to make the functions vectorized. See also:

```
from pyriemann.utils import covariance
import numpy as np

A = np.random.randn(100, 10, 50)

covs = covariance.covariances(A)
assert covs.shape == (100, 10, 10)
```

timnaher:
Okay, sure. When I get back from SfN, I will make the vectorized versions of these. Thanks for clarifying!

qbarthelemy:
Thanks @timnaher  for adding the figure, it's great! because in pyRiemann, we are not used to such signals.

The best accuracy obtained during cross-validation leads to overly optimistic results regarding the performance on the test set, which is the only metric that matters. IMHO, this is a critical point that you should also address in your submission.

My biggest concern is that when I run the code,
the best result is obtained with shrinkage=0.01 and not with the block-specific shrinkage [0.01, 0.1].

In addition, the current cross-validation explodes the documentation build time, which goes from 20min to 25min.
https://github.com/pyRiemann/pyRiemann/actions/workflows/deploy_ghpages.yml
To reduce this time, I removed from the grid the `classifier__metric` parameter and set it to `"riemann"`.

agramfort:
and if you avoid the for loop over samples it should be even faster.

timnaher:
although pyriemann.utils.covariance.covariances also loops over matrices and is not fully vectorized

agramfort:
take a look at something like this:

```python
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd

from sklearn.model_selection import cross_val_score, StratifiedKFold
from sklearn.compose import ColumnTransformer
from sklearn.base import TransformerMixin
from sklearn.pipeline import make_pipeline
from sklearn.linear_model import LogisticRegression

from pyriemann.tangentspace import TangentSpace
from pyriemann.estimation import Covariances

n_matrices, n_channels, n_times = 500, 3, 40
X = np.random.randn(n_matrices, n_channels, n_times)
y = np.random.choice([-1, 1], n_matrices, replace=True)

class Stacker(TransformerMixin):
    """Stacks values of a DataFrame column into a 3D array."""
    def fit(self, X, y=None):
        return self

    def transform(self, X):
        assert isinstance(X, pd.DataFrame)
        assert X.shape[1] == 1
        return np.stack(X.iloc[:, 0].values)

X_df = pd.DataFrame({"data": list(X)})

scm = make_pipeline(Stacker(), Covariances("scm"), TangentSpace())
oas = make_pipeline(Stacker(), Covariances("oas"), TangentSpace())
preprocess = ColumnTransformer([("scm", scm, ["data"]), ("oas", oas, ["data"])])
print(preprocess.fit_transform(X_df).shape)

clf = make_pipeline(preprocess, LogisticRegression())

cv = StratifiedKFold(n_splits=5, shuffle=True, random_state=42)
scores = cross_val_score(clf, X_df, y, cv=cv)
scores
```

I have the impression that this could heavily simplify your code. Basically I define two parallel processing done on the same "data" column and the ColumnTransformer then concatenates them on the axis=1 before feeding this to a logistic regression.

timnaher:
thanks! Now I know what you mean, sorry for not getting it earlier. Fantastic idea, this will heavily simplify the code. I will implement this once I get back from SfN! thanks again

timnaher:
I changed the implementation of `BlockKernels` to use the ColumnTrasformer. It's my first time using this but so cool! thanks for the pointer! I will definitely keep using this in the future :)

To check whether a metric is a valid (channel) kernel function, I import the `kernel_functions` list from `estimation.py`. This is a list with valid kernel functions. However, there is also a dict in `kernel.py` that refers to kernels between matrices. To avoid confusion and double naming, I renamed the `kernel_functions` list in `estimation.py` to `channel_kernel_functions`.

examples/fnirs/plot_classif_fnirs.py
```diff
@@ -0,0 +1,489 @@
...
+        ColumnTransformer requires 2D output, so this transformer
+        is needed"""
+    def fit(self, X, y=None):
+        return self
+
+    def transform(self, X):
+        return X.reshape(X.shape[0], -1)
+
+
+class BlockKernels(BaseEstimator, TransformerMixin):
```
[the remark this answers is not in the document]
timnaher: Yes thats right. What do you think about `BlockCovKernels` or `HybridBlockEstimator`?

qbarthelemy:
My main concern remains that this example increases the documentation build time from 19min to 25min, because of the cross-validation, which ultimately does not select the block-specific shrinkage and only uses covariance estimators and instead of kernels.

What I propose is to run a heavy offline cross-validation on a fine grid, and find the shrinkage values and estimatros ​​that highlight your method, and to use these optimal parameters in the example, without cross-validation.

Can you compare your pipeline to a basic SVC?

timnaher:
Sure, no problem! Ill do that and adjust the example accordingly

timnaher:
Sorry for the delay! I have removed the hyperparameter search and found some values that work for an example!

I have also added a comparison of the newly renamed ``HybridBlocks`` estimator to the ``BlockCovariances`` estimator.
This highlights that for fNIRS data (and maybe for other, naturally "blocked" data as well), it can be useful to select separate regularization strengths and metrics for each block.

qbarthelemy (changes requested):

examples/fnirs/plot_classif_fnirs.py
```diff
@@ -0,0 +1,484 @@
...
+
+cv_scores_blockcovariances = cross_val_score(
+    pipeline_blockcovariances, X, y,
+    cv=cv, scoring="accuracy", n_jobs=n_jobs)
+
+###############################################################################
+# Print and plot results
+# -------------
+acc_hybrid_blocks = np.median(cv_scores_hybrid_blocks)
+acc_blockcovariances = np.median(cv_scores_blockcovariances)
```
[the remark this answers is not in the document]
timnaher: sure, fine by me

examples/fnirs/plot_classif_fnirs.py
```diff
@@ -0,0 +1,482 @@
...
+    cv=cv, scoring="accuracy", n_jobs=n_jobs)
+
+###############################################################################
+# Print and plot results
+# ----------------------
+acc_hybrid_blocks = np.mean(cv_scores_hybrid_blocks)
+acc_blockcovariances = np.mean(cv_scores_blockcovariances)
+
+print(f"Median accuracy for HybridBlocks: {acc_hybrid_blocks:.2f}")
+print(f"Median accuracy for BlockCovariances: {acc_blockcovariances:.2f}")
```
qbarthelemy, line 450:
```suggestion
print(f"Mean accuracy for HybridBlocks: {acc_hybrid_blocks:.2f}")
print(f"Mean accuracy for BlockCovariances: {acc_blockcovariances:.2f}")
```

examples/fnirs/plot_classif_fnirs.py
```diff
@@ -0,0 +1,482 @@
...
+
+###############################################################################
+# Define hyperparameter cross-validation
+# --------------------------------------
+
+# Define the hyperparameters for fitting
+pipeline_hybrid_blocks.set_params(
+    block_kernels__shrinkage=[0.01, 0],
+    block_kernels__metrics=['cov', 'rbf'],
+    classifier__C=0.1
```
qbarthelemy, line 415: Hyper-parameters with only one value should be removed from the grid search.

Same remark for `pipeline_blockcovariances`.

examples/fnirs/plot_classif_fnirs.py
```diff
@@ -0,0 +1,482 @@
...
+        """Construct a block diagonal matrix from a list of square matrices."""
+        shapes = [m.shape[0] for m in matrices]
+        total_size = sum(shapes)
+        result = np.zeros((total_size, total_size), dtype=matrices[0].dtype)
+        start = 0
+        for m in matrices:
+            end = start + m.shape[0]
+            result[start:end, start:end] = m
+            start = end
+        return result
```
agramfort, line 331: I may have said it before but did you consider using https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.block_diag.html ?

agramfort (approved):
looked at the rendered doc. Doc looks good for me.

Did not do a full pass on code.

examples/fnirs/plot_classif_fnirs.py
timnaher: thanks for the pointer. This simplifies the block diagonal computation a lot

qbarthelemy (approved):
Thx @timnaher for this new example !

timnaher:
Thank you for taking the example and for your valuable feedback! This was a great learning experience, and I appreciate all your pointers and tips.

agramfort:
thanks @timnaher !

PR #329, opened 2024-10-14
Enhance mean_covariance

qbarthelemy:
This PR adds  `*arg` to `mean_covariance` to support power mean and power euclidean mean using an exponent.

### DIFF
--- a/pyriemann/spatialfilters.py
+++ b/pyriemann/spatialfilters.py
@@ -472,11 +472,11 @@ class SPoC(CSP):
         target -= target.mean()
         target /= target.std()
 
-        C = mean_covariance(X, self.metric)
+        C = mean_covariance(X, metric=self.metric)
         Ce = np.zeros_like(X)
         for i in range(Ce.shape[0]):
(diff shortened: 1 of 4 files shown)


pyriemann/spatialfilters.py
```diff
@@ -472,11 +472,11 @@ def fit(self, X, y):
         target -= target.mean()
         target /= target.std()
 
-        C = mean_covariance(X, self.metric)
+        C = mean_covariance(X, metric=self.metric)
         Ce = np.zeros_like(X)
         for i in range(Ce.shape[0]):
             Ce[i] = X[i] * target[i]
-        Cz = mean_covariance(Ce, self.metric)
+        Cz = mean_covariance(Ce, metric=self.metric)
```
agramfort, line 479: this shows that this is a non-backward compatible change without deprecation. I would suggest to avoid that.

qbarthelemy: You are the all seeing eye.

agramfort:
years of training ;)

PR #333, opened 2024-11-12
Update numpy version and remove warning

qbarthelemy:
This PR updates NumPy version and removes the following warning:

```
pyriemann\utils\base.py:4: DeprecationWarning: numpy.core.numerictypes is deprecated and has been renamed to numpy._core.numerictypes.
The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release.
In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API.
If that is the case, use the public NumPy API. If not, you are using NumPy internals.
If you would still like to access an internal attribute, use numpy._core.numerictypes.typecodes.
    from numpy.core.numerictypes import typecodes
```

### DIFF
--- a/pyriemann/utils/base.py
+++ b/pyriemann/utils/base.py
@@ -1,7 +1,6 @@
 """Base functions for SPD/HPD matrices."""
 
 import numpy as np
-from numpy._core.numerictypes import typecodes
 
 from .test import is_pos_def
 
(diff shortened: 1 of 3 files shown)


gcattan (commented):
I think we can just import typecodes from numpy:

image

qbarthelemy:
Great suggestion! Thx!

gcattan (approved):

PR #347, opened 2025-02-26
Added Riemannian t-SNE

thibaultdesurrel:
Hello everybody, 

With @sylvchev and my other supervisors, we have been working on a Riemannian version of the t-SNE algorithm for SPD matrices. The goal is to adapt the usual t-SNE algorithm to handle SPD matrices, reducing $d \times d$ SPD matrices into $2 \times 2$ SPD matrices, as those low-dimensional matrices can be visualized in a 3D space. For this, we take into account the Riemannian geometry of the manifold of SPD matrices to avoid any distortion induced by flattening the
representation in a Euclidean setup.  

Our work has finally been [published in TMLR ](https://openreview.net/pdf?id=DYCSRf3vby) and @sylvchev told me it could be a nice addition to pyRiemann. 

For this, I added the `tSNE` class in `pyriemann.embedding`. This class takes as parameter `n_components` and reduced a set of high-dimensional SPD matrices into a set of `(n_components, n_components)` SPD matrices. 

For this, I needed several utility functions, so I added them in `pyriemann.utils.base` (for `multitransp` and `multisym`) and in a new file `pyriemann.utils.gradient_descent` (for `retraction` and `norm_SPD`). Those functions are adapted from [Pymanopt](https://pymanopt.org/docs/stable/) and as Pymanopt is distributed under the open source [3-clause BSD license](https://github.com/pymanopt/pymanopt/blob/master/LICENSE), @sylvchev told me it was not a problem. (I cited them in the docstrings of the aforementioned functions). 

I also added an example `plot_Riemaniann_tSNE` in `simulated` to explain how the t-SNE works. 

Finally, I modified the file `test.embedding` to correctly test the t-SNE. I had to personalise some of the tests for the t-SNE as the result of the t-SNE is an array of matrices of size `(n_matrices, n_components, n_components)`  while the other embeddings return an array of vectors of size `(n_matrices, n_components)`.


sylvchev:
Thanks for the work @thibaultdesurrel , let's wait feedback on this big pull request.

qbarthelemy (changes requested):
I saw your TLMR article (congrats!) and I was secretly hoping for this PR, adding this nice embedding to pyRiemann.

pyriemann/embedding.py
```diff
@@ -438,28 +446,348 @@ def locally_linear_embedding(
     return embd, error
 
 
+class tSNE(BaseEstimator):
+    """AIRM Riemannian version of t-SNE algorithm.
```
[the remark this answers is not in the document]
thibaultdesurrel: I have the implemenation of the log-Euclidean t-SNE, so I can add it if you think it is a good idea. The main difference is in the computation of the gradient in the Riemannian gradient descent.

pyriemann/utils/gradient_descent.py
```diff
@@ -0,0 +1,66 @@
...
+    return multisym(point + tangent_vector + tangent_vector @ p_inv_tv / 2)
+
+
+def norm_SPD(point, tangent_vector):
+    """
+    Compute the norm of a tangent vector at a point on the manifold of
+    SPD matrices.
+    Parameters
+    ----------
+    point : ndarray, shape (..., n, n)
```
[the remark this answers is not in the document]
thibaultdesurrel: Yes, and it works, it is actually used to compute the norm of a 3d array.

examples/biosignal-erp/plot_embedding_MEG.py
```diff
@@ -1,23 +1,23 @@
...
-============================================
+===============================================
+Comparison of embeddings of covariance matrices
+===============================================
 
-Riemannian embeddings via Laplacian Eigenmaps (LE) and Locally Linear
-Embedding (LLE) of a set of ERP data. Embedding via Laplacian Eigenmaps is
-referred to as Spectral Embedding (SE).
+Comparison of several embeddings of a set of ERP covariance matrices extracted
+on MEG data: SE and LLE.
```
qbarthelemy, line 7: Rather than adding a new example, I would prefer that we complete this example:
```suggestion
on MEG data: SE, LLE and t-SNE.
```

pyriemann/utils/gradient_descent.py
```diff
@@ -0,0 +1,83 @@
...
+    This code is an adaptation from pyManopt [1]_.
+
+    Parameters
+    point : ndarray, shape (..., n, n)
+        A point on the SPD manifold.
+    tangent_vector : ndarray, shape (..., n, n)
+        A tangent vector at the point on the SPD manifold.
+    Returns
+    -------
+    norm : float
```
qbarthelemy, line 63:
```suggestion
    norm : float or ndarray, shape (...,)
```

pyriemann/utils/gradient_descent.py
```diff
@@ -0,0 +1,83 @@
...
+def norm(point, tangent_vector):
+    """Compute the norm.
+    This code is an adaptation from pyManopt [1]_.
+
+    Parameters
+    point : ndarray, shape (..., n, n)
+        A point on the SPD manifold.
+    tangent_vector : ndarray, shape (..., n, n)
+        A tangent vector at the point on the SPD manifold.
+    Returns
```
qbarthelemy, line 61: Jump a line
```suggestion
        A tangent vector at the point on the SPD manifold.

    Returns
```

pyriemann/utils/gradient_descent.py
```diff
@@ -0,0 +1,83 @@
...
+    This code is an adaptation from pyManopt [1]_.
+
+    Parameters
+    ----------
+    point : ndarray, shape (n_matrices, n, n)
+        A point on the manifold.
+    tangent_vector : array, shape (n_matrices, n, n)
+        A tangent vector at the point.
+
+    Returns
```
qbarthelemy, line 34:
```suggestion
    Returns
    -------
```

qbarthelemy (changes requested):

pyriemann/utils/gradient_descent.py
```diff
@@ -0,0 +1,85 @@
+"""Tools used for Riemannian Gradient Descent"""
```
qbarthelemy, line 1: In `pyriemann/transfer/_rotate.py` , there are already optimization functions for rotations on Grassmann manifold.

I suggest to:
- create a new folder `pyriemann/optimization`; 
- move `transfer/_rotate.py` inside and rename it into `grassmann.py`;
- and move the new `utils/gradient_descent.py` inside too, and rename it into `positive_definite.py`.

What do you think? @agramfort @sylvchev @plcrodrigues

agramfort:
no opinion on my end. You're in charge :)

pyriemann/utils/gradient_descent.py
gcattan: Just putting my naive perspective here, as this is something we could also reuse in `pyRiemann-qiskit`.

That would be nice to encapsulate all the optimization process into some kind of object. Something similar to what is doing [Pennylane](https://docs.pennylane.ai/en/stable/code/api/pennylane.RiemannianGradientOptimizer.html) or [QGOpt ](https://github.com/LuchnikovI/QGOpt/blob/master/QGOpt/optimizers/SGD.py)for tensor flow.

plcrodrigues: Sounds good to me @qbarthelemy !

qbarthelemy: @gcattan, I propose to study the idea of ​​objectifying optimization functions in another issue.

@thibaultdesurrel , do you want to take care of refactoring the code as suggested, or do you want me to take care of it?

thibaultdesurrel: @qbarthelemy I can take care of it. However, I may be missing other optim functions (other than the one that you cited above) but I'll do my best !

pyriemann/embedding.py
```diff
@@ -438,28 +446,344 @@ def locally_linear_embedding(
...
+                @ Y_i_sqrt[i]
+            )
+            grad[i] = -4 * np.sum(
+                ((P[i] - Q[i]) / (1 + Dsq[i]))[:, np.newaxis, np.newaxis] *
+                log_riemann,
+                axis=0,
+            )
+        return grad
+
+    def _run_minimization(self, P):
```
qbarthelemy, line 613: If you look at `TLRotate.fit()`, you can see that most of the optimization is done on the `grassmann` module.
https://github.com/pyRiemann/pyRiemann/blob/master/pyriemann/transfer/_estimators.py#L693

Can you do the same for `TNSE.fit()`, and put `_run_minimization` (and subfunctions) in module `positive_definite`?

pyriemann/embedding.py
```diff
@@ -438,28 +446,344 @@ def locally_linear_embedding(
...
+        self : TSNE instance
+            The TSNE instance.
+        """
+        n_matrices, _, _ = X.shape
+
+        if self.perplexity is None:
+            self.perplexity = int(0.75 * n_matrices)
+
+        # Compute similarities in the high dimension space
+        P = self._compute_similarties(X)
```
qbarthelemy, line 724: Typo
```suggestion
        P = self._compute_similarities(X)
```

pyriemann/embedding.py
```diff
@@ -438,28 +446,344 @@ def locally_linear_embedding(
...
+
+        # Sample initial solution close to the identity
+        sigma = 1
+        self.initial_point = sample_gaussian_spd(
+            n_matrices,
+            mean=np.eye(self.n_components),
+            sigma=sigma,
```
qbarthelemy, line 731: `sigma` is not re-used
```suggestion
        self.initial_point = sample_gaussian_spd(
            n_matrices,
            mean=np.eye(self.n_components),
            sigma=1,
```

pyriemann/embedding.py
```diff
@@ -438,28 +446,344 @@ def locally_linear_embedding(
...
+                    )
+                )
+        return current_sol
+
+    def fit(self, X, y=None):
+        """Fit Riemannian t-SNE.
+
+        Parameters
+        ----------
+        X : array_like of shape (n_matrices, n_channels, n_channels)
```
qbarthelemy, line 708:
```suggestion
        X : ndarray, shape (n_matrices, n_channels, n_channels)
```

qbarthelemy (changes requested):

pyriemann/utils/viz.py
```diff
@@ -44,32 +48,51 @@ def plot_embedding(X,
...
     if embd_type == "Spectral":
-        lapl = SpectralEmbedding(n_components=2, metric=metric)
+        e = SpectralEmbedding(n_components=2, metric=metric)
     elif embd_type == "LocallyLinear":
-        lapl = LocallyLinearEmbedding(n_components=2,
+        e = LocallyLinearEmbedding(n_components=2,
                                       n_neighbors=X.shape[1],
                                       metric=metric)
+    elif embd_type == "TSNE":
+        e = TSNE(n_components=2, metric=metric)
```
qbarthelemy, line 57: The example takes a looong time to build...
```suggestion
        e = TSNE(n_components=2, metric=metric, max_iter=1000)
```

qbarthelemy (changes requested):
Thanks for all the work already done!

pyriemann/embedding.py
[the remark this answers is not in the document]
qbarthelemy: Is it possible to add the calculation of the euclidean gradient?

pyriemann/utils/viz.py
sylvchev: I think we should start with smaller number of electrodes for the t-SNE in the example. That being said, it is a good idea to introduce a default limit for the max iteration.

pyriemann/embedding.py
```diff
@@ -438,28 +444,208 @@ def locally_linear_embedding(
...
+        n_matrices, _, _ = X.shape
+
+        if self.perplexity is None:
+            self.perplexity = int(0.75 * n_matrices)
+
+        # Compute similarities in the high dimension space
+        P = self._compute_similarities(X)
+
+        # Sample initial solution close to the identity
+        self.initial_point = sample_gaussian_spd(
```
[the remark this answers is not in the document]
thibaultdesurrel: I'm not sure to understand correclty: do you want me to put the computations of the initial solution in `positive_definite` or do you want me to adapt `grassmann._warm_start()` for the t-SNE ? For the latter, I'm not sure to see how to do, and in the t-SNE algorithm, the initial solution is sampled with a multivariate Gaussian with a same variance.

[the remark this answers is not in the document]
qbarthelemy: Yes, put the computation of the initial solution in `positive_definite`. Sorry if my comment was ambiguous.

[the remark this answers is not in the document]
thibaultdesurrel: Ok !

pyriemann/embedding.py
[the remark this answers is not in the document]
thibaultdesurrel: I was thinking of returning the [t-SNE of sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.manifold.TSNE.html) when 'metric == euclid' (with the euclidean Frobenius distance) but it turns out that it is more complicated than that as we expect arrays of matrices of size '(n_matrices, n_channels, n_channels)' as input of the `fit` function but the sklearn version expects an array of vectors of size ` (n_samples, n_features)`. So a non-optimal solution (IMHO) is to implement the Euclidean gradient but it will be a lot slower than the sklearn version... Or I can use the sklearn version with `metric == "precomputed"` and add a special case in the `fit` function when the metric is Euclidean... Do you have an opinion on which option to choose?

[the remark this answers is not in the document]
qbarthelemy: Implementing the Euclidean gradient is ok for me.

pyriemann/embedding.py
```diff
@@ -438,28 +443,209 @@ def locally_linear_embedding(
...
+
+        Parameters
+        ----------
+        Y : ndarray, shape (n_matrices, n_components, n_components)
+            Set of low-dimensional SPD matrices.
+
+        Returns
+        -------
+        Q : ndarray, shape (n_matrices, n_matrices)
+            Low-dimensional similarities conditional probabilities of Y.
```
qbarthelemy, line 545: Can you be more precise on descriptions?
For `Q`, are these similarities or probabilities? Are they conditional or joint?

examples/biosignal-erp/plot_embedding_MEG.py
```diff
@@ -59,23 +62,24 @@
 y = epochs.events[:, -1]
 
 ###############################################################################
-# Embedding of Xdawn covariance matrices
-# --------------------------------------
+# Extract Xdawn covariance matrices
+# ---------------------------------
 
-nfilter = 4
+nfilter = 2
```
sylvchev, line 68: We should keep 4 XDAWN components

examples/simulated/plot_Riemaniann_tSNE.py
```diff
@@ -0,0 +1,111 @@
...
+from pyriemann.datasets import make_matrices, sample_gaussian_spd
+
+
+print(__doc__)
+
+###############################################################################
+# Set parameters for sampling from the Riemannian Gaussian distribution
+
+n_matrices_per_class = 50  # how many SPD matrices to generate in each class
+n_dim = 6  # number of dimensions of the SPD matrices
```
sylvchev, line 30: Lower n_dim

examples/biosignal-erp/plot_embedding_MEG.py
```diff
@@ -85,10 +89,23 @@
 plt.show()
 
 
+###############################################################################
+# TNSE
+# ------------------------------
+
+plot_embedding(covs, ytest, metric="riemann", embd_type="TSNE",
```
sylvchev, line 96: For tSNE, we could select only 2 classes on the 4 from this dataset, and use 2 components from XDAWN. In this case, we should generate a new set of cov mats. This should largely decrease the computational time for doc building.

sylvchev (commented):
I am quite confident that we could keep the doc building time reasonable. If this is the case, we could keep the 2 examples, as they illustrates different setup.

qbarthelemy:
Thx @thibaultdesurrel !

sylvchev:
Kudos!

PR #350, opened 2025-03-13
Enhance pairwise_distance

qbarthelemy:
`pairwise_distance` used with metric = "euclid", "harmonic", "logchol", "logeuclid" on HPD matrices raises
`ValueError: Complex data not supported`
because `euclidean_distances` of `sklearn` does not support complex-valued arrays.

Related to #204

@thibaultdesurrel


gcattan (approved):
All good to me! Thx @qbarthelemy !

PR #364, opened 2025-05-27
Add MeanShift

qbarthelemy:
This PR:
- adds mean shift for non-parametric clustering in the manifold of SPD/HPD matrices,
 see algo described in Fig 4 of Subbarao2009;
- refactors and completes tests.


sylvchev (approved):
Thanks for implementing MeanShift, it is a very usefull addition!

PR #365, opened 2025-06-19
added a more robust downloading helper function

AmmarMian:
Seems to solve docs issue.


AmmarMian:
Ok it fails because of flake8. Maybe I can let you handle that @qbarthelemy ?

PR #388, opened 2025-10-20
Add parallel transports for positive-definite matrices

qbarthelemy:
Add `transport()` function for "euclid", "logeuclid" and "riemann" metrics for positive-definite matrices, and complete tests.

I don't know what reference to use for parallel transport for the log-Euclidean metric.
Do you have any suggestions? @plcrodrigues @gabelstein @jbschiratti

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -190,6 +190,10 @@ Datasets
     sample_gaussian_spd
     make_classification_transfer
 
+    :template: class.rst
+
+    RandomOverSampler
+
 Utils function
 --------------
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -18,6 +18,8 @@ v0.10.dev
 
 - Deprecate ``fit_transform()`` of :class:`pyriemann.spatialfilters.AJDC` due to incompatible dimensions. :pr:`382` by :user:`qbarthelemy`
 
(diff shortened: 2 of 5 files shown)


jbschiratti:
Hello @qbarthelemy :-) 

I couldn't find a good reference for parallel transport with the log-Euclidean metric. [This reference](https://www.sciencedirect.com/science/article/pii/S0004370225000669) has a sentence about it, though... just before equation (1). In addition to this, it may be enough to explain what's happening: the diffeomorphism exp : SYM(n) -> SPD(n) is a isometry for the log-Euclidean metric. Since SYM(n) - equipped with the Frobenius inner product - is a flat Euclidean vector space (i.e. curvature tensor = 0 everywhere), SPD(n) - equipped with the log-Euclidean metric - is also flat. That's why parallel transport is identity in this setting. Unless I'm mistaken, it corresponds to what you implemented [here](https://github.com/pyRiemann/pyRiemann/pull/388/files#diff-cdd86c52ad8ca0de79216e68c67e807eb9b6976dae2930c3433bb7cc1541d6e4R720).

gabelstein:
@qbarthelemy Probably has to do with the sqrtm function (guessing it is the one from pyriemann). It is only properly defined on SPD matrices and `(B @ np.linalg.inv(A))` is not symmetric.

qbarthelemy:
Yes, you're right, that was the problem. Using scipy's sqrtm solves it. Thanks!

Issue #409, opened 2026-01-05
Improve sample_gaussian_spd to support ndarray for sigma


qbarthelemy:
Improve the `sample_gaussian_spd` function so that the `sigma` parameter can be an ndarray.

```
    sigma : float | ndarray, shape (n_dim * (n_dim + 1) / 2, \
            n_dim * (n_dim + 1) / 2)
        If float, dispersion of the Riemannian Gaussian distribution [1]_.
        If ndarray, covariance matrix of the Riemannian Gaussian distribution
        projected in tangent space [2]_. 
```

@thibaultdesurrel

PR #411, opened 2026-01-06
Add GaussianMixture

qbarthelemy:
Implement [Gaussian Mixture Model (GMM)](https://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html) for SPD matrices, from [https://calinon.ch/papers/Jaquier-IROS2017](https://calinon.ch/papers/Jaquier-IROS2017.pdf) and https://gitlab.com/martijnzeestraten/riepybdlib

To add `sample` function, first solve #409.

@thibaultdesurrel 

Also, fix `TypeError: mean_euclid() got an unexpected keyword argument 'init'` when adding Gaussian Mixture in the comparison of clustering algorithms in the gallery of examples.

### DIFF
--- a/pyriemann/clustering.py
+++ b/pyriemann/clustering.py
@@ -582,7 +582,7 @@ class Gaussian():
             metric, ["mean", "map"]
         )
 
-    def pdf(self, X, reg=1e-16, use_pi=True):
+    def pdf(self, X, *, reg=1e-16, use_pi=True):
         """Compute approximate probability density function (pdf) of matrices.
 
         Parameters
@@ -683,11 +683,11 @@ class GaussianMixture(SpdClustMixin, BaseEstimator):
     Attributes
     ----------
     weights_ : ndarray, shape (n_components,)
-        The weights of each mixture components.
+        Weight of each mixture component.
     means_ : ndarray, shape (n_components, n_channels, n_channels)
-        The mean of each mixture component.
+        Mean of each mixture component.
     covariances_ : ndarray, shape (n_components, n_ts, n_ts)
-        The covariance of each mixture component.
+        Covariance of each mixture component.
 
     Notes
     -----
@@ -748,7 +748,8 @@ class GaussianMixture(SpdClustMixin, BaseEstimator):
         """
         wlik = np.zeros((X.shape[0], self.n_components))
         for k in range(self.n_components):
-            wlik[:, k] = self.weights_[k] * self._components[k].pdf(X)
+            lik = self._components[k].pdf(X, use_pi=use_pi)
+            wlik[:, k] = self.weights_[k] * lik
         return wlik
 
     def _get_proba(self, X, reg=1e-16):
@@ -760,7 +761,6 @@ class GaussianMixture(SpdClustMixin, BaseEstimator):
             Set of SPD matrices.
         reg : float, default=1e-16
             Regularization parameter for probabilities normalization.
-
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


pyriemann/clustering.py
```diff
@@ -518,13 +525,333 @@ def predict(self, X):
...
+        """Update covariance in tangent space.
+
+        Compute weighted covariance of tangent vectors.
+
+        Parameters
+        ----------
+        X : ndarray, shape (n_matrices, n, n)
+            Set of SPD matrices.
+        sample_weight : ndarray, shape (n_matrices,)
+            Weights for each matrix.
```
sylvchev:
```suggestion
        sample_weight : ndarray, shape (n_matrices,)
            Weights for each matrix.
        reg : float, default=1e-16
             Regularization parameter for sigma estimator
```

pyriemann/clustering.py
```diff
@@ -518,13 +525,333 @@ def predict(self, X):
...
+        """
+        TangVec = tangent_space(X, self.mu, metric=self._metric_map)
+        sigma = TangVec.T @ (sample_weight[:, np.newaxis] * TangVec)
+        self.sigma = sigma / (sample_weight.sum() + reg)
```
sylvchev: What is the point of adding reg here? If there are constraints that sample weight sum should be positive or bounded, we could state it and maybe test it?

pyriemann/clustering.py
```diff
@@ -518,13 +525,333 @@ def predict(self, X):
...
+
+
+class GaussianMixture(SpdClustMixin, BaseEstimator):
+    """Gaussian mixture model.
+
+    Representation of a Gaussian mixture model probability distribution for SPD
+    matrices.
```
sylvchev:
```suggestion
    matrices [1]_.
```

sylvchev (commented):
Nice addition @qbarthelemy !
I don't have a definitive idea regarding the default number of components for GMM, this is just a suggestion. 
Do we need to check the weights more thoroughly in the gaussian estimator?

pyriemann/clustering.py
```diff
@@ -849,9 +871,36 @@ def score(self, X, y=None):
...
+        y = self.random_state.randint(self.n_components, size=(n_matrices,))
+
+        X = np.zeros((n_matrices, self.dim, self.dim))
+        for i in np.unique(y):
+            X[y == i] = sample_gaussian_spd(
+                np.count_nonzero(y == i),
+                mean=self.components_[i].mu,
+                sigma=self.components_[i].sigma,
+                random_state=self.random_state
+            )
```
thibaultdesurrel: Caution @qbarthelemy , the sampling I implemented with a `ndarray` covariance matrix samples from the [wrapped Gaussian distribution](https://openreview.net/pdf?id=EhStXG4dCS) which is not exactly the same as the one used in the paper of [Jaquier et al](https://calinon.ch/papers/Jaquier-IROS2017.pdf). Mathematically speaking, the two distributions are not the same, I am not sure how different they are in practice and if sampling from the wrapped Gaussian will really differ from the other one...

pyriemann/clustering.py
```diff
@@ -518,13 +525,333 @@ def predict(self, X):
...
+    References
+    ----------
+    .. [1] `Gaussian mixture regression on symmetric positive definite matrices
+        manifolds: Application to wrist motion estimation with sEMG
+        <https://calinon.ch/papers/Jaquier-IROS2017.pdf>`_
+        N. Jacquier & S. Calinon. IEEE IROS, 2017
+    """
+    def __init__(
+        self,
+        n_components=3,
```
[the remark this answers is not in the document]
sylvchev: Ok, seems reasonable to use 1

PR #420, opened 2026-02-05
Correct log-Euclidean parallel transport

qbarthelemy:
1) Correct log-Euclidean parallel transport

The log-Euclidean parallel transport added in #388 was implemented as

$$ \mathbf{X}_\text{new} = \mathbf{X} + (\log \mathbf{B} - \log \mathbf{A}) $$

However, this formula is not correct,
see Table 4 in Thanwerdas2023,  [O(n)-invariant Riemannian metrics on SPD matrices](https://www.sciencedirect.com/science/article/pii/S0024379522004360)

The log space is Euclidean, so the parallel transport is the identity **after** going through $[D_{\mathbf{A}} \log]$ (to push vector towards flat space in $\mathbf{A}$) and returning through $[D_{\log \mathbf{B}} \exp]$ (to pull transported vector to $\mathbf{B}$):

 $$ \mathbf{X}_\text{new} = [D_{\log \mathbf{B}} \exp] \left( [D_{\mathbf{A}} \log]\left(\mathbf{X}\right) \right)  $$

where $[D_{\mathbf{U}} \log] \left( \mathbf{V}\right)$ indicates the differential of the matrix logarithm at point $\mathbf{U}$ applied to $\mathbf{V}$,
and $[D_{\mathbf{U}} \exp] \left( \mathbf{V}\right)$ the differential of the matrix exponential.

@ythanwerdas @plcrodrigues @gabelstein @jbschiratti, is it ok for you?

2) Complete tests and correct formula of log-Cholesky parallel transport for HPD matrices.

### DIFF
--- a/pyriemann/utils/tangentspace.py
+++ b/pyriemann/utils/tangentspace.py
@@ -730,7 +730,7 @@ def transport_logchol(X, A, B):
     Parameters
     ----------
     X : ndarray, shape (..., n, n)
-        Symmetric/Hermitian matrices in tangent space.
+        Symmetric/Hermitian matrices in tangent space at A.
     A : ndarray, shape (n, n)
         Initial SPD/HPD matrix.
     B : ndarray, shape (n, n)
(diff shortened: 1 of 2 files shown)


gabelstein:
@qbarthelemy Looks good.

Issue #425, opened 2026-03-09
[EHN] Some utility functions within Pyriemann are not broadcast compatible


bruAristimunha:
Hey @qbarthelemy, @agramfort, @sylvchev,

As we discussed by email, I am working to port the utility part here to have a backend within pytorch. ONe thing that I notice is that not all the util functions are broadcast compatible. 

mostly because of: `_check_inputs`, `np.einsum`, `_recursive`,  and `np.spacing`.

I will make one PR to make everything broadcast compatible, and then the torch PR.

PR #445, opened 2026-04-30
Move geometry modules from utils/ to standalone geometry/ package

bruAristimunha:
Reopens the reorganization proposed in #425, replacing the now-conflicting #432.

## Summary

Reorganize the 8 geometry modules out of `utils/` into a dedicated `geometry/` package, with backward-compat shims so every existing import path keeps working unchanged.

The PR is structured as **three reviewable commits** — each one a single, self-contained operation:

| # | Commit | Diff |
|---|---|---|
| 1 | Move geometry modules from utils/ to geometry/ package | 8 × R100 renames, **+0 / −0** |
| 2 | Fix geometry imports and add backward-compat shims in utils/ | 15 files, +65 / −17 |
| 3 | Extract shared helpers into pyriemann/_helpers.py | 11 files, +379 / −331 |

### Commit 1 — pure file moves

`git mv` of `pyriemann/utils/{base, covariance, distance, geodesic, mean, median, tangentspace, ajd}.py` into `pyriemann/geometry/`. Git's rename detection reports R100 for all 8 modules — zero content changes — so reviewers can verify the move trivially.

### Commit 2 — wire it up

- Inside `pyriemann/geometry/*.py`, rewrite non-sibling imports that still pointed back into `utils/` siblings (`._backend`, `.test`, `.utils`, `. import deprecated`) to absolute paths. Sibling imports between geometry modules remain intact.
- Add 8 thin shim files at `pyriemann/utils/{base, covariance, ...}.py` using the `sys.modules` aliasing pattern, which transparently forwards every public *and* private attribute to the new `pyriemann.geometry.*` location. No caller in `pyriemann/` or in user code needs an import update.

### Commit 3 — extract shared helpers

Introduce `pyriemann/_helpers.py` as a leaf module (only stdlib + numpy + array_api_compat, zero internal pyriemann imports). It hosts: `deprecated`, `check_function`, `check_init`, `check_like`, `check_matrix_pair`, `check_weights`, `is_real_type`, `is_square`.

`pyriemann/utils/{utils, test, docs}.py` re-export those names from `_helpers` so every prior public path still resolves. `pyriemann/geometry/*.py` imports the helpers directly from `pyriemann._helpers` — the only `utils/` dependency left in geometry/ is `pyriemann.utils._backend` (the array-API backend selector, itself a leaf).

### Architecture

```
pyriemann/_helpers.py              ← leaf: warnings, numpy, array_api_compat
     ↑                ↑
     |                |
pyriemann/geometry/   pyriemann/utils/
  (canonical)           ├─ _backend.py            (unchanged)
                        ├─ utils.py, test.py, docs.py  (re-exports from _helpers)
                        └─ {base,covariance,...}.py    (sys.modules shims)
```

## Test plan

- [x] `pytest tests/` — 3421 passed, 1227 skipped (same as master)
- [x] All previous import paths still resolve: `pyriemann.utils.base.sqrtm`, `pyriemann.utils.mean.mean_riemann`, `pyriemann.utils.docs.deprecated`, etc.
- [x] Both public and private (`_eigvalsh`, `_first_divided_difference`, `_complex_estimator`, ...) names accessible through the shims thanks to `sys.modules` aliasing
- [x] `pyriemann.utils.<module> is pyriemann.geometry.<module>` after import — same module object, no attribute drift

Closes #432.


pyriemann/utils/ajd.py
```diff
@@ -1,415 +1,6 @@
-"""Aproximate joint diagonalization algorithms."""
+"""Backward-compatibility shim. Module moved to pyriemann.geometry.ajd."""
+import sys
 
-import warnings
+from pyriemann.geometry import ajd as _moved
```
qbarthelemy: Can you keep relative import, instead of absolute import?
```suggestion
from ..geometry import ajd as _moved
```
Same remark for all modified imports.

pyriemann/utils/utils.py
```diff
@@ -1,117 +1,12 @@
 import inspect
```
qbarthelemy: Can you take advantage of this PR to rename `utils.utils` to `utils._check`?

qbarthelemy (changes requested):
Thx @bruAristimunha for this improvement!

Backward-compatibility shim is great, but is should raise a deprecation warning for users.

Can you complete whatsnew file?
Can you also rename the tests test_utils_* to test_geometry_*?

qbarthelemy (changes requested):
Can you update `api.rst`, splitting "Utility Functions" section into "Geometry Functions" and  "Utility Functions?

For `geometry` to be a standalone subpackage, it shouldn't have any external imports.
However, current `geometry` submodules often import `_helpers` and `utils._backend`.
Why not put these two private submodules within `geometry`?

The entire `test` module should be moved to `geometry`, to avoid splitting it in two parts.

Why not move `kernel` from `utils` to `geometry`?

qbarthelemy (approved):
Thank you for this new tree structure, @bruAristimunha !

Issue #396, opened 2025-11-04
Improve geodesic to take ndarray of alpha


qbarthelemy:
The `pyriemann.utils.geodesic` function allows to calculate the geodesic between matrices `A` and `B`, which are ndarrays of shape (..., n, n).
The `alpha` parameter is currently a float, and this value is applied to all stacked n x n matrices.

The purpose of this issue is to allow `alpha` to be an ndarray of shape (..., 1).
In this case, each pair of input matrices is associated with a different value of alpha.

Don't forget to complete the tests.

Issue related to this comment https://github.com/pyRiemann/pyRiemann/pull/387#discussion_r2427937344

Fashad-Ahmed:
@qbarthelemy Hi, I want to pick this issue. Is it available to be picked?

qbarthelemy:
Yes, gladly. Don't forget to add tests for the new format of the `alpha` parameter.

Fashad-Ahmed:
> Yes, gladly. Don't forget to add tests for the new format of the `alpha` parameter.

Sure, thanks!

PR #446, opened 2026-05-02
ENH: allow array-valued alpha in utils.geodesic

Fashad-Ahmed:
Issue #396

## Summary
- Extend `pyriemann.utils.geodesic.geodesic` to accept `alpha` as either a scalar or an ndarray of shape `(..., 1)`.
- When `alpha` is array-valued, apply one `alpha` per stacked pair of input matrices and return the same stacked output shape.
- Add input validation for array-shaped `alpha` with a clear `ValueError` when shape is invalid.
- Add tests for:
  - valid ndarray `alpha` behavior across all supported geodesic metrics
  - invalid ndarray `alpha` shape rejection
- Add a `whatsnew` entry under `v0.12.dev`.

## Motivation
Issue #396 requests support for per-matrix interpolation factors when `A` and `B` are stacked matrices.  
Before this change, `alpha` was effectively scalar-only and applied uniformly to all stacked matrices.

## Behavior
- Supported:
  - `alpha` scalar (existing behavior)
  - `alpha` ndarray with shape matching `(*A.shape[:-2], 1)`
- Invalid shapes now raise:
  - `ValueError: alpha must have shape (..., 1) ...`

## Test Plan
- `pytest -q tests/test_utils_geodesic.py -k "alpha_array or geodesic_broadcasting"`
- `pytest -q tests/test_utils_geodesic.py`


qbarthelemy:
Thanks for your work.
Actually, I think it's more optimal to use broadcasting of operators and to implement the modifications for each metric.

For example, for Euclidean geodesic:

```
def geodesic_euclid(A, B, alpha=0.5):
    """Euclidean geodesic between matrices.

    ...

    alpha : float | ndarray, shape (...,), default=0.5
        Position on the geodesic.
    """
    if hasattr(alpha, "ndim"):
        if alpha.ndim == A.ndim - 2:
            alpha = alpha[..., None, None]
        else:
            raise an error
    return (1 - alpha) * A + alpha * B
```

Fashad-Ahmed:
> Thanks for your work. Actually, I think it's more optimal to use broadcasting of operators and to implement the modifications for each metric.
> 
> For example, for Euclidean geodesic:
> 
> ```
> def geodesic_euclid(A, B, alpha=0.5):
>     """Euclidean geodesic between matrices.
> 
>     ...
> 
>     alpha : float | ndarray, shape (...,), default=0.5
>         Position on the geodesic.
>     """
>     if hasattr(alpha, "ndim"):
>         if alpha.ndim == A.ndim - 2:
>             alpha = alpha[..., None, None]
>         else:
>             raise an error
>     return (1 - alpha) * A + alpha * B
> ```

Sure, I will do it so that it looks cleaner and more interpretable.

qbarthelemy (changes requested):
Thx @Fashad-Ahmed !

In your code, you modify the general function `geodesic`, reshaping the matrices, then processing them in a 3D stack, and finally un-reshaping them (to do this, we could simply use the decorator `base._vectorize_nd`, see `mean` module).

But, currently, if we import `geodesic_euclid`, it cannot treat `alpha` as an ndarray.

In fact, the goal of this issue is to enable all `geodesic_*` functions to treat `alpha` as an ndarray. And to achieve this, linear algebra function broadcasting must be used instead of reshaping.

doc/whatsnew.rst
```diff
@@ -10,6 +10,10 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.12.dev
 ---------
 
+- Enhance :func:`pyriemann.utils.geodesic.geodesic` to accept ``alpha`` as an ndarray of shape ``(..., 1)``,
```
qbarthelemy: After careful consideration, it is more logical to ask that alpha be of the form (...,) rather than with an additional dimension, (..., 1).

pyriemann/utils/geodesic.py
```diff
@@ -416,5 +458,37 @@ def geodesic(A, B, alpha, metric="riemann"):
...
+        elif metric_name == "thompson":
+            alpha = alpha[..., 0]
+        else:
+            A_flat = xp.reshape(A, (-1, *A.shape[-2:]))
+            B_flat = xp.reshape(B, (-1, *B.shape[-2:]))
+            alpha_flat = xp.reshape(alpha, (-1,))
+            C_flat = xp.stack(
+                [
+                    geodesic_function(A_flat[i], B_flat[i], alpha_flat[i])
+                    for i in range(A_flat.shape[0])
```
qbarthelemy: This part has become unnecessary, since each function handles the broadcasting.

pyriemann/utils/geodesic.py
```diff
@@ -55,6 +58,12 @@ def geodesic_chol(A, B, alpha=0.5):
         Ann Appl Stat, 2009, 3(3), pp. 1102-1123.
     """
     xp = get_namespace(A, B)
+    if hasattr(alpha, "ndim"):
+        alpha = xp.asarray(alpha, device=xpd(A))
```
qbarthelemy: Why this step in each function?
It seems logical to require that `A`, `B`, and `alpha` all be on the same backend.

qbarthelemy (changes requested):
Thanks for the edits, that's exactly the idea behind the issue.

Following the merge of #445, the code will need to be moved into `geometry` subpackage.

Fashad-Ahmed:
> Thanks for the edits, that's exactly the idea behind the issue.
> 
> Following the merge of #445, the code will need to be moved into `geometry` subpackage.

Thanks for the response, I was on quick rest due to sickness. I will surely resolve the feedbacks asap :)

qbarthelemy:
No problem. Get well soon.

qbarthelemy (approved):
@Fashad-Ahmed, I've made a few changes to the code;
if they work for you, we can merge them.
Thanks for the improvement!

Fashad-Ahmed:
> @Fashad-Ahmed, I've made a few changes to the code; if they work for you, we can merge them. Thanks for the improvement!

Thank you for the feedback. The change seems fine to me, we can merge it ( and thanks for the wishes) :)

PR #450, opened 2026-05-29
Add innerproduct_logchol for symmetric matrices

qbarthelemy:
Add `innerproduct_logchol` for symmetric matrices,
following [Lin2019, Riemannian geometry of symmetric positive definite matrices via Cholesky decomposition](https://arxiv.org/pdf/1908.09326).

Test `test_innerproduct_property_linearity` is KO for Hermitian matrices, even transforming `.T` into `.conj().T`

@bruAristimunha , @gabelstein , do you have the opportunity to look at this PR?


gabelstein:
```def innerproduct_logchol2(X, Y, Cref):
    same = Y is None or Y is X
    if Y is None:
        Y = X

    xp = check_matrix_pair(X, Cref, require_square=True)
    n = Cref.shape[-1]
    C_chol = xp.linalg.cholesky(Cref)
    eye_n = xp.eye(n, dtype=Cref.dtype, device=xpd(Cref))
    C_invchol = xp.linalg.solve(C_chol, eye_n)

    tri0, tri1 = tril_indices(n, -1)
    diag0, diag1 = diag_indices(n)

    def _diff(W):
        S = C_invchol @ W @ C_invchol.conj().T      # L^{-1} W L^{-*}
        Z = xp.zeros_like(S)
        Z[..., tri0, tri1] = S[..., tri0, tri1]
        Z[..., diag0, diag1] = S[..., diag0, diag1] / 2
        dL = C_chol @ Z                             # tangent vector at L
        return dL[..., tri0, tri1], S[..., diag0, diag1]

    offX, dgX = _diff(X)
    offY, dgY = (offX, dgX) if same else _diff(Y)

    off = xp.sum(offX.conj() * offY, axis=-1)
    diag = xp.sum(dgX.conj() * dgY, axis=-1) / 4
    return (off + diag).real
```

should do the trick and is about twice as fast. Main problem was the `diffX_ = C_invchol @ X @ C_invchol.T` which needed to be `diffX_ = C_invchol @ X @ C_invchol.conj().T`.

On another note: the validity of the inner poducts can be tested by leveraging knownlegde about the distance. The test could then be

```
@pytest.mark.parametrize("kindX, kindC", [("sym", "spd"), ("herm", "hpd")])
@pytest.mark.parametrize("metric", metrics)
def test_innerproduct_x_x(kindX, kindC, metric, get_mats):
    """Test innerproduct for X = Y"""
    if kindC == "hpd" and metric == "logchol":
        pytest.skip()
    n_matrices, n_channels = 5, 3
    X = get_mats(n_matrices, n_channels, kindX)

    Cref = get_mats(1, n_channels, kindC)[0]
    G = innerproduct(X, X, Cref, metric=metric)
    assert G.shape == (n_matrices,)
    assert is_real(G)

    G1 = innerproduct(X, None, Cref, metric=metric)
    assert_array_equal(G, G1)

    Xexp = exp_map(X, Cref, metric=metric)
    G2 = distance(Xexp, Cref, metric=metric).flatten()
    assert_array_almost_equal(G, G2**2)
```
However, this fails for the 'riemann' metric as the AIR exp map and log map are still incorrect with Cm12=False.

qbarthelemy:
Thank you for these helpful comments, @gabelstein 
I will incorporate them.

qbarthelemy:
Your suggestion optimizes the calculation well when Y is None, 
but there is the same errors as before on the linearity test `test_innerproduct_property_linearity`.

gwvb:
True. To be honest, I am unsure whether this test is actually testing what we would like the inner products to do. 
And I also don't know if the Log Cholesky inner product can actually ever fulfil this test as is it is not defined for HPD matrices in the paper and the transfer can be a bit weird with the Cholesky decomposition. I will have to look a lot deeper into this, but this may take a while.

qbarthelemy (commented):
The documentation only indicates that it supports SPD matrices.
We will open another pull request once we have resolved the linearity test issue for the HPD case.

gabelstein:
@qbarthelemy I looked into it a bit more and I think the problem is in the tests. `a * X + b * Y` is not hermitean anymore for `X` and `Y` hermitean and `a` and `b` complex. The other metrics seem to still work for this case but it is not really the linearity test that is performed here. 

There are some things that I'm not entirely clear on yet on the theory side. But if I'm not mistaken, the tangent space (even that of HPD matrices) is a real vector space. So testing linearity with complex numbers does not really make sense, they should always only be real numbers. Then everything should work fine for the logchol metric.

qbarthelemy:
Yes, you are right: `a * X + b * Y` is Hermitian for `X` and `Y` Hermitian, and `a` and `b` real, not complex.
I'm going to correct the tests.

PR #453, opened 2026-06-15
Move Potato and PotatoField from clustering to artifactdetection

qbarthelemy:
Move classes `Potato` and `PotatoField` from module `clustering` to a new module `artifactdetection`,

like https://scikit-learn.org/stable/modules/outlier_detection.html


agramfort (approved):

agramfort (commented):
you may just consider calling the module artifact_detection.py as it's 2 words

PR #455, opened 2026-06-24
Add innerproduct_wasserstein for Hermitian matrices

qbarthelemy:
Add `innerproduct_wasserstein` for Hermitian matrices,
following [Malago2018, Wasserstein Riemannian geometry of Gaussian densities](https://arxiv.org/pdf/1801.09269)
and [Bathia2019, On the Bures–Wasserstein distance between positive definite matrices](https://www.sciencedirect.com/science/article/pii/S0723086918300021)

Tests `test_innerproduct_property_distance` are KO.

### DIFF
--- a/pyriemann/geometry/tangentspace.py
+++ b/pyriemann/geometry/tangentspace.py
@@ -237,16 +237,8 @@ def exp_map_wasserstein(X, Cref, **kwargs):
         pp. 137–179.
     """
     xp = check_matrix_pair(X, Cref, require_square=True)
-    d, V = xp.linalg.eigh(Cref)
-    Vh = ctranspose(V)
-    C = 1 / (d[:, None] + d[None, :])
-
-    X_rotated = Vh @ X @ V
-    X_tmp = C * X_rotated
-    X_tmp = X_tmp @ (d[..., None] * X_tmp)
-    X_tmp = V @ X_tmp @ Vh
-
-    return Cref + X + X_tmp
+    LX = xp.asarray(solve_continuous_lyapunov(Cref, X))
+    return Cref + X + LX @ Cref @ LX
 
 
 exp_map_functions = {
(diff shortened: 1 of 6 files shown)


pyriemann/geometry/tangentspace.py
```diff
@@ -236,16 +237,8 @@ def exp_map_wasserstein(X, Cref, **kwargs):
...
-    C = 1 / (d[:, None] + d[None, :])
-
-    X_rotated = Vh @ X @ V
-    X_tmp = C * X_rotated
-    X_tmp = X_tmp @ (d[..., None] * X_tmp)
-    X_tmp = V @ X_tmp @ Vh
-
-    return Cref + X + X_tmp
+    LX = xp.asarray(solve_continuous_lyapunov(Cref, X))
+    return Cref + X + LX @ Cref @ LX
```
qbarthelemy, line 241: @gabelstein, are you ok with this simplification?

gabelstein: formally yes, but may be slower on GPU as the scipy part needs to be shifted to CPU

qbarthelemy: Scipy supports array-api backends, including PyTorch.
https://docs.scipy.org/doc/scipy-1.12.0/dev/api-dev/array_api.html

gabelstein: Ah, I did not know this existed. But the function still seems not supported by GPU at this point. Not sure how relevant this is generally, but using solve_continuous_lyapunov is basically just saving on code lines with no added improvement otherwise and the downside of not being able to use GPU for this function (at least not until scipy implements this exact function for GPU). 

Again, not sure how relevant gpu computation is for this.

tests/test_geometry_tangentspace.py
```diff
@@ -312,6 +311,22 @@ def test_innerproduct_broadcasting(finnerproduct, backend, get_mats):
     assert finnerproduct(C, D, Cref).shape == (n_sets, n_matrices)  # 4D arrays
 
 
+@pytest.mark.parametrize("kindX, kindC", [("sym", "spd"), ("herm", "hpd")])
+@pytest.mark.parametrize("metric", metrics)
+def test_innerproduct_property_distance(kindX, kindC, metric, get_mats):
+    """Test d(C,exp_C(X))^2 = g_C(X,X) locally"""
```
qbarthelemy, line 308: This test is KO for Hermitian case used with Wasserstein metric.
@gabelstein , do you have an opinion on why?

PR #457, opened 2026-06-26
Make stateless transformers report as fitted (fix #456)

### LINKED ISSUE #456: Stateless transformers (e.g. Covariances) fail check_is_fitted, breaking sklearn Pipeline(transform_input=...)
### Summary

pyRiemann's stateless transformers (e.g. `Covariances`) have a no-op `fit` that sets no fitted attribute and do not define `__sklearn_is_fitted__`. As a result, after `.fit(...)` they fail scikit-learn's `check_is_fitted`, which breaks scikit-learn ≥ 1.6's `Pipeline(transform_input=...)`.

### Why it matters

`Pipeline(transform_input=[...])` (sklearn ≥ 1.6) transforms an auxiliary `fit` input through the pipeline's steps before the step that consumes it (e.g. passing a validation set, or — in transfer learning — an unlabeled target slice that a mid-pipeline alignment step like RPA needs as *covariances*, not raw epochs). Internally it calls `transform` on the fitted sub-pipeline, which runs `check_is_fitted`. For a stateless step such as `Covariances`, that raises `NotFittedError`, so the whole pattern is unusable with pyRiemann.

### Minimal reproducer

```python
import numpy as np
from sklearn.utils.validation import check_is_fitted
from pyriemann.estimation import Covariances

X = np.random.randn(10, 4, 100)
cov = Covariances("oas").fit(X)
check_is_fitted(cov)
# NotFittedError: This Covariances instance is not fitted yet.
```

And the resulting `transform_input` failure:

```python
import numpy as np
from sklearn import config_context
from sklearn.base import BaseEstimator, TransformerMixin
from sklearn.pipeline import make_pipeline
from pyriemann.estimation import Covariances

class Consumer(TransformerMixin, BaseEstimator):
    def fit(self, X, y=None, aux=None):
        return self
    def transform(self, X):
        return X

X = np.random.randn(10, 4, 100); y = np.zeros(10)
with config_context(enable_metadata_routing=True):
    pipe = make_pipeline(
        Covariances("oas"),
        Consumer().set_fit_request(aux=True),
        transform_input=["aux"],
    )
    pipe.fit(X, y, aux=X)
# NotFittedError raised from inside Pipeline._get_metadata_for_step ->
# sub_pipeline.transform(...) -> check_is_fitted(Covariances)
```

### Suggested fix

Have stateless transformers report as fitted, e.g.:

```python
def __sklearn_is_fitted__(self):
    return True
```

(or set a trivial fitted attribute such as `self.fitted_ = True` in `fit`). This is a one-liner per stateless transformer (`Covariances`, and likely others such as the coherence/kernel estimators that are similarly stateless).

### Versions

pyRiemann 0.12.dev, scikit-learn 1.10.dev (also reproduces on sklearn ≥ 1.6 where `transform_input` was introduced).

### Context

Surfaced while integrating Riemannian transfer-learning pipelines (RPA after `Covariances`) into MOABB's cross-subject evaluation, where `transform_input` lets the alignment step receive the unlabeled target in the same representation as `X`.

bruAristimunha:
Closes #456.

## Problem

Stateless transformers (`Covariances`, `CrossSpectra`, `TimeDelayCovariances`, `Kernels`, `Shrinkage`, `TLDummy`, and their subclasses) have a no-op `fit` and define no fitted attribute nor `__sklearn_is_fitted__`. After `.fit(...)` they fail scikit-learn's `check_is_fitted`, raising `NotFittedError`.

This breaks scikit-learn ≥ 1.6's `Pipeline(transform_input=...)`, which transforms an auxiliary `fit` input through the fitted sub-pipeline (calling `transform` → `check_is_fitted`) before the consuming step — e.g. feeding an unlabeled target slice to a mid-pipeline alignment step (RPA after `Covariances`) in transfer-learning pipelines.

## Fix

Add `__sklearn_is_fitted__(self): return True` to each stateless base class (the maintainer-approved suggestion in the issue). Subclasses (`BlockCovariances`, `CoSpectra`, `Coherences`) inherit it via MRO.

## Tests

- `tests/test_estimation.py::test_stateless_check_is_fitted` — `check_is_fitted` passes after `fit` for all stateless estimation transformers and subclasses.
- `tests/test_transfer.py::test_tldummy` — extended with a `check_is_fitted` assertion for `TLDummy`.

Both reproducers from the issue (bare `check_is_fitted` and the `Pipeline(transform_input=...)` pattern) now pass. Full `test_estimation.py` and `test_transfer.py` suites are green.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,6 +10,11 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.12.dev
 ---------
 
+- Add ``__sklearn_is_fitted__`` to stateless transformers, so they pass
+  scikit-learn's ``check_is_fitted`` after ``fit`` and can be used inside
+  ``Pipeline(transform_input=...)`` (introduced in scikit-learn 1.6).
+  :pr:`457` by :user:`bruAristimunha`
+
 - Deprecate ``covariances_X`` and ``cospectrum``.
   :pr:`442` by :user:`qbarthelemy`
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,11 +10,6 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.12.dev
 ---------
 
-- Add ``__sklearn_is_fitted__`` to stateless transformers, so they pass
-  scikit-learn's ``check_is_fitted`` after ``fit`` and can be used inside
-  ``Pipeline(transform_input=...)`` (introduced in scikit-learn 1.6).
-  :pr:`457` by :user:`bruAristimunha`
-
 - Deprecate ``covariances_X`` and ``cospectrum``.
   :pr:`442` by :user:`qbarthelemy`
 
@@ -65,6 +60,11 @@ v0.12.dev
 - Add log-Cholesky inner product for SPD matrices :func:`pyriemann.geometry.tangentspace.innerproduct_logchol`,
   and improve other inner products.
   :pr:`450` by :user:`qbarthelemy`
+
+- Add ``__sklearn_is_fitted__`` to stateless transformers, so they pass
+  scikit-learn's ``check_is_fitted`` after ``fit`` and can be used inside
+  ``Pipeline(transform_input=...)`` (introduced in scikit-learn 1.6).
+  :pr:`457` by :user:`bruAristimunha`
 
 v0.11 (April 2026)
 ------------------
--- a/tests/test_estimation.py
+++ b/tests/test_estimation.py
@@ -404,7 +404,7 @@ def test_shrinkage(shrinkage, kind, get_mats):
     ],
 )
 def test_stateless_check_is_fitted(estimator, n_dim, kind, get_mats):
-    """Stateless transformers report as fitted after ``fit`` (issue #456).
+    """Stateless transformers report as fitted after ``fit``.
 
     They have a no-op ``fit`` and set no fitted attribute, so without
     ``__sklearn_is_fitted__`` scikit-learn's ``check_is_fitted`` raised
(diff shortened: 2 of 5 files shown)


doc/whatsnew.rst
```diff
@@ -10,6 +10,11 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.12.dev
 ---------
 
+- Add ``__sklearn_is_fitted__`` to stateless transformers, so they pass
```
qbarthelemy: To maintain the file's consistency, could you place that at the bottom of the 0.12 contributions?

tests/test_estimation.py
```diff
@@ -387,3 +388,28 @@ def test_shrinkage(shrinkage, kind, get_mats):
...
+        (CrossSpectra(), [3, 100], "real"),
+        (CoSpectra(), [3, 100], "real"),
+        (Coherences(), [3, 100], "real"),
+        (TimeDelayCovariances(), [3, 100], "real"),
+        (Kernels(), [3, 100], "real"),
+        (Shrinkage(), 3, "spd"),
+    ],
+)
+def test_stateless_check_is_fitted(estimator, n_dim, kind, get_mats):
+    """Stateless transformers report as fitted after ``fit`` (issue #456).
```
qbarthelemy:
```suggestion
    """Stateless transformers report as fitted after ``fit``.
```

tests/test_transfer.py
```diff
@@ -121,6 +122,7 @@ def test_tldummy(rndstate, space):
 
     dum = TLDummy()
     dum.fit(X, y_enc)
+    check_is_fitted(dum)  # issue #456: stateless step must report as fitted
```
qbarthelemy:
```suggestion
    check_is_fitted(dum)
```

qbarthelemy (changes requested):
Thx for this nice correction @bruAristimunha !

bruAristimunha:
Thanks for the review @qbarthelemy! Addressed all three comments in e1d3831: moved the changelog entry to the bottom of the 0.12 section, and simplified the test docstring and assertion. Ready for another look when you have a moment.

PR #464, opened 2026-07-07
Add transductive re-estimation to TLCenter for LOSO transfer

gcattan:
Add a transductive option that recenters a batch to its own recomputed mean instead, following a old idea in Barachant et al. (2013) for inter-session adaptation.

cc: @aquemy

PR #467, opened 2026-07-09
Revert "Add transductive re-estimation to TLCenter for LOSO transfer"

gcattan:
Reverts pyRiemann/pyRiemann#464


qbarthelemy:
The PR was merged quickly, but it was functional.
So, I think it's better to open a new PR, improving tests, rather than reverting it. Thx.

PR #313, opened 2024-07-11
Add ajd_method parameter to CSP

qbarthelemy:
This PR:
- adds a parameter `ajd_method` to `CSP`, solving partially #49;
- improves `ajd` methods, replacing while loop by for-else loop;
- completes tests.

PR #474, opened 2026-07-21
Rename rjd into jade and transpose outputed diagonalizer

binggao1230:
`CSP(ajd_method="rjd")` on 3+ classes returns spatial filters that do not jointly diagonalize the class covariances.

The three AJD methods return the diagonalizer in two different conventions: `rjd` gives `D = V.T C V` (the joint eigenvectors are the **columns** of `V`), while `ajd_pham` and `uwedge` give `D = V C V.T` (eigenvectors are the **rows**). The multiclass branch of `CSP.fit` unconditionally does `evecs = evecs.T`, which is only correct for the row convention, so for `rjd` it selects the wrong vectors.

This slipped because the 2-class path uses `scipy.linalg.eigh` and never touches `ajd_method`, and `test_csp` only exercises `ajd_method` with 2 classes while asserting shapes — the multiclass `rjd` path was never covered. `ajd_method` was added to `CSP` in #313 but `rjd`'s convention was never reconciled in the consumer.

Repro:

```python
import numpy as np
from pyriemann.spatialfilters import CSP

n = 6
Q, _ = np.linalg.qr(np.random.randn(n, n))
Cs = [Q @ np.diag(np.random.uniform(0.5, 5, n)) @ Q.T for _ in range(3)]  # jointly diagonalizable
X = np.array([C for C in Cs for _ in range(10)])
y = np.array([c for c in range(3) for _ in range(10)])

for m in ("ajd_pham", "uwedge", "rjd"):
    W = CSP(nfilter=n, metric="euclid", ajd_method=m).fit(X, y).filters_
    r = [W @ C @ W.T for C in Cs]
    off = sum(np.sum((M - np.diag(np.diag(M))) ** 2) for M in r)
    dia = sum(np.sum(np.diag(np.diag(M)) ** 2) for M in r)
    print(m, np.sqrt(off / dia))
# ajd_pham 4e-11, uwedge 4e-16, rjd 4e-01   <- rjd filters do not diagonalize
```

Fix: skip the transpose for `rjd` (matching both the `"rjd"` string and the `rjd` callable, since `ajd_method` accepts a callable). `ajd_pham`/`uwedge` output is byte-identical to before.

The per-method convention is intentional and pinned by `test_geometry_ajd` (which also asserts `ajd(method="rjd")` returns exactly what `rjd()` returns), so the fix belongs in the consumer, not in `rjd`. I checked the other AJD consumers — `SPoC`/`AJDC` and the ALE mean hardcode `ajd_pham` — so `CSP` is the only one exposed to this. Added a multiclass regression test asserting the fitted filters jointly diagonalize the class means for every `ajd_method`; it fails on `rjd` (string and callable) before the fix and passes after.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,6 +10,13 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.13.dev
 ---------
 
+- Fix multiclass :class:`pyriemann.spatialfilters.CSP` with
+  ``ajd_method="rjd"``: ``rjd`` returns its diagonalizer in the transposed
+  convention of ``ajd_pham`` and ``uwedge``, so ``fit`` transposed it the wrong
+  way and produced spatial filters that did not jointly diagonalize the class
+  covariances.
+  :pr:`474` by :user:`gaoflow`
+
 - Update pyRiemann from Python 3.10 - 3.12 to 3.11 - 3.13.
   :pr:`462` by :user:`qbarthelemy`
 
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,13 +10,6 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.13.dev
 ---------
 
-- Fix multiclass :class:`pyriemann.spatialfilters.CSP` with
-  ``ajd_method="rjd"``: ``rjd`` returns its diagonalizer in the transposed
-  convention of ``ajd_pham`` and ``uwedge``, so ``fit`` transposed it the wrong
-  way and produced spatial filters that did not jointly diagonalize the class
-  covariances.
-  :pr:`474` by :user:`gaoflow`
-
 - Update pyRiemann from Python 3.10 - 3.12 to 3.11 - 3.13.
   :pr:`462` by :user:`qbarthelemy`
 
@@ -36,6 +29,8 @@ v0.13.dev
 - Add :func:`pyriemann.geometry.tangentspace.transport_wasserstein` for parallel
   transport with the Bures-Wasserstein metric.
   :pr:`476` by :user:`AmitSubhash`
+- Deprecate ``rjd`` renamed into :func:`pyriemann.geometry.ajd.jade`, and transpose outputed diagonalizer.
+  :pr:`474` by :user:`gaoflow`
 
 v0.12 (July 2026)
 -----------------
@@ -436,7 +431,7 @@ v0.4 (Feb 2023)
   then used for ``transform()`` as well as for ``inverse_transform()``.
   :pr:`195` by :user:`qbarthelemy`
 
-- Enhance AJD: add ``init`` to :func:`pyriemann.geometry.ajd.ajd_pham` and :func:`pyriemann.geometry.ajd.rjd`,
+- Enhance AJD: add ``init`` to :func:`pyriemann.geometry.ajd.ajd_pham` and :func:`pyriemann.geometry.ajd.jade`,
   add ``warm_restart`` to :class:`pyriemann.spatialfilters.AJDC`.
   :pr:`196` by :user:`qbarthelemy`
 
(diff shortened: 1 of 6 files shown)


qbarthelemy (changes requested):
Good catch @gaoflow !

Indeed, the fact that `rjd`/JADE does not follow the usual convention causes many problems,
like the rising curve of the 2nd figure of the AJD benchmark:
https://github.com/benchopt/benchmark_jointdiag/pull/9#issuecomment-2215184981

To avoid placing the burden of handling this inconsistency on the user,
I would prefer to correct `rjd`, transposing the output `V`,
and take the opportunity to deprecate `rjd` and rename it to `jade`.

Can you update your PR with these changes?
New test `test_csp_multiclass_diagonalization` should be kept.

qbarthelemy:
I've pushed the correction. Does that work for you, @gaoflow ?

binggao1230:
Yes, that works — and it's the better fix. Correcting the convention in the function is right; my version compensated at the one call site and would have left `jade` still disagreeing with `ajd_pham`/`uwedge` for anyone calling it directly.

Checked it on exactly-jointly-diagonalizable input (`C_k = A diag(d_k) A^T`, so a correct AJD recovers `A` and `V C_k V^T` is exactly diagonal), measuring max off-diagonal / max diagonal:

| | before | after |
|---|---|---|
| `rjd` / `jade` | 3.3e-01 | 1.2e-09 |
| `ajd_pham` | 2.3e-14 | 2.3e-14 |

So `jade` now agrees with the documented `D = V X V^T`. The `rjd` shim still imports and raises the `DeprecationWarning`. `tests/test_geometry_ajd.py` + `tests/test_spatialfilters.py`: 130 passed.

I also rebased onto master to clear the conflict — it was only the `whatsnew.rst` entry colliding with #455, both entries kept. Your commit is unchanged apart from the rebase, but the hash moved to `df286cb`, so re-pull before touching the branch again.

binggao1230:
One gap I hit while checking: `ajd_functions` lost its `"rjd"` key, so `ajd(X, method="rjd")` and `CSP(ajd_method="rjd")` now raise `ValueError: Unknown function name 'rjd'. Must be one of ajd_pham jade uwedge`. The `rjd()` function keeps a deprecation window until 0.15.0, but the string doesn't get one, and `ajd_method` has been public since 0.7. Keeping `"rjd": rjd` in the dict would route it through the shim and emit the DeprecationWarning instead of erroring. Happy to push that if you want it.

Otherwise nothing else consumes `jade` — inside the package the only readers of an AJD result are `CSP` via `ajd()` and two direct `ajd_pham` calls (`AJDC`, `mean.py`), so the transpose doesn't reach anything else.

Note the workflows on the rebased head are sitting at `action_required`, so nothing has run yet.

pyriemann/geometry/ajd.py
```diff
@@ -352,7 +371,7 @@ def uwedge(X, *, init=None, eps=1e-7, n_iter_max=100):
 
 ajd_functions = {
     "ajd_pham": ajd_pham,
-    "rjd": rjd,
+    "jade": jade,
```
qbarthelemy, line 375: @gaoflow , you are right: `"rjd"` key must be kept.
```suggestion
    "rjd": rjd,
    "jade": jade,
```

binggao1230:
Applied — your `"rjd": rjd` commit is on the branch (kept intact through the rebase, now `311eae2`). `ajd(X, method="rjd")` and `CSP(ajd_method="rjd")` route through the shim again and emit the DeprecationWarning instead of erroring. 130 tests pass; rebased onto master (whatsnew conflict with #476 resolved, both entries kept).

doc/whatsnew.rst
```diff
@@ -29,6 +29,8 @@ v0.13.dev
 - Add :func:`pyriemann.geometry.tangentspace.transport_wasserstein` for parallel
   transport with the Bures-Wasserstein metric.
   :pr:`476` by :user:`AmitSubhash`
+- Deprecate ``rjd`` renamed into :func:`pyriemann.geometry.ajd.jade`, and transpose outputed diagonalizer.
```
qbarthelemy, line 33:
```suggestion

- Deprecate ``rjd`` renamed into :func:`pyriemann.geometry.ajd.jade`, and transpose outputed diagonalizer.
```

qbarthelemy (approved):
Thx @gaoflow for this contribution!

Issue #465, opened 2026-07-07
Update array API support for geometry


qbarthelemy:
`geometry._backend` contains temporary functions intended to be replaced by those from the official libraries.

Related to #433 and #444

### Update with respect to array-api

- [ ] `diag_indices`, `tril_indices` and `triu_indices`
https://github.com/data-apis/array-api-extra/pull/692
Resolvable updating [array-api-extra>=0.11.1](https://github.com/data-apis/array-api-extra/releases#release-v0.11.1)
- [ ] `cov` and `corrcoef` using a custom function `apply_xp_cov`
https://github.com/data-apis/array-api-extra/pull/690
https://github.com/data-apis/array-api-extra/issues/687
Resolvable updating [array-api-extra>=0.11.2](https://github.com/data-apis/array-api-extra/releases/tag/v0.11.2)
- [ ] `nanmean`
https://github.com/data-apis/array-api-extra/pull/910
Resolvable updating [array-api-extra>=0.11.2](https://github.com/data-apis/array-api-extra/releases/tag/v0.11.2)
- [x] remove deprecation warning from Python 3.12
```DeprecationWarning: `xpx.expand_dims` is deprecated and will be removed in v1.0.0. `xp.expand_dims` with support for a tuple of ints in `axis` exists in the standard as of v2025.12. denom = expand_dims(denom, axis=tuple(range(denom.ndim, X.ndim)))```
Resolvable updating [array-api-compat>=1.14](https://data-apis.org/array-api-compat/changelog.html)
Solved by https://github.com/pyRiemann/pyRiemann/pull/478

### Update after scikit-learn conversion

- [ ] use `_add_to_diagonal` from `sklearn.utils._array_api`

General issue for `covariance` module
https://github.com/scikit-learn/scikit-learn/issues/33584
- [ ] `ledoit_wolf`
https://github.com/scikit-learn/scikit-learn/pull/33573
Resolvable updating sklearn>=WAITING_FOR_NEW_RELEASE
- [ ] `oas`
https://github.com/scikit-learn/scikit-learn/pull/33600
Resolvable updating sklearn>=WAITING_FOR_NEW_RELEASE
- [ ] `fast_mcd`
TODO

AmitSubhash:
Is the `expand_dims` item still open? Happy to take it.

[...]

PR #433, opened 2026-03-27
Add array API support for NumPy/PyTorch backend transparency for utils

bruAristimunha:
## Summary

Add [Python Array API](https://data-apis.org/array-api/latest/) support to pyRiemann's core utility functions, enabling transparent execution on both NumPy and PyTorch backends. This follows the same approach used by [scikit-learn's array API adoption](https://labs.quansight.org/blog/array-api-scikit-learn-2026).

### Architecture

- **`_backend.py`** — Thin re-exports from `array-api-compat` and `array-api-extra`, plus device-aware index helpers (`diag_indices`, `tril_indices`, `triu_indices`) and conversion utilities (`to_numpy`, `from_numpy`)
- **`_fixes.py`** — Array-API-compatible reimplementations of sklearn covariance functions (`ledoit_wolf`, `oas`) for non-numpy backends. `fast_mcd` falls back to numpy (no torch implementation exists anywhere). This module can be removed when sklearn adds native support.
- **`_vectorize_nd`** — `batch_native` parameter: for numpy, loops over batch elements; for torch, calls the function directly (GPU parallelism, autograd support)

### What works on both NumPy and PyTorch

| Module | Functions | Notes |
|--------|-----------|-------|
| `base` | `expm`, `logm`, `sqrtm`, `invsqrtm`, `powm`, `nearest_sym_pos_def`, `ddexpm`, `ddlogm` | All via `xp.linalg.eigh` |
| `distance` | All 13 distance metrics | `riemann`/`thompson`: scipy path for numpy, `invsqrtm` path for torch |
| `mean` | All 16 mean functions | 6 broadcast natively, 5 batch-native for torch, 5 loop-only |
| `covariance` | `scm`, `sch`, `mest`, `cov`, `corr`, `lwf`, `oas` | `mcd` falls back to numpy |
| `geodesic` | All 7 geodesic functions | |
| `tangentspace` | All log/exp maps, transport, upper/unupper, innerproduct | |
| `ajd` | `rjd`, `ajd_pham`, `uwedge` | |
| `kernel` | `euclid`, `logeuclid`, `riemann` | |
| `median` | `median_euclid`, `median_riemann` | Partial torch support |

### Test results

- **NumPy**: 898 passed, 93 skipped
- **PyTorch**: 775 passed, 93 skipped, 16 failed (all in `median` — known)
- **Full project** (numpy only): 2413 passed, 0 failed

### Key design decisions

[...]

PR #475, opened 2026-07-23
Use array-api-extra cov() (PR #690) for batched covariance estimation

bruAristimunha:
## What

Route the `"cov"` covariance estimator through
[`array_api_extra.cov`](https://data-apis.org/array-api-extra/), using the
new parameters added in **[data-apis/array-api-extra#690](https://github.com/data-apis/array-api-extra/pull/690)**
(`axis` / `correction` / `fweights` / `aweights`).

This is a small, concrete downstream consumer of that PR — see the
[motivating comment](https://github.com/data-apis/array-api-extra/pull/690#issuecomment-5057787560).

Solve a part of https://github.com/pyRiemann/pyRiemann/issues/465

## Why

`pyriemann.geometry.covariance._cov` used to:

1. hand-translate `bias`/`ddof` → `correction` (the `_cov_kwargs_to_xp` /
   `apply_xp_cov` shim in `_backend.py`) to paper over the NumPy-vs-PyTorch
   keyword split, and
2. loop over every leading batch dimension in Python (`_vectorize_nd`).

`array_api_extra.cov` now does **both** upstream: it takes the array-API
standard `correction` keyword and estimates covariances for all leading
batch dimensions in a single call, across backends. So for the `"cov"`
path we can delegate and drop the local machinery:

```python
def _cov(X, **kwds):
    xp = get_namespace(X)
    cov = xpx_cov(X, axis=-1, **_cov_kwargs_to_xp(kwds))
    return xp.reshape(cov, (*X.shape[:-2], X.shape[-2], X.shape[-2]))
```

Observations are the time samples, hence `axis=-1`. The reshape restores
the two matrix axes that `cov()` collapses when `n_channels == 1`, so the
output stays `(..., n_channels, n_channels)` in every case.

`_corr` still uses `apply_xp_cov(xp.corrcoef, ...)` — array-api-extra has no
`corrcoef` — so the shim stays for that estimator.

## Behavior

Output is **unchanged** for existing usage (`n_channels >= 2`): `xpx.cov`'s
defaults (`axis=-1, correction=1`) match `numpy.cov`, verified bit-for-bit
against a per-matrix `np.cov` loop up to 5-D batches, on both NumPy and
PyTorch. The only new coverage is the `n_channels == 1` shape guard.

## Tests

- `tests/test_geometry_covariance.py`: **all green** on NumPy + PyTorch.
- Added `test_covariances_single_channel` locking the `(..., 1, 1)` shape.

## Dependency

Needs array-api-extra #690, not yet released. `setup.py` and
`requirements.txt` temporarily pin the branch:

```
array-api-extra @ git+https://github.com/bruAristimunha/array-api-extra@cov_parameters
```

Marked as **draft** — revert the pin to a released `>=` floor once #690 is
merged and published.

### DIFF
--- a/.github/workflows/deploy_ghpages.yml
+++ b/.github/workflows/deploy_ghpages.yml
@@ -118,6 +118,7 @@ jobs:
           docs-folder: "doc/"
           pre-build-command: |
             apt-get update
+            apt-get install -y git
             pip install -e .
             pip install -r doc/requirements.txt
         env:
--- a/pyriemann/geometry/covariance.py
+++ b/pyriemann/geometry/covariance.py
@@ -2,10 +2,11 @@ from functools import wraps
 import warnings
 
 from array_api_compat import array_namespace as get_namespace, device as xpd
-from array_api_extra import expand_dims
+from array_api_extra import cov as xpx_cov, expand_dims
 from scipy.stats import chi2
 
 from ._backend import (
+    _cov_kwargs_to_xp,
     apply_xp_cov,
     diag_indices,
     hann_window,
@@ -21,9 +22,17 @@ from .test import is_real_type, is_square
 
 
 def _cov(X, **kwds):
-    """Covariance matrix estimator."""
+    """Covariance matrix estimator, batched over leading dimensions.
+
+    Delegates to :func:`array_api_extra.cov`, which estimates covariances for
+    all leading batch dimensions in a single call, across array-API backends.
+    Time samples are the observations, hence ``axis=-1``.
+    """
     xp = get_namespace(X)
-    return apply_xp_cov(xp.cov, X, **kwds)
+    cov = xpx_cov(X, axis=-1, **_cov_kwargs_to_xp(kwds))
+    # array_api_extra.cov drops the two matrix axes when n_channels == 1;
+    # restore them so the output is always (..., n_channels, n_channels).
+    return xp.reshape(cov, (*X.shape[:-2], X.shape[-2], X.shape[-2]))
 
 
 def _corr(X, **kwds):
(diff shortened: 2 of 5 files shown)


bruAristimunha:
### Code review

Found 1 issue:

1. The default `"cov"` estimator silently breaks PyTorch autograd on batched input. Excluding `_cov` from `_vectorize_nd` feeds the full batched array into `array_api_extra.cov`, whose batched path runs `xp.asarray(m)`; `torch.asarray` defaults to `requires_grad=False`, detaching the graph. So `covariances(X)` with `X.requires_grad=True` and shape `(n_trials, n_channels, n_times)` returns `requires_grad=False`, and inside a larger graph `.backward()` runs but leaves `X.grad is None` (silent gradient loss). The 2-D path and `estimator="scm"` are unaffected. This is the autograd property that #433 added torch support to preserve, and no covariance autograd test currently guards it.

Root cause — the batched array now bypasses the per-slice `torch.cov` path:

https://github.com/pyRiemann/pyRiemann/blob/a6499c36a11cc5255c466889878729cd877b3ecf/pyriemann/geometry/covariance.py#L415-L418

Detaching call:

https://github.com/pyRiemann/pyRiemann/blob/a6499c36a11cc5255c466889878729cd877b3ecf/pyriemann/geometry/covariance.py#L29-L35

bruAristimunha:
Update: the root cause is fixed upstream in `array-api-extra` (the `cov_parameters` branch this PR already pins), commit bruAristimunha/array-api-extra@b6e9e1c352d3b7dbeecc5983f01ad9f48d7a5bab.

The generic `cov` path called `xp.asarray(m)` unconditionally; on torch that detaches gradients and mutates the input tensor in place. It is now guarded to run only for genuine non-array input, so arrays pass through untouched. A torch autograd regression test was added to `array-api-extra`'s `TestCov`.

With that branch, batched `covariances(X)` with `X.requires_grad=True` returns `requires_grad=True` and `X.grad` flows (verified locally; values still match `numpy.cov`). So the native-batching change in this PR no longer regresses autograd.

PR #476, opened 2026-07-23
Add `transport_wasserstein` for Bures-Wasserstein parallel transport

### LINKED ISSUE #418: Add parallel transport for Wasserstein metric
Add function `transport_wasserstein`, see Section 7.1.2 of Malago2018, [Wasserstein-Riemannian Geometry of Positive-definite Matrices](https://arxiv.org/pdf/1801.09269)

AmitSubhash:
Closes #418.

Adds Bures-Wasserstein parallel transport, integrating the transport ODE with a fixed-step RK4 scheme. Tests cover SPD/HPD inputs, batching, and the commuting closed form as an exact oracle.

Also fixes `innerproduct_wasserstein` for complex HPD inputs, which un-skips the HPD Wasserstein distance-property test.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -26,6 +26,10 @@ v0.13.dev
 - Add Wasserstein inner product for Hermitian matrices :func:`pyriemann.geometry.tangentspace.innerproduct_wasserstein`.
   :pr:`455` by :user:`qbarthelemy`
 
+- Add :func:`pyriemann.geometry.tangentspace.transport_wasserstein` for parallel
+  transport with the Bures-Wasserstein metric.
+  :pr:`476` by :user:`AmitSubhash`
+
 v0.12 (July 2026)
 -----------------
 
--- a/pyriemann/geometry/tangentspace.py
+++ b/pyriemann/geometry/tangentspace.py
@@ -1,6 +1,7 @@
 """Tangent space for SPD/HPD matrices."""
 
 import math
+import operator
 
 from array_api_compat import (
     array_namespace as get_namespace,
@@ -986,10 +987,10 @@ def innerproduct_wasserstein(X, Y, Cref):
         pp. 165-191.
     """
     xp = check_matrix_pair(X, Cref, require_square=True)
-    LX = xp.asarray(solve_continuous_lyapunov(Cref, xp.conj(X)))
+    LX = xp.asarray(solve_continuous_lyapunov(Cref, X))
     if Y is None:
         Y = X
-    G = 0.5 * xp.einsum("...ij,...ij->...", LX, Y).real
+    G = 0.5 * xp.einsum("...ij,...ji->...", LX, Y).real
     return _prepare_output(G, xp)
 
 
@@ -1311,15 +1312,112 @@ def transport_riemann(X, A, B):
     return X_new
 
 
+def transport_wasserstein(X, A, B, n_steps=100):
+    r"""Parallel transport for Wasserstein metric.
+
+    The parallel transport of matrices :math:`\mathbf{X}` in tangent space
+    from an initial SPD/HPD matrix :math:`\mathbf{A}` to a final SPD/HPD
+    matrix :math:`\mathbf{B}` according to the Levi-Civita connection of the
+    Bures-Wasserstein metric, described in Section 7.5 of [1]_.
+
+    Contrary to the other metrics, Bures-Wasserstein parallel transport has no
+    closed form in the general case [1]_ [2]_: it is defined by a linear
+    ordinary differential equation along the Wasserstein geodesic, integrated
+    here with a fixed-step Runge-Kutta scheme of order 4. A closed form exists
+    only when :math:`\mathbf{A}` and :math:`\mathbf{B}` commute.
+
+    Warning: this function must be applied to matrices :math:`\mathbf{X}`
+    already projected in tangent space with a logarithmic map at
+    :math:`\mathbf{A}`, not to SPD/HPD matrices in manifold.
+
+    Parameters
+    ----------
+    X : ndarray, shape (..., n, n)
+        Symmetric/Hermitian matrices in tangent space at A.
+    A : ndarray, shape (n, n)
+        Initial SPD/HPD matrix.
+    B : ndarray, shape (n, n)
+        Final SPD/HPD matrix.
+    n_steps : int, default=100
+        Number of Runge-Kutta steps used to integrate the transport equation.
+        Must be a positive integer; larger values increase accuracy at the
+        cost of computation.
+
+    Returns
+    -------
+    X_new : ndarray, shape (..., n, n)
+        Matrices in tangent space transported from A to B.
+
+    Notes
+    -----
+    .. versionadded:: 0.13
+
+    See Also
+    --------
+    transport
+
+    References
+    ----------
+    .. [1] `Wasserstein Riemannian geometry of Gaussian densities
+        <https://link.springer.com/article/10.1007/s41884-018-0014-4>`_
+        L. Malagò, L. Montrucchio, G. Pistone. Information Geometry, 2018, 1,
+        pp. 137–179.
+    .. [2] `O(n)-invariant Riemannian metrics on SPD matrices
+        <https://www.sciencedirect.com/science/article/pii/S0024379522004360>`_
+        Y. Thanwerdas & X. Pennec. Linear Algebra and its Applications, 2023.
+    """
+    try:
+        n_int = operator.index(n_steps)
+    except TypeError:
+        n_int = None
+    if isinstance(n_steps, bool) or n_int is None or n_int < 1:
+        raise ValueError(
+            f"n_steps must be a positive integer, got {n_steps!r}."
+        )
+    n_steps = n_int
+    xp = get_namespace(X, A, B)
+    n = A.shape[-1]
+    eye = xp.eye(n, dtype=A.dtype, device=xpd(A))
+
+    A12, A12inv = sqrtm(A), invsqrtm(A)
+    T = A12inv @ sqrtm(A12 @ B @ A12) @ A12inv
+    T = (T + ctranspose(T)) / 2
+    K = T - eye
+    AK, KA = A @ K, K @ A
+
+    # generator of the transported field: X0(t) solves X0 gamma + gamma X0 = X
+    X0 = xp.asarray(solve_continuous_lyapunov(A, X))
+
+    def _deriv(t, X0):
+        Mt = (1 - t) * eye + t * T
+        gamma = Mt @ A @ Mt
+        rhs = Mt @ AK @ X0 + X0 @ KA @ Mt
+        rhs = (rhs + ctranspose(rhs)) / 2
+        return -xp.asarray(solve_continuous_lyapunov(gamma, rhs))
+
+    h = 1 / n_steps
+    for i in range(n_steps):
+        t = i * h
+        k1 = _deriv(t, X0)
+        k2 = _deriv(t + h / 2, X0 + (h / 2) * k1)
+        k3 = _deriv(t + h / 2, X0 + (h / 2) * k2)
+        k4 = _deriv(t + h, X0 + h * k3)
+        X0 = X0 + (h / 6) * (k1 + 2 * k2 + 2 * k3 + k4)
+
+    X_new = B @ X0 + X0 @ B
+    return (X_new + ctranspose(X_new)) / 2
+
+
 transport_functions = {
     "euclid": transport_euclid,
     "logchol": transport_logchol,
     "logeuclid": transport_logeuclid,
     "riemann": transport_riemann,
+    "wasserstein": transport_wasserstein,
 }
 
 
-def transport(X, A, B, metric="riemann"):
+def transport(X, A, B, metric="riemann", **kwargs):
     r"""Parallel transport according to a specified metric.
 
     Parallel transport of matrices :math:`\mathbf{X}` in tangent space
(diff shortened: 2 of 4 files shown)


pyriemann/geometry/tangentspace.py
```diff
@@ -1261,15 +1262,127 @@ def transport_riemann(X, A, B):
...
+    eye = xp.eye(n, dtype=A.dtype, device=xpd(A))
+
+    A12, A12inv = sqrtm(A), invsqrtm(A)
+    T = A12inv @ sqrtm(A12 @ B @ A12) @ A12inv
+    T = (T + ctranspose(T)) / 2
+    K = T - eye
+    AK, KA = A @ K, K @ A
+
+    # generator of the transported field: X0(t) solves X0 gamma + gamma X0 = X
+    X0 = _lyapunov_wasserstein(X, A)
```
qbarthelemy: You can use `solve_continuous_lyapunov` from Scipy, see https://github.com/pyRiemann/pyRiemann/pull/455/changes#diff-dd03aed44847b41f678b71d0bd8005ce87ae013425eb77e1805c5775fc0a82e6R240
```suggestion
    X0 = xp.asarray(solve_continuous_lyapunov(X, A)
```

tests/test_geometry_tangentspace.py
qbarthelemy, line 476: `innerproduct_wasserstein` is now available in the last master.
So, if you add "wasserstein" here, `test_transport_wasserstein_properties` is useless.
```suggestion
    "riemann",
    "wasserstein",
```

doc/whatsnew.rst
```diff
@@ -10,6 +10,10 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.13.dev
 ---------
 
+- Add :func:`pyriemann.geometry.tangentspace.transport_wasserstein` for parallel
```
qbarthelemy: For consistency, your contribution must be placed at the bottom of 0.13 section.

qbarthelemy (changes requested):
Thanks @AmitSubhash for this nice contribution!

AmitSubhash:
Thanks for the review, @qbarthelemy! Rebased on current `master` and addressed all three points:

1. **`solve_continuous_lyapunov`** — dropped the custom Lyapunov helper and use SciPy in both transport steps, consistent with `exp_map_wasserstein`. One small correction to the suggested snippet: the call needs the argument order `solve_continuous_lyapunov(A, X)` (reference first), since the equation being solved is `X0·A + A·X0 = X`.
2. **Tests** — added `"wasserstein"` to `test_transport_properties` (and to `test_transport_broadcasting`) and removed the dedicated `test_transport_wasserstein_properties`. One necessary tweak: the composition check now builds its waypoints with `geodesic(..., metric=metric)`. Bures-Wasserstein transport is path-dependent, so it needs Wasserstein geodesic waypoints; this is a no-op for the flat/affine-invariant metrics.
3. **whatsnew** — moved to the bottom of the 0.13 section.

While wiring `"wasserstein"` into the shared isometry test, I ran into a complex-Hermitian (HPD) issue in `innerproduct_wasserstein` from #455: with `solve_continuous_lyapunov(Cref, conj(X))` and a same-index (`"...ij,...ij"`) contraction it departs from the Bures-Wasserstein metric for HPD inputs (it agrees for real SPD). I switched it to `solve_continuous_lyapunov(Cref, X)` with a trace contraction (`"...ij,...ji"`), which matches both the Eq. (6) oracle and the eigenbasis Bures metric to machine precision, and preserves transport isometry for HPD. That let me un-skip the previously-skipped HPD Wasserstein distance-property test and parameterize the inner-product oracle over SPD/HPD. Happy to split this `innerproduct_wasserstein` fix into a separate PR if you'd prefer to keep this one transport-only.

One note on backends: like `exp_map_wasserstein` and `innerproduct_wasserstein`, `transport_wasserstein` routes its Lyapunov solves through SciPy, so it works on NumPy and CPU PyTorch but shares the family's current CPU/no-autograd limitation (it raises under `requires_grad`/CUDA). If you'd like the Wasserstein family to support autograd/GPU, I'm glad to add a backend-native (`eigh`) path for the PyTorch backend.

Full test suite is green and `flake8` is clean.

doc/whatsnew.rst
AmitSubhash: Done!

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,67 @@ def test_transport_riemann_vs_whitening(get_mats):
     Tt = transport(T, M, np.eye(n_channels), metric="riemann")
     Xt = exp_map_riemann(Tt, np.eye(n_channels), Cm12=True)
     assert Xw == approx(Xt)
+
+
+@pytest.mark.numpy_only
+@pytest.mark.parametrize("kind", ["spd", "hpd"])
```
qbarthelemy:
```suggestion
@pytest.mark.parametrize("kindX, kindQ", [("sym", "orth"), ("herm", "unit")])
```

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,67 @@ def test_transport_riemann_vs_whitening(get_mats):
...
+def test_transport_wasserstein_commuting(kind, get_mats, rndstate):
+    """BW transport has a closed form for commuting endpoints."""
+    n_channels = 4
+    if kind == "spd":
+        Q = np.linalg.qr(rndstate.randn(n_channels, n_channels))[0]
+    else:
+        Q = np.linalg.qr(
+            rndstate.randn(n_channels, n_channels)
+            + 1j * rndstate.randn(n_channels, n_channels)
+        )[0]
```
qbarthelemy:
```suggestion
    Q = get_mats(1, n_channels, kind=kindQ)[0]
```

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,67 @@ def test_transport_riemann_vs_whitening(get_mats):
...
+    d = rndstate.uniform(0.5, 3, n_channels)
+    delta = rndstate.uniform(0.5, 3, n_channels)
+    A = Q @ np.diag(d) @ Q.conj().T
+    B = Q @ np.diag(delta) @ Q.conj().T
+    X = get_mats(6, n_channels, "sym" if kind == "spd" else "herm")
```
qbarthelemy:
```suggestion
    X = get_mats(6, n_channels, kind=kindX)
```

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,67 @@ def test_transport_riemann_vs_whitening(get_mats):
...
+    Xr = Q.conj().T @ X @ Q
+    fac = np.sqrt(np.add.outer(delta, delta) / np.add.outer(d, d))
+    expected = Q @ (fac * Xr) @ Q.conj().T
+    assert Xt == approx(expected)
+
+
+@pytest.mark.numpy_only
+@pytest.mark.parametrize("n_steps", [0, -1, 1.5, True])
+def test_transport_wasserstein_invalid_n_steps(n_steps, get_mats):
+    """transport_wasserstein rejects non-positive-integer n_steps (#418)."""
```
qbarthelemy:
```suggestion
    """transport_wasserstein rejects non-positive-integer n_steps."""
```

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,67 @@ def test_transport_riemann_vs_whitening(get_mats):
...
+    A, B = get_mats(2, 3, "spd")
+    X = get_mats(1, 3, "sym")[0]
+    out_np = transport_wasserstein(X, A, B, n_steps=np.int64(20))
+    out_py = transport_wasserstein(X, A, B, n_steps=20)
+    assert out_np == approx(out_py)
+
+
+@pytest.mark.numpy_only
+def test_transport_wasserstein_dispatch_forwards_n_steps(get_mats):
+    """transport(metric='wasserstein') forwards n_steps (#418)."""
```
qbarthelemy:
```suggestion
    """transport(metric='wasserstein') forwards n_steps."""
```

qbarthelemy (changes requested):
Thank you for fixing inner product for Hermitian matrices in this PR!

For `solve_continuous_lyapunov` on PyTorch CPU, this should be fixed on Scipy side.

AmitSubhash:
Thanks @qbarthelemy, all five suggestions applied (the `Q` one read `kinQ`, so I used `kindQ`). Full suite green, flake8 clean. Noted on `solve_continuous_lyapunov` being upstream, I will raise it there ( tho i don't think they;re willing to make that change).

Really appreciate the thorough reviews and the time you have put into this:) Very grateful!

pyriemann/geometry/tangentspace.py
```diff
@@ -1311,15 +1312,112 @@ def transport_riemann(X, A, B):
...
+        Y. Thanwerdas & X. Pennec. Linear Algebra and its Applications, 2023.
+    """
+    try:
+        n_int = operator.index(n_steps)
+    except TypeError:
+        n_int = None
+    if isinstance(n_steps, bool) or n_int is None or n_int < 1:
+        raise ValueError(
+            f"n_steps must be a positive integer, got {n_steps!r}."
+        )
```
qbarthelemy: Parameter validation shouldn't need a try-catch block. What do you think about that?   
```suggestion
    if not isinstance(n_steps, (int, np.int64)) or (n_steps <= 0):
        raise ValueError(
            f"n_steps must be a positive integer, got {n_steps!r}."
        )
```

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,61 @@ def test_transport_riemann_vs_whitening(get_mats):
...
     Xt = exp_map_riemann(Tt, np.eye(n_channels), Cm12=True)
     assert Xw == approx(Xt)
+
+
+@pytest.mark.numpy_only
+@pytest.mark.parametrize("kindX, kindQ", [("sym", "orth"), ("herm", "unit")])
+def test_transport_wasserstein_commuting(kindX, kindQ, get_mats, rndstate):
+    """BW transport has a closed form for commuting endpoints."""
+    n_channels = 4
+    Q = get_mats(1, n_channels, kind=kindQ)[0]
```
qbarthelemy, line 544: Q^H should be computed once:
```suggestion
    Q = get_mats(1, n_channels, kind=kindQ)[0]
    Qh = Q.conj().T
```

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,61 @@ def test_transport_riemann_vs_whitening(get_mats):
     Tt = transport(T, M, np.eye(n_channels), metric="riemann")
     Xt = exp_map_riemann(Tt, np.eye(n_channels), Cm12=True)
     assert Xw == approx(Xt)
+
+
+@pytest.mark.numpy_only
+@pytest.mark.parametrize("kindX, kindQ", [("sym", "orth"), ("herm", "unit")])
+def test_transport_wasserstein_commuting(kindX, kindQ, get_mats, rndstate):
+    """BW transport has a closed form for commuting endpoints."""
```
qbarthelemy: Can you add a reference for this?

pyriemann/geometry/tangentspace.py
```diff
@@ -1311,15 +1312,112 @@ def transport_riemann(X, A, B):
...
+
+    Parameters
+    ----------
+    X : ndarray, shape (..., n, n)
+        Symmetric/Hermitian matrices in tangent space at A.
+    A : ndarray, shape (n, n)
+        Initial SPD/HPD matrix.
+    B : ndarray, shape (n, n)
+        Final SPD/HPD matrix.
+    n_steps : int, default=100
```
qbarthelemy: Isn't the default value of 100 too costly?
Would 50 or 10 be sufficient?

pyriemann/geometry/tangentspace.py
AmitSubhash: Agreed, there was a better way to do it - numbers.Integral does the same job

tests/test_geometry_tangentspace.py
AmitSubhash: Done.

tests/test_geometry_tangentspace.py
AmitSubhash: Added: Thanwerdas & Pennec 2023, already reference [2], Table 7, proved in Appendix A.

pyriemann/geometry/tangentspace.py
AmitSubhash: It is, 100 was a conservative choice thana measured one. Updated it to 50. (10 could be too less, I could run some tests if you;d like that)

tests/test_geometry_tangentspace.py
```diff
@@ -529,3 +531,67 @@ def test_transport_riemann_vs_whitening(get_mats):
...
     Xt = exp_map_riemann(Tt, np.eye(n_channels), Cm12=True)
     assert Xw == approx(Xt)
+
+
+@pytest.mark.numpy_only
+@pytest.mark.parametrize("kindX, kindQ", [("sym", "orth"), ("herm", "unit")])
+def test_transport_wasserstein_commuting(kindX, kindQ, get_mats, rndstate):
+    """BW transport has a closed form for commuting endpoints.
+
+    Table 7 of ref [2] in transport_wasserstein, proved in its Appendix A.
```
qbarthelemy:
```suggestion
    Table 7 of [Thanwerdas2023], proved in its Appendix A.
```

pyriemann/geometry/tangentspace.py
```diff
@@ -1311,15 +1312,111 @@ def transport_riemann(X, A, B):
...
+        <https://link.springer.com/article/10.1007/s41884-018-0014-4>`_
+        L. Malagò, L. Montrucchio, G. Pistone. Information Geometry, 2018, 1,
+        pp. 137–179.
+    .. [2] `O(n)-invariant Riemannian metrics on SPD matrices
+        <https://www.sciencedirect.com/science/article/pii/S0024379522004360>`_
+        Y. Thanwerdas & X. Pennec. Linear Algebra and its Applications, 2023.
+    """
+    if (
+        isinstance(n_steps, bool)
+        or not isinstance(n_steps, numbers.Integral)
```
qbarthelemy: This two-line condition seems to be equivalent to this one-line condition:
```suggestion
        not isinstance(n_steps, numbers.Integral)
```

qbarthelemy (changes requested):
My last comments

qbarthelemy (approved):
Thx @AmitSubhash for this first and great contribution!

AmitSubhash:
Glad I could help, I hope to do more ! Thanks for all the reviews you did and for your patience!

PR #478, opened 2026-07-31
Use namespace-native expand_dims instead of array-api-extra

AmitSubhash:
Part of #465.

`array_api_extra.expand_dims` is deprecated and will be removed in v1.0.0; tuple-of-int `axis` has been in the array API standard since v2025.12. `array-api-compat` 1.14 ships it, so this bumps the floor from 1.11 to 1.14 and swaps the four call sites in `geometry/covariance.py` and `geometry/geodesic.py` to the namespace-native version.

Tested at the 1.14 floor: suite passes and warnings drop from ~3540 to ~190.

Leaving the `*_indices` items (still unreleased in array-api-extra) and `cov` (#475).

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -30,6 +30,10 @@ v0.13.dev
   transport with the Bures-Wasserstein metric.
   :pr:`476` by :user:`AmitSubhash`
 
+- Update ``array-api-compat`` from >=1.11 to >=1.14, and replace the deprecated
+  ``array_api_extra.expand_dims`` by the namespace-native ``expand_dims``.
+  :pr:`478` by :user:`AmitSubhash`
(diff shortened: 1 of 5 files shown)


qbarthelemy (changes requested):
Thanks @AmitSubhash for this nice update!
Just one comment: can you complete whatsnew file?

@bruAristimunha , is it ok for you?

qbarthelemy (approved):

AmitSubhash:
Thanks for the quick merge @qbarthelemy :)

AmitSubhash:
Sorry, closed this by mistake, I confused it with #476. Reopened, nothing changed.

bruAristimunha (approved):

PR #140, opened 2021-09-14
Adding Riemannian Gaussian to pyRiemann

plcrodrigues:
This PR implements the discussion started in Issue #138 

It consists of:
- A new `sampling.py` feature containing all the code relevant for sampling from Riemannian Gaussian distributions as defined in [arXiv:1507.01760](https://arxiv.org/pdf/1507.01760.pdf). I chose to give this rather general name because we could one day end up implementing other kinds of pdfs in the SPD manifold, like mixture of Gaussians or Riemannian Laplacian distributions. I am, however, open to suggestions!
- Two examples illustrating why this feature is useful. They reproduce the figures presented in Issue #138  

It took me longer than initially expected because I ended up implementing a MCMC procedure for sampling the Riemannian Gaussian distribution. At least this ensures that we won't depend on other rather heavy packages like `pyMC3` and `pyro`.

@sylvchev had mentioned the idea of also implementing the Bayesian classifiers from [arXiv:1507.01760](https://arxiv.org/pdf/1507.01760.pdf). This could indeed be useful for the package, however, I think it goes out of the scope of this PR. Furthermore, it does not apply directly the code that I'm including here... it depends mostly of an EM algorithm that I haven't touched yet. I suggest we keep this for (near) future work :)

Cheers,
Pedro

PR #479, opened 2026-08-05
Generalize make_classification_transfer

### LINKED ISSUE #338: Generalize make_classification_transfer to several classes and domains
In `datasets` module, function `make_classification_transfer` currently supports only 2 classes and 2 domains.

The goal is to generalize it to support more than 2 classes, and more than 2 domains.

@plcrodrigues

adityasingh2400:
Fixes #338

`make_classification_transfer` was limited to two classes, two domains and 2x2 SPD matrices. This generalizes it along the four axes listed in the issue. The number of classes and the number of domains are now taken from the length of `class_names` and `domain_names`, a new `n_dim` parameter sets the dimension of the matrices, and a new `kind` parameter selects between SPD and HPD matrices.

The first domain is the source domain and every other one is a target domain built from it. Since there can now be several target domains, `domain_sep`, `theta` and `stretch` accept either a scalar, applied to all target domains, or one value per target domain. The rotation is generalized from the 2x2 rotation matrix to a rotation of angle `theta` in the plane spanned by the two first axes, which reduces exactly to the previous matrix when `n_dim` is 2.

Class centers are built as before. The first class sits at the identity, and each other class is drawn along a random unit-norm tangent vector taken back to the manifold, so its Riemannian distance to the first class center is `class_sep`. Domain transformations use the same construction with `domain_sep`, and since the Riemannian distance is congruence invariant the global mean of a target domain ends up exactly at distance `domain_sep` from the source one. The new test `test_make_classification_transfer_domain_sep` checks that property for several dimensions and two target domains.

The output for default parameters is unchanged, and this is exact rather than approximate. The sequence of draws from the random state is identical to the previous implementation, so for a given `random_state` the generated matrices match bit for bit. I checked this by dumping the output of both versions and comparing, the max absolute difference is 0. `test_make_classification_transfer_default_is_unchanged` guards the public shape and labels, and it is the only new test that already passed before the change.

`n_dim` and `kind` are appended at the end of the signature so that positional calls keep working. The full test suite passes and flake8 is clean on the changed files.

### DIFF
--- a/pyriemann/datasets/simulated.py
+++ b/pyriemann/datasets/simulated.py
@@ -415,6 +415,39 @@ def make_outliers(n_matrices, mean, sigma, outlier_coeff=10,
     return outliers
 
 
+def _sample_tangent_vector(rs, n_dim, is_complex):
+    """Draw a random unit-norm tangent vector at the identity."""
+    Pv = rs.randn(n_dim, n_dim)
+    if is_complex:
+        Pv = Pv + 1j * rs.randn(n_dim, n_dim)
+    Pv = (Pv + ctranspose(Pv)) / 2  # symmetrize
+    Pv /= np.linalg.norm(Pv)  # normalize
+    return Pv
+
+
+def _make_rotation(theta, n_dim):
+    """Rotation of angle theta in the plane of the two first axes."""
+    Q = np.eye(n_dim)
+    Q[0, 0], Q[0, 1] = np.cos(theta), -np.sin(theta)
+    Q[1, 0], Q[1, 1] = np.sin(theta), np.cos(theta)
+    return Q
+
+
+def _check_target_param(param, n_targets, name):
+    """Check a parameter defined for a scalar or for each target domain."""
+    param = np.atleast_1d(param)
+    if param.ndim != 1:
+        raise ValueError(f"{name} must be a scalar or a 1d array")
+    if param.size == 1:
+        return np.repeat(param, n_targets)
+    if param.size != n_targets:
+        raise ValueError(
+            f"{name} must be a scalar, or contain {n_targets} elements, one "
+            f"for each target domain (Got {param.size})"
+        )
+    return param
+
+
 def make_classification_transfer(
     n_matrices,
     class_sep=3.0,
@@ -425,48 +458,68 @@ def make_classification_transfer(
     random_state=None,
     class_names=[1, 2],
     domain_names=["source_domain", "target_domain"],
+    n_dim=2,
+    kind="spd",
 ):
-    """Generate 2x2 SPD matrices for two classes in source and target domains.
+    """Generate SPD or HPD matrices for several classes and domains.
 
-    Generate a set of 2x2 SPD matrices drawn from Riemannian Gaussian
+    Generate a set of SPD or HPD matrices drawn from Riemannian Gaussian
     distributions, one per class and per domain.
-    Currently, it supports two classes and two domains.
     The distributions have the same dispersions.
-    You can stretch the target domain and control a rotation matrix that maps
-    the source domain to the target domain.
+    The first domain is the source domain, and its global mean is the identity
+    matrix. Each other domain is a target domain, obtained by stretching the
+    matrices and by applying a transformation controlling its distance and its
+    rotation with respect to the source domain.
     Useful for testing classification or clustering methods on transfer
     learning applications.
 
     Parameters
     ----------
     n_matrices : int
-        Number of 2x2 matrices to generate for each class on each domain.
+        Number of matrices to generate for each class on each domain.
     class_sep : float, default=3.0
-        Distance between the centers of the classes.
+        Distance between the center of the first class and the centers of the
+        other classes.
     class_disp : float, default=1.0
         Dispersion of the matrices for each class.
-    domain_sep : float, default=5.0
-        Distance between the global means of each source and target domains.
-    theta : float, default=0.0
-        Angle of the 2x2 rotation matrix from source to target domain.
-    stretch : float, default=1.0
-        Factor to stretch the matrices in target domain. Note that when it
+    domain_sep : float | array-like, default=5.0
+        Distance between the global means of the source domain and of each
+        target domain. If a scalar, the same distance is used for all target
+        domains.
+    theta : float | array-like, default=0.0
+        Angle of the rotation matrix from source domain to each target domain,
+        in the plane spanned by the two first axes. If a scalar, the same
+        angle is used for all target domains.
+    stretch : float | array-like, default=1.0
+        Factor to stretch the matrices in each target domain. Note that when it
         is != 1.0 the class dispersions in target domain will be different than
-        those in source domain (fixed at class_disp).
+        those in source domain (fixed at class_disp). If a scalar, the same
+        factor is used for all target domains.
     random_state : None | int | RandomState instance, default=None
         Pass an int for reproducible output across multiple function calls.
     class_names : list, default=[1, 2]
-        Names of classes.
+        Names of classes, at least two.
     domain_names : list, default=["source_domain", "target_domain"]
-        Names of domains, source and target.
+        Names of domains, at least two. The first one is the source domain,
+        the other ones are target domains.
 
         .. versionadded:: 0.8
+    n_dim : int, default=2
+        Dimension of the generated matrices, at least two.
+
+        .. versionadded:: 0.13
+    kind : {"spd", "hpd"}, default="spd"
+        Kind of matrices to generate: symmetric positive-definite, or Hermitian
+        positive-definite.
+
+        .. versionadded:: 0.13
 
     Returns
     -------
-    X_enc : ndarray, shape (4*n_matrices, 2, 2)
-        Set of 2x2 SPD matrices, for two classes and two domains.
-    y_enc : ndarray, shape (4*n_matrices,)
+    X_enc : ndarray, shape (n_matrices_tot, n_dim, n_dim)
+        Set of SPD or HPD matrices, where n_matrices_tot is equal to
+        n_matrices x len(class_names) x len(domain_names).
+    y_enc : ndarray, shape (n_matrices_tot,)
         Extended labels for each matrix.
 
     Notes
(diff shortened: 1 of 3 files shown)


adityasingh2400:
Rebased onto master (now on top of #493) to clear the merge conflict.

The only conflict was in `doc/whatsnew.rst`: both this PR and #493 appended an entry to the unreleased section. I kept the #493 entry and placed this PR's entry after it. `pyriemann/datasets/simulated.py` and `tests/test_simulated.py` applied without conflict, so the code is unchanged from the previous revision.

Verified after the rebase:

```
flake8 examples tests pyriemann setup.py   clean
pytest tests/test_simulated.py             131 passed, 9 skipped
pytest                                     3714 passed, 1423 skipped
```

pyriemann/datasets/simulated.py
```diff
@@ -425,144 +458,156 @@ def make_classification_transfer(
...
+    is_complex = kind == "hpd"
+    rs = check_random_state(random_state)
+    seeds = rs.randint(100, size=n_classes * n_domains)
+
+    # create the class means, the first one at identity
+    means = [np.eye(n_dim, dtype=complex if is_complex else float)]
+    for _ in range(n_classes - 1):
+        Pv = _sample_tangent_vector(rs, n_dim, is_complex)
+        P = expm(Pv)  # take it back to the manifold
+        means.append(powm(P, alpha=class_sep))  # control distance to identity
```
qbarthelemy: Here, you control distance of each `means[i]` to `means[0]`, set to identity.
But how do you control distance of each `means[i]` to each `means[j]`?

See discussion here https://github.com/pyRiemann/pyRiemann/pull/140#discussion_r710417235

qbarthelemy: Here, you control the distance between each `means[i]` and `means[0]`, set to identity.
How do you control distance between each `means[i]` and each `means[j]`?

See discussion https://github.com/pyRiemann/pyRiemann/pull/140#discussion_r710417235

pyriemann/datasets/simulated.py
```diff
@@ -425,144 +458,156 @@ def make_classification_transfer(
...
+
+    # create the transformations from source domain to each target domain
+    transfos = []
+    for i in range(n_targets):
+        # create SPD/HPD matrix for the translation between domains
+        Pv = _sample_tangent_vector(rs, n_dim, is_complex)
+        P = expm(Pv)  # take it to the manifold
+        P = powm(P, alpha=domain_seps[i])  # control distance to identity
```
qbarthelemy: Similarly, how do you control pairwise distance between domains?

pyriemann/datasets/simulated.py
```diff
@@ -425,144 +458,156 @@ def make_classification_transfer(
...
+    """Generate SPD or HPD matrices for several classes and domains.
 
-    Generate a set of 2x2 SPD matrices drawn from Riemannian Gaussian
+    Generate a set of SPD or HPD matrices drawn from Riemannian Gaussian
     distributions, one per class and per domain.
-    Currently, it supports two classes and two domains.
     The distributions have the same dispersions.
-    You can stretch the target domain and control a rotation matrix that maps
-    the source domain to the target domain.
+    The first domain is the source domain, and its global mean is the identity
```
qbarthelemy: In general, transfer learning involves working with multiple source domains and a single target domain.
Instead, I propose referring to a "reference domain" from which all others are constructed; it is up to the user to determine whether these are source or target domains.

qbarthelemy (commented):
Thx @adityasingh2400 for this PR.

The generalization problem is non-trivial, as we must ensure that the pairwise distance between classes (resp. domains) is controlled by `class_sep` (resp. `domain_sep`).

pyriemann/datasets/simulated.py
adityasingh2400: You are right, the previous construction only fixed the distance to the first center. Fixed in ed9ba14 with what that thread from #140 was looking for: the centers are commuting matrices. One random unit-norm tangent vector is drawn as before, its eigenvectors give a common eigenbasis, and in that basis the log-eigenvalues of the centers are the vertices of a regular simplex with edge `class_sep`, the first vertex at the origin so the first center stays the identity. For commuting matrices the Riemannian distance is the Euclidean distance between the log-eigenvalues, so every pair of centers is at distance `class_sep`, exactly, for SPD and HPD alike.

The one cost is that a regular simplex with n vertices spans n - 1 dimensions, so `n_dim` must be at least `len(class_names) - 1`, and the function raises a `ValueError` otherwise. With the default `n_dim=2` that allows up to three classes. Non-commuting centers would lift that limit but only give a lower bound on the pairwise distance, since the exponential at identity does not decrease distances on this manifold, so I preferred exact. Happy to switch if you would rather have the other trade-off.

`test_make_equidistant_matrices` checks all pairwise distances on the helper for 2 to 4 matrices, dimensions 3 and 4, SPD and HPD. The random draws are unchanged, so the default output matches master to round-off, max abs diff 1.4e-12 on entries of order 1e3.

pyriemann/datasets/simulated.py
adityasingh2400: Same construction for the domains in ed9ba14: the global means of the domains are commuting matrices whose log-eigenvalues are the vertices of a regular simplex, the reference domain at the origin. With a scalar `domain_sep` all domains are pairwise at that distance. With one value per domain other than the reference, domain i is at distance s_i from the reference and domains i and j are at distance sqrt(s_i^2 + s_j^2 - s_i s_j), since the simplex directions are at 60 degrees. That formula is in the Notes of the docstring. `n_dim` must be at least `len(domain_names) - 1` for the same reason as for the classes.

`test_make_classification_transfer_domain_sep` now also checks the distance between the two non-reference domains against that formula, and `test_make_classification_transfer_domain_sep_scalar` checks that four domains in dimension 3 are all pairwise at `domain_sep`. Both work on the empirical Riemannian means of the generated matrices, so this holds for the output, not just for the construction.

pyriemann/datasets/simulated.py
adityasingh2400: Agreed, done in ed9ba14. The docstring now describes the first domain as the reference domain from which the other domains are built, and says it is up to the user to consider them as source or target domains. The per-domain parameters are documented as one value per domain other than the reference one, and the code uses `n_others` and `_check_domain_param` instead of the target wording. The default `domain_names` are unchanged so existing calls keep working.

adityasingh2400:
Thanks @qbarthelemy. All three points are addressed in ed9ba14, replies are inline. The short version: class centers and domain means are now commuting matrices whose log-eigenvalues sit on a regular simplex, which makes all pairwise distances exactly `class_sep` and `domain_sep` rather than only the distances to the first one, at the price of requiring `n_dim >= n_classes - 1` and `n_dim >= n_domains - 1`. The first domain is documented as the reference domain.

The random draws are the same as before, so the default output matches master to round-off. Verified after the change:

```
flake8 examples tests pyriemann setup.py   clean
pytest tests/test_simulated.py             135 passed, 9 skipped
pytest tests/test_transfer.py              193 passed, 127 skipped
pytest                                     3718 passed, 1423 skipped
```

I also updated the PR description, which claimed a bit-for-bit match with the previous output. It is now a match to about 1e-12.

pyriemann/datasets/simulated.py
```diff
@@ -415,6 +415,97 @@ def make_outliers(n_matrices, mean, sigma, outlier_coeff=10,
...
+    seps : ndarray, shape (n_matrices - 1,)
+        Distance between identity and each other matrix.
+    n_dim : int
+        Dimension of the matrices, at least n_matrices - 1.
+    is_complex : bool
+        Whether to generate HPD matrices rather than SPD matrices.
+
+    Returns
+    -------
+    mats : list of n_matrices ndarray, shape (n_dim, n_dim)
```
qbarthelemy:
```suggestion
    X : ndarray, shape (n_matrices, n_dim, n_dim)
```

pyriemann/datasets/simulated.py
```diff
@@ -415,6 +415,97 @@ def make_outliers(n_matrices, mean, sigma, outlier_coeff=10,
...
+    # unit norm, give the first vertex of the simplex in this eigenbasis
+    eigvals, eigvecs = np.linalg.eigh(Pv)
+    basis, _ = np.linalg.qr(np.column_stack([eigvals, np.eye(n_dim)]))
+    basis = basis[:, :n_matrices - 1] * np.sign(basis[:, 0] @ eigvals)
+    coeffs = (seps[:, None] * _make_simplex(n_matrices)[1:]) @ basis.T
+
+    mats = [np.eye(n_dim, dtype=eigvecs.dtype)]
+    for coeff in coeffs:
+        mats.append(eigvecs @ (np.exp(coeff)[:, None] * ctranspose(eigvecs)))
+    return mats
```
qbarthelemy:
```suggestion
    return np.asarray(mats)
```

pyriemann/datasets/simulated.py
```diff
@@ -415,6 +415,97 @@ def make_outliers(n_matrices, mean, sigma, outlier_coeff=10,
...
+
+
+def _make_rotation(theta, n_dim):
```
qbarthelemy:
```suggestion
def _make_rotation(n_dim, theta):
```

pyriemann/datasets/simulated.py
```diff
@@ -415,6 +415,97 @@ def make_outliers(n_matrices, mean, sigma, outlier_coeff=10,
...
+    """
+    simplex = np.zeros((n_vertices, n_vertices - 1))
+    for k in range(1, n_vertices):
+        # new vertex above the centroid of the previous ones
+        simplex[k] = simplex[:k].mean(axis=0)
+        simplex[k, k - 1] += np.sqrt((k + 1) / (2 * k))
+    return simplex
+
+
+def _make_equidistant_matrices(rs, seps, n_dim, is_complex):
```
qbarthelemy:
```suggestion
def _make_equidistant_matrices(n_dim, seps, is_complex, rs):
```

pyriemann/datasets/simulated.py
```diff
@@ -425,144 +516,184 @@ def make_classification_transfer(
     random_state=None,
     class_names=[1, 2],
     domain_names=["source_domain", "target_domain"],
+    n_dim=2,
+    kind="spd",
 ):
-    """Generate 2x2 SPD matrices for two classes in source and target domains.
+    r"""Generate SPD or HPD matrices for several classes and domains.
```
qbarthelemy:
```suggestion
    r"""Generate a random classification problem.
```

tests/test_simulated.py
```diff
@@ -189,3 +196,165 @@ def test_make_outliers(rndstate, get_mats, n_matrices, n_dim):
...
+    sep = 2.5
+    mats = np.array(_make_equidistant_matrices(
+        rndstate, np.full(n_matrices - 1, sep), n_dim, kind == "hpd"
+    ))
+    assert mats.shape == (n_matrices, n_dim, n_dim)
+    assert_array_equal(mats[0], np.eye(n_dim))
+    if kind == "spd":
+        assert is_spd(mats)
+    else:
+        assert is_hpd(mats)
```
qbarthelemy: `is_hpd` is valid for "spd" matrices
```suggestion
    assert is_hpd(mats)
```

tests/test_simulated.py
```diff
@@ -189,3 +196,165 @@ def test_make_outliers(rndstate, get_mats, n_matrices, n_dim):
...
+    for i in range(3):
+        for j in range(i + 1, 3):
+            expected = np.sqrt(seps[i]**2 + seps[j]**2 - seps[i] * seps[j])
+            assert distance_riemann(mats[i + 1], mats[j + 1]) \
+                == approx(expected)
+
+
+@pytest.mark.parametrize(
+    "n_classes, n_domains, n_dim",
+    [(2, 2, 2), (3, 3, 2), (4, 2, 3), (2, 4, 3), (3, 4, 3)],
```
qbarthelemy: Add test for a single domain
```suggestion
    [(2, 1, 3), (2, 2, 2), (3, 3, 2), (4, 2, 3), (2, 4, 3), (3, 4, 3)],
```

pyriemann/datasets/simulated.py
```diff
@@ -425,144 +516,184 @@ def make_classification_transfer(
...
-    # move the matrices in X_target with a random matrix A = P * Q
+    n_classes, n_domains = len(class_names), len(domain_names)
+    if n_classes < 2:
+        raise ValueError(
+            f"class_names must contain at least 2 elements (Got {n_classes})"
+        )
+    if n_domains < 2:
+        raise ValueError(
+            f"domain_names must contain at least 2 elements (Got {n_domains})"
+        )
```
qbarthelemy: If we delete this restriction, we are able to generate a n-class classification problem, for a single domain.
```suggestion
```
Given the major change to the function, we could take the opportunity to deprecate `make_classification_transfer` and rename it `make_classification`.

pyriemann/datasets/simulated.py
```diff
@@ -425,144 +516,184 @@ def make_classification_transfer(
...
+        those in the reference domain (fixed at class_disp). If a scalar, the
+        same factor is used for all other domains.
     random_state : None | int | RandomState instance, default=None
         Pass an int for reproducible output across multiple function calls.
     class_names : list, default=[1, 2]
-        Names of classes.
+        Names of classes, at least two.
     domain_names : list, default=["source_domain", "target_domain"]
-        Names of domains, source and target.
+        Names of domains, at least two. The first one is the reference domain,
```
qbarthelemy:
```suggestion
        Names of domains, at least one. The first one is the reference domain,
```

pyriemann/datasets/simulated.py
```diff
@@ -415,6 +415,97 @@ def make_outliers(n_matrices, mean, sigma, outlier_coeff=10,
...
+    simplex = np.zeros((n_vertices, n_vertices - 1))
+    for k in range(1, n_vertices):
+        # new vertex above the centroid of the previous ones
+        simplex[k] = simplex[:k].mean(axis=0)
+        simplex[k, k - 1] += np.sqrt((k + 1) / (2 * k))
+    return simplex
+
+
+def _make_equidistant_matrices(rs, seps, n_dim, is_complex):
+    """Generate random commuting SPD/HPD matrices at controlled distances.
```
qbarthelemy:
```suggestion
def _make_equidistant_matrices(n_dim, seps, is_complex, rs):
    """Generate random equidistant SPD/HPD matrices.
```

pyriemann/datasets/simulated.py
```diff
@@ -425,144 +516,184 @@ def make_classification_transfer(
...
+        )
+    if not isinstance(n_dim, (int, np.integer)) or n_dim < 2:
+        raise ValueError(
+            f"n_dim must be an integer at least equal to 2 (Got {n_dim})"
+        )
+    if n_dim < n_classes - 1:
+        raise ValueError(
+            f"n_dim must be at least {n_classes - 1} to place {n_classes} "
+            f"classes at the same pairwise distance (Got {n_dim})"
+        )
```
qbarthelemy: This test should be moved in `_make_equidistant_matrices`.

qbarthelemy (changes requested):
Thx @adityasingh2400 for this valuable generalization!

PR #481, opened 2026-08-05
Correct rotation estimated by TLRotate in tangent space

adityasingh2400:
`pyriemann/optimization/grassmann.py` has three defects that make `TLRotate` return wrong rotations.

On current master, a known-answer check plus a finite difference check:

```
Q equals the rotation      : False
Q equals its transpose     : True
|| X_source Q - X_target ||: 11.957493614804005
documented loss sum_i w_i d^2: 1.031879145975589
value returned by _loss      : 1.009228770614949
max error of _grad vs numerical gradient: 0.2401209229939022
```

With this branch:

```
Q equals the rotation      : True
Q equals its transpose     : False
|| X_source Q - X_target ||: 5.570116854550151e-15
documented loss sum_i w_i d^2: 1.031879145975589
value returned by _loss      : 1.031879145975589
max error of _grad vs numerical gradient: 6.466127633331098e-10
```

The reproduction script is at the bottom of this description.

The three defects, independent and each about one line:

**1. `_get_rotation_tangentspace` returns the transpose of the Procrustes solution.** Its docstring documents minimizing `|| X_source Q - X_target ||_F^2`, whose solution is `U V^T` with `U S V^T = svd(X_source^T X_target)`. `numpy.linalg.svd` returns `u, s, vh` with `vh = V^T`, so the solution is `u @ vh`, but the code returns `vh.T @ u.T`, which is its transpose, that is its inverse. `TLRotate.fit_transform` then applies `X[idx] @ self.rotations_[d]`, so every source domain in tangent space is rotated by the inverse of the intended rotation. Fixed to `u @ vh`.

**2. `_loss` returns a sum of distances, not of squared distances.** The docstring of `_get_rotation_manifold` states the objective is `L(Q) = sum_i w_i d^2(...)`, and `_grad` is the gradient of that squared objective for both metrics, confirmed above by finite differences. The backtracking line search in `_run_minimization` was therefore testing the descent of a different function from the one being differentiated. Fixed by passing `squared=True`.

**3. `_grad` with `metric="riemann"` diagonalized a non-symmetric matrix with `eigh`.** `M = X^-1 Q Y Q^T` is a product of two SPD matrices, so it has real positive eigenvalues but is not symmetric. `numpy.linalg.eigh` reads only one triangle, so `expm(logM)` does not recover `M`. This came in with #468, which correctly replaced `eig` by `eigh` in `_warm_start` where the inputs really are symmetric, but the same replacement in `_grad` does not hold. Rather than reverting to `eig`, which is what raised the warning #468 removed, this uses the similar SPD matrix `X^-1/2 S X^-1/2`, since `logm(X^-1 S) = X^-1/2 logm(X^-1/2 S X^-1/2) X^1/2`. `eigh` stays valid, the result stays real, and the gradient error drops from 2.4e-1 to 6.5e-10.

**Why this was never caught.** `test_tlrotate_tangentspace` had a vacuous assertion. It compared `X_rot[domain == "target_domain"]` while the domains in that test are `"tgt"`, `"src1"` and `"src2"`, so both sides of the comparison were empty arrays. It now compares `domain == "tgt"`.

Tests added to `tests/test_transfer.py`: `test_get_rotation_tangentspace` (exact recovery of a known rotation, for both forms of `expl_var`), `test_tlrotate_tangentspace_recovers_rotation` (end to end through `TLRotate`), `test_grassmann_loss` (`_loss` equals the weighted sum of squared distances), and `test_grassmann_grad` (`_grad` equals the numerical gradient of `_loss`, for both metrics).

7 selected tests fail on master and pass with the branch. Full `tests/` run gives 3595 passed and 1272 skipped. `flake8` is clean and there is a `doc/whatsnew.rst` entry.

Reproduction:

```python
import numpy as np
from pyriemann.geometry.distance import distance_riemann
from pyriemann.optimization.grassmann import _get_rotation_tangentspace, _loss, _grad

rs = np.random.RandomState(0)
X_source = rs.randn(20, 4)
rotation = np.linalg.qr(rs.randn(4, 4))[0]
X_target = X_source @ rotation

Q = _get_rotation_tangentspace(X_source, X_target, expl_var=1.0)
print("Q equals the rotation      :", np.allclose(Q, rotation))
print("Q equals its transpose     :", np.allclose(Q, rotation.T))
print("|| X_source Q - X_target ||:", np.linalg.norm(X_source @ Q - X_target))

Xm = make_matrices(3, 4, "spd", rs=21)
Ym = make_matrices(3, 4, "spd", rs=22)
w = np.array([0.5, 0.3, 0.2])
Qm = np.linalg.qr(rs.randn(4, 4))[0]

print("documented loss sum_i w_i d^2:", w @ distance_riemann(Xm, Qm @ Ym @ Qm.T) ** 2)
print("value returned by _loss      :", _loss(Qm, Xm, Ym, w, metric="riemann"))

eps, grad_num = 1e-6, np.zeros((4, 4))
f = lambda Q: w @ distance_riemann(Xm, Q @ Ym @ Q.T) ** 2
for i in range(4):
    for j in range(4):
        step = np.zeros((4, 4))
        step[i, j] = eps
        grad_num[i, j] = (f(Qm + step) - f(Qm - step)) / (2 * eps)
print("max error of _grad vs numerical gradient:",
      np.abs(_grad(Qm, Xm, Ym, w, metric="riemann") - grad_num).max())
```

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -71,6 +71,11 @@ v0.13.dev
   diagonalizes a non-symmetric matrix with ``numpy.linalg.eigh``.
   :pr:`492` by :user:`adityasingh2400`
 
+- Correct rotation estimation of :class:`pyriemann.transfer.TLRotate` in tangent space, which applied the
+  transpose of the Procrustes rotation, ie the inverse of the rotation mapping
+  each source domain onto the target domain.
+  :pr:`481` by :user:`adityasingh2400`
+
 v0.12 (July 2026)
 -----------------
 
--- a/pyriemann/optimization/grassmann.py
+++ b/pyriemann/optimization/grassmann.py
@@ -4,6 +4,7 @@ import warnings
 
 import numpy as np
 
+from ..geometry.base import invsqrtm, logm, sqrtm
 from ..geometry.distance import distance
 from ..utils._check import check_weights
 
@@ -34,7 +35,7 @@ def _retract(X, v):
 def _loss(Q, X, Y, weights, metric="euclid"):
     """Loss function for estimating the rotation matrix."""
 
-    return weights @ distance(X, Q @ Y @ Q.T, metric=metric)
+    return weights @ distance(X, Q @ Y @ Q.T, metric=metric, squared=True)
 
 
 def _grad(Q, X, Y, weights, metric="euclid"):
@@ -44,11 +45,13 @@ def _grad(Q, X, Y, weights, metric="euclid"):
         return np.einsum("a,abc->bc", weights, -4 * (X - Q @ Y @ Q.T) @ Q @ Y)
 
     elif metric == "riemann":
-        M = np.linalg.solve(X, Q) @ Y @ Q.T
-        eigvals, eigvecs = np.linalg.eigh(M)
-        logeigvals = np.expand_dims(np.log(eigvals), -2)
-        inveigvecs = np.linalg.solve(eigvecs, np.eye(eigvecs.shape[-1]))
-        logM = (eigvecs * logeigvals) @ inveigvecs
+        # logm of X^-1 S, with S = Q Y Q^T. This product of two SPD matrices
+        # is not symmetric, so it cannot be diagonalized by eigh. It is
+        # however similar to the SPD matrix X^-1/2 S X^-1/2, which is:
+        # logm(X^-1 S) = X^-1/2 logm(X^-1/2 S X^-1/2) X^1/2
+        X_invsqrt, X_sqrt = invsqrtm(X), sqrtm(X)
+        S = Q @ Y @ Q.T
+        logM = X_invsqrt @ logm(X_invsqrt @ S @ X_invsqrt) @ X_sqrt
         return np.einsum("a,abc->bc", weights, 4 * logM @ Q)
 
     else:
@@ -297,6 +300,8 @@ def _get_rotation_tangentspace(X_source, X_target, expl_var):
     Notes
     -----
     .. versionadded:: 0.8
+    .. versionchanged:: 0.13
+        Correct rotation estimation.
 
     References
     ----------
--- a/pyriemann/optimization/grassmann.py
+++ b/pyriemann/optimization/grassmann.py
@@ -4,7 +4,6 @@ import warnings
 
 import numpy as np
 
-from ..geometry.base import invsqrtm, logm, sqrtm
 from ..geometry.distance import distance
 from ..utils._check import check_weights
 
@@ -35,7 +34,7 @@ def _retract(X, v):
 def _loss(Q, X, Y, weights, metric="euclid"):
     """Loss function for estimating the rotation matrix."""
 
-    return weights @ distance(X, Q @ Y @ Q.T, metric=metric, squared=True)
+    return weights @ distance(X, Q @ Y @ Q.T, metric=metric)
 
 
 def _grad(Q, X, Y, weights, metric="euclid"):
@@ -45,13 +44,11 @@ def _grad(Q, X, Y, weights, metric="euclid"):
         return np.einsum("a,abc->bc", weights, -4 * (X - Q @ Y @ Q.T) @ Q @ Y)
 
     elif metric == "riemann":
-        # logm of X^-1 S, with S = Q Y Q^T. This product of two SPD matrices
-        # is not symmetric, so it cannot be diagonalized by eigh. It is
-        # however similar to the SPD matrix X^-1/2 S X^-1/2, which is:
-        # logm(X^-1 S) = X^-1/2 logm(X^-1/2 S X^-1/2) X^1/2
-        X_invsqrt, X_sqrt = invsqrtm(X), sqrtm(X)
-        S = Q @ Y @ Q.T
-        logM = X_invsqrt @ logm(X_invsqrt @ S @ X_invsqrt) @ X_sqrt
+        M = np.linalg.solve(X, Q) @ Y @ Q.T
+        eigvals, eigvecs = np.linalg.eigh(M)
+        logeigvals = np.expand_dims(np.log(eigvals), -2)
+        inveigvecs = np.linalg.solve(eigvecs, np.eye(eigvecs.shape[-1]))
+        logM = (eigvecs * logeigvals) @ inveigvecs
         return np.einsum("a,abc->bc", weights, 4 * logM @ Q)
 
     else:
--- a/pyriemann/optimization/grassmann.py
+++ b/pyriemann/optimization/grassmann.py
@@ -323,4 +323,4 @@ def _get_rotation_tangentspace(X_source, X_target, expl_var):
 
     u = u[:, :n_comps]
     vh = vh[:n_comps, :]
-    return vh.T @ u.T
+    return u @ vh
--- a/tests/test_optimization.py
+++ b/tests/test_optimization.py
@@ -6,6 +6,7 @@ from pytest import approx
 from pyriemann.geometry.distance import distance
 from pyriemann.optimization.grassmann import (
     _get_rotation_manifold,
+    _get_rotation_tangentspace,
     _grad,
     _loss,
 )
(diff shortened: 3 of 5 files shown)


tests/test_transfer.py
```diff
@@ -440,15 +445,93 @@ def test_tlrotate_tangentspace(rndstate, get_weights,
...
+    rotation = np.linalg.qr(rndstate.randn(n_ts, n_ts))[0]
+    X_tgt = X_src @ rotation
+
+    X = np.concatenate([X_tgt, X_src])
+    domain = np.array(n_vectors_d * ["tgt"] + n_vectors_d * ["src"])
+    _, y_enc = encode_domains(X, np.concatenate([y, y]), domain)
+
+    tlrot = TLRotate(target_domain="tgt", n_components=1, n_clusters=2)
+    X_rot = tlrot.fit_transform(X, y_enc)
+
```
qbarthelemy, line 497: We can add an assertion comparing rotation matrices:
```suggestion
    assert_array_almost_equal(rotation, tlrot.rotations_["src"])

```

tests/test_transfer.py
```diff
@@ -440,15 +445,93 @@ def test_tlrotate_tangentspace(rndstate, get_weights,
...
+            np.mean(X_src[y == label], axis=0) - m_tgt
+        )
+        dist_after = np.linalg.norm(
+            np.mean(X_rot[domain == "src"][y == label], axis=0) - m_tgt
+        )
+        assert dist_after < dist_before / 10
+
+
+@pytest.mark.parametrize("expl_var", [0.999, 4])
+def test_get_rotation_tangentspace(rndstate, expl_var):
```
qbarthelemy, line 511: New tests `test_get_rotation_tangentspace`, `test_grassmann_loss` and `test_grassmann_grad` should be placed into a new file `test_optimization.py`

qbarthelemy (changes requested):
Thanks @adityasingh2400 for these corrections.

For point 1 on TSA,
formula comes from code shared into https://github.com/pyRiemann/pyRiemann/issues/319, in `TSAxPCA.transform()`

For points 2 and 3 on RPA,
@plcrodrigues , could you take a look at them?

qbarthelemy:
Correction of point 1 significantly improves the TSA results,
see figure from example "Comparison of pipelines for transfer learning"

on [master](https://pyriemann.readthedocs.io/en/latest/auto_examples/transfer/plot_comparison_pipelines.html)

pr_481_master

on branch

pr_481_branch

qbarthelemy:
Is it possible to split this PR in two: one for the rotation in tangent space and one for that in the manifold? @adityasingh2400

adityasingh2400:
Split done. This PR is now only the tangent-space rotation, and the manifold loss and gradient moved to #492.

I also moved the tests. `tests/test_optimization.py` is new and holds `test_get_rotation_tangentspace` here, and `test_grassmann_loss` and `test_grassmann_grad` in #492. The two PRs both touch that new file, so whichever lands second will need a small merge.

On the assertion, I tried it and it does not hold, so I left it out rather than push something red. `assert_array_almost_equal(rotation, tlrot.rotations_["src"])` fails with max abs difference 0.73 on that test. The reason is that `TLRotate` aligns cluster centroids rather than the vectors themselves, so with `n_clusters=2` there are only 2 point correspondences constraining a 4x4 rotation, which is underdetermined. Raising the cluster count does not converge either, since the KMeans partition shifts:

```
n_clusters   max|Q - rotation|
    2          0.729
    3          0.121
    4          0.071
    5          1.547
   10          0.139
   30          0.896
```

The check you are after does hold on the Procrustes routine itself, and that is already asserted in `test_get_rotation_tangentspace`, now in `tests/test_optimization.py`:

```python
assert_array_almost_equal(Q, rotation)
assert_array_almost_equal(X_source @ Q, X_target)
```

There it is exact, because all 20 vectors constrain the fit. Happy to add it to the integration test too if you would rather assert something weaker there, such as the rotation being orthogonal.

Reverting the one-line fix fails 3 tests. Whole suite is 195 passed, 127 skipped.

tests/test_transfer.py
adityasingh2400: I tried this and it does not hold, so I left it out rather than push a red test. It fails at max abs difference 0.73.

`TLRotate` fits the Procrustes rotation to cluster centroids, not to the vectors, so with `n_clusters=2` there are only two point correspondences constraining a 4x4 rotation. That is underdetermined, and raising the cluster count does not converge either, since the KMeans partition shifts:

```
n_clusters   max|Q - rotation|
    2          0.729
    3          0.121
    4          0.071
    5          1.547
   10          0.139
   30          0.896
```

The assertion does hold on the Procrustes routine itself, where all 20 vectors constrain the fit, and it is already there in `test_get_rotation_tangentspace`, now in `tests/test_optimization.py`:

```python
assert_array_almost_equal(Q, rotation)
assert_array_almost_equal(X_source @ Q, X_target)
```

Happy to assert something weaker here instead if you want a check at this level, orthogonality of `tlrot.rotations_["src"]` for example. Just say which.

tests/test_transfer.py
adityasingh2400: Done. `tests/test_optimization.py` is new and holds `test_get_rotation_tangentspace` here. `test_grassmann_loss` and `test_grassmann_grad` went to the same file in #492, since they belong with the manifold fix. Whichever of the two PRs lands second will need a small merge on that file.

pyriemann/optimization/grassmann.py
qbarthelemy, line 305:
```suggestion
    .. versionchanged:: 0.13
        Correct rotation estimation.

```

tests/test_transfer.py
qbarthelemy: `randstate = 1234` should be deleted
```suggestion
```

tests/test_optimization.py
```diff
@@ -0,0 +1,19 @@
...
+
+@pytest.mark.parametrize("expl_var", [0.999, 4])
+def test_get_rotation_tangentspace(rndstate, expl_var):
+    """Test that Procrustes analysis maps source onto target"""
+    n_vectors, n_ts = 20, 4
+    X_source = rndstate.randn(n_vectors, n_ts)
+    rotation = np.linalg.qr(rndstate.randn(n_ts, n_ts))[0]
+    X_target = X_source @ rotation
+
+    Q = _get_rotation_tangentspace(X_source, X_target, expl_var)
```
qbarthelemy, line 82: When this PR will be rebased on master
```suggestion
    Q = _get_rotation_tangentspace(X_source, X_target, expl_var)
    assert _is_orth(Q)
```

doc/whatsnew.rst
```diff
@@ -64,6 +64,11 @@ v0.13.dev
   leftover ``print()`` in its bandwidth estimation.
   :pr:`483` by :user:`adityasingh2400`
 
+- Fix :class:`pyriemann.transfer.TLRotate` in tangent space, which applied the
```
qbarthelemy:
```suggestion
- Correct rotation estimation of :class:`pyriemann.transfer.TLRotate` in tangent space, which applied the
```

qbarthelemy (changes requested):
Some minor comments.

adityasingh2400:
Rebased on master and applied all four, in `958fe62`.

Since #492 landed, this branch now carries only the tangent space fix. The `whatsnew` entry is split accordingly, so #492 keeps the manifold entry and #481 has its own with your wording.

Worth flagging one thing from the rebase. Replaying the old commits reintroduced the pre-review manifold code, which quietly reverted #492 in `grassmann.py`, dropping `squared=True` from `_loss` and restoring the `eigh` on the non-symmetric product in `_grad`. The four `test_grassmann_*` cases caught it. I rebuilt the file from master so the only change here is `return u @ vh` plus the `versionchanged` note.

Verified against master as the baseline. Master is 3 failed and 195 passed, this branch is 3 failed and 198 passed, the same three being a local scikit-learn incompatibility in `test_tlclassifier_manifold` that is not related to either PR. Reverting just the one rotation line fails all three of the new tests.

qbarthelemy (approved):
Thanks for this valuable correction @adityasingh2400 !

PR #482, opened 2026-08-05
Fix TLScale modifying input matrices in-place

adityasingh2400:
`TLScale.fit_transform` writes into the array given by the caller, so calling it twice on the same matrices gives two different results.

On current master:

```python
import numpy as np
from pyriemann.datasets import make_classification_transfer
from pyriemann.transfer import TLScale

X, y_enc = make_classification_transfer(n_matrices=10, random_state=42)
X_copy = X.copy()

X_new = TLScale(target_domain="target_domain").fit_transform(X, y_enc)
print("input matrices left untouched:", np.array_equal(X, X_copy))

X_new2 = TLScale(target_domain="target_domain").fit_transform(X, y_enc)
print("same call, same result       :", np.array_equal(X_new, X_new2))
print("max difference between them  :", np.abs(X_new - X_new2).max())
```

```
input matrices left untouched: False
same call, same result       : False
max difference between them  : 3.6974495398852905
```

The cause is `X[idx] = self._center(X[idx], self._means[d])` in `fit_transform`. The second run sees data that is already centered on each domain mean, so it scales the wrong thing.

`TLCenter`, `TLRotate` and `TLScale.transform` all rebind a local name instead, so the fix follows that same pattern with a local `X_d`. The returned `X_new` is unchanged bit for bit, only the caller's array is left alone.

Test added to `tests/test_transfer.py`: `test_transformer_does_not_modify_inputs`, parametrized over `TLCenter`, `TLScale` and `TLRotate` and over matrix and vector inputs, checking both that the input is untouched and that two identical calls agree.

Of 6 selected cases, 1 fails on master (the `TLScale` matrix branch) and all 6 pass with the branch. Full `tests/` run gives 3594 passed and 1272 skipped. `flake8` is clean and there is a `doc/whatsnew.rst` entry.

### DIFF
--- a/tests/test_transfer.py
+++ b/tests/test_transfer.py
@@ -345,6 +345,31 @@ def test_tlscale_tangentspace(rndstate, get_weights, use_weight):
     assert X_new.shape == X.shape
 
 
+@pytest.mark.parametrize("estimator", [TLCenter, TLScale, TLRotate])
+@pytest.mark.parametrize("use_vectors", [True, False])
+def test_transformer_does_not_modify_inputs(rndstate, estimator, use_vectors):
+    """Test that fit_transform does not modify the inputs"""
+    if use_vectors:
+        X, y_enc = make_classification_transfer_tangspace(
+            rndstate, ["tgt", "src"], n_vectors_d=20, n_ts=4,
+        )
+    else:
+        X, y_enc = make_classification_transfer(
+            n_matrices=20, random_state=rndstate,
+        )
+        _, y, domain = decode_domains(X, y_enc)
+        domain = np.where(domain == "target_domain", "tgt", "src")
+        _, y_enc = encode_domains(X, y, domain)
+    X_copy = X.copy()
+
+    X_new = estimator(target_domain="tgt").fit_transform(X, y_enc)
+
+    assert_array_equal(X, X_copy)
+    # calling fit_transform twice on the same inputs gives the same outputs
+    X_new2 = estimator(target_domain="tgt").fit_transform(X, y_enc)
+    assert_array_equal(X_new, X_new2)
+
+
 @pytest.mark.parametrize("metric", ["euclid", "riemann"])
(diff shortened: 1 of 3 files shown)


tests/test_transfer.py
```diff
@@ -345,6 +345,31 @@ def test_tlscale_tangentspace(rndstate, get_weights, use_weight):
     assert X_new.shape == X.shape
 
 
+@pytest.mark.parametrize("estimator", [TLCenter, TLScale, TLRotate])
+@pytest.mark.parametrize("use_vectors", [True, False])
+def test_transformer_does_not_modify_inputs(rndstate, estimator, use_vectors):
```
qbarthelemy: Can you rename this test into `test_tltransformer_fittransform` and move it before `test_tldummy` (because it is a generic test)?

tests/test_transfer.py
```diff
@@ -345,6 +345,31 @@ def test_tlscale_tangentspace(rndstate, get_weights, use_weight):
...
+            n_matrices=20, random_state=rndstate,
+        )
+        _, y, domain = decode_domains(X, y_enc)
+        domain = np.where(domain == "target_domain", "tgt", "src")
+        _, y_enc = encode_domains(X, y, domain)
+    X_copy = X.copy()
+
+    X_new = estimator(target_domain="tgt").fit_transform(X, y_enc)
+
+    assert_array_equal(X, X_copy)
```
qbarthelemy: Remove empty line between act and assert
```suggestion
    # calling fit_transform does not modify inputs
    X_new = estimator(target_domain="tgt").fit_transform(X, y_enc)
    assert_array_equal(X, X_copy)

```

qbarthelemy (changes requested):
Good catch @adityasingh2400 !
I just have a few minor remarks.

qbarthelemy (approved):
Thx @adityasingh2400 !

PR #483, opened 2026-08-05
Fix parameter bandwidth of MeanShift

adityasingh2400:
Two estimators in `pyriemann/clustering.py` raise `AttributeError` because `__init__` and `fit` do not store the parameters they are given.

On current master:

```python
import numpy as np
from sklearn.base import clone
from pyriemann.clustering import MeanShift, KmeansPerClassTransform

X = make_matrices(12, 3, "spd", rs=42)

MeanShift(bandwidth=1.0).fit(X)
KmeansPerClassTransform().get_params()
clone(KmeansPerClassTransform(metric="logeuclid"))
```

```
MeanShift(bandwidth=1.0) -> AttributeError 'MeanShift' object has no attribute '_bandwidth'
KmeansPerClassTransform().get_params() -> AttributeError 'KmeansPerClassTransform' object has no attribute 'n_clusters'
clone(KmeansPerClassTransform()) -> AttributeError 'KmeansPerClassTransform' object has no attribute 'n_clusters'
```

`MeanShift.fit` sets `self._bandwidth` only in the `bandwidth is None` branch, then reads `self._bandwidth ** 2` unconditionally on the next line. The documented `bandwidth` parameter has therefore never worked since it was added in #364 for v0.9. The same method also printed the estimated bandwidth to stdout, a leftover from that PR which pollutes any `cross_val_score` or `GridSearchCV` loop, so it is removed here.

`KmeansPerClassTransform.__init__` built a `Kmeans` instance instead of storing its parameters, which breaks the scikit-learn estimator contract, so `get_params`, `clone`, `cross_val_score`, `GridSearchCV` and `Pipeline` introspection all fail.

Note that keeping `**params` cannot fix this, since `BaseEstimator._get_param_names` skips `VAR_KEYWORD` parameters, so any extra keyword would be silently dropped by `clone` even after storing `n_clusters`. The signature now lists the parameters of `Kmeans` explicitly, `__init__` only assigns them, and the `Kmeans` instance is built in `fit`. Every existing call site passes these by keyword, so this is backward compatible.

```
get_params -> {'init': 'random', 'max_iter': 100, 'metric': 'riemann', 'n_clusters': 2,
               'n_init': 10, 'n_jobs': 1, 'random_state': None, 'tol': 0.0001}
```

Tests added to `tests/test_clustering.py`: `test_meanshift_bandwidth` (explicit bandwidth is honored and nothing is printed, via `capsys`), `test_clustering_get_params` (`clone` round trip for all four clusterers) and `test_kmeansperclasstransform_get_params` (its parameters match `Kmeans`).

Of 7 selected tests, 4 fail on master and all 7 pass with the branch. Full `tests/` run gives 3595 passed and 1274 skipped. `flake8` is clean and there is a `doc/whatsnew.rst` entry.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -62,11 +62,6 @@ v0.13.dev
 - Fix parameter ``bandwidth`` of :class:`pyriemann.clustering.MeanShift`,
   which raised an ``AttributeError`` when set to a value, and remove a
   leftover ``print()`` in its bandwidth estimation.
-  Fix ``__init__`` of :class:`pyriemann.clustering.KmeansPerClassTransform`,
-  which did not store its parameters, so that ``get_params()``, ``clone()``
-  and ``cross_val_score()`` raised an ``AttributeError``. Parameters of
-  :class:`pyriemann.clustering.Kmeans` are now explicit in its signature,
-  replacing ``**params``.
   :pr:`483` by :user:`adityasingh2400`
 
 v0.12 (July 2026)
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -30,6 +30,16 @@ v0.13.dev
   transport with the Bures-Wasserstein metric.
   :pr:`476` by :user:`AmitSubhash`
 
+- Fix parameter ``bandwidth`` of :class:`pyriemann.clustering.MeanShift`,
+  which raised an ``AttributeError`` when set to a value, and remove a
+  leftover ``print()`` in its bandwidth estimation.
+  Fix ``__init__`` of :class:`pyriemann.clustering.KmeansPerClassTransform`,
+  which did not store its parameters, so that ``get_params()``, ``clone()``
+  and ``cross_val_score()`` raised an ``AttributeError``. Parameters of
+  :class:`pyriemann.clustering.Kmeans` are now explicit in its signature,
+  replacing ``**params``.
+  :pr:`483` by :user:`adityasingh2400`
+
 v0.12 (July 2026)
 -----------------
 
--- a/pyriemann/clustering.py
+++ b/pyriemann/clustering.py
@@ -280,8 +280,38 @@ class KmeansPerClassTransform(SpdTransfMixin, BaseEstimator):
     ----------
     n_clusters : int, default=2
         Number of clusters.
-    **params : dict
-        The keyword arguments passed to :class:`pyriemann.clustering.Kmeans`.
+    max_iter : int, default=100
+        Maximum number of iteration to reach convergence.
+    metric : string | dict, default="riemann"
+        Metric used for mean estimation (for the list of supported metrics,
+        see :func:`pyriemann.geometry.mean.gmean`) and for distance estimation
+        (see :func:`pyriemann.geometry.distance.distance`).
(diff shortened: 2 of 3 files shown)


qbarthelemy (changes requested):
Thanks @adityasingh2400 for all the suggested corrections; I’ll go through them one by one.

Is it possible to split this PR into two? One for `MeanShift` and one for `KmeansPerClassTransform`.

For `MeanShift`, for tests, you can simply extend `test_meanshift` adding a new `parametrize`:
```
@pytest.mark.parametrize("kernel", ["normal", "uniform", callable_kernel])
@pytest.mark.parametrize("bandwidth", [None, 0.1])
@pytest.mark.parametrize("metric", ["euclid", "logeuclid"])
def test_meanshift(kernel, bandwidth, metric, get_mats):
    n_matrices, n_channels = 10, 3
    X = get_mats(n_matrices, n_channels, "spd")

    clt = MeanShift(
        kernel=kernel,
        bandwidth=bandwidth,
        metric=metric,
    )
    clt.fit(X)
```

For `KmeansPerClassTransform`, I’d suggest this alternative fix instead, as it avoids having to copy all the parameters:
```
class KmeansPerClassTransform(SpdTransfMixin, BaseEstimator):

    def __init__(self, **params):
        """Init."""
        defaults = Kmeans().get_params()
        defaults.update(params)
        self.params = defaults
        for name, value in defaults.items():
            setattr(self, name, value)

    def get_params(self, deep=True):
        return self.params.copy()

    def set_params(self, **params):
        for name, value in params.items():
            self.params[name] = value
            setattr(self, name, value)
        return self

    def fit(self, X, y):
        self._km = Kmeans(**self.params)
        self.metric = self._km.metric
        ...
```
and `n_clusters` should be removed from docstring.

adityasingh2400:
Yes, splitting works for me. `MeanShift` stays here and `KmeansPerClassTransform` moves to its own PR.

Both of your suggestions are better than what I had. Extending `test_meanshift` with the extra `parametrize` covers the same combinations without a separate test, and building the defaults from `Kmeans().get_params()` avoids restating the signature, which was the part of my change I liked least.

I have merged master in, the only conflict was the `whatsnew` entry.

adityasingh2400:
Split done, and I used both of your suggestions.

`MeanShift` stays here. `KmeansPerClassTransform` is now #493, built the way you proposed with `defaults = Kmeans().get_params()`, so nothing restates the signature. One consequence to flag: `n_clusters` is no longer an explicit keyword in the constructor, it comes from the `Kmeans` defaults like the rest. `KmeansPerClassTransform(n_clusters=3)` still works.

For the tests here I folded the bandwidth coverage into `test_meanshift` with your extra `parametrize` rather than keeping a separate function, and kept the two assertions that actually pin the fix:

```python
if bandwidth is not None:
    assert clt._bandwidth == bandwidth
assert capsys.readouterr().out == ""
```

The second one is what catches the stray `print()`. Reverting the source fails 12 of the 16 parametrizations.

I moved `test_clustering_get_params` to #493, since it fails against unpatched `master` for `KmeansPerClassTransform` and so belongs with that fix rather than this one.

This PR is 54 passed, 57 skipped.

qbarthelemy (approved):
Thx @adityasingh2400 !

PR #492, opened 2026-09-07
Correct loss and gradient minimized by TLRotate in manifold

adityasingh2400:
Split out of #481 at @qbarthelemy's request, so that the tangent-space rotation and the manifold loss/gradient are reviewed separately. #481 now carries only the tangent-space fix.

Two problems in `pyriemann/optimization/grassmann.py`, both in the manifold branch of `TLRotate`.

**`_loss` did not return what it documents.** `TLRotate` in manifold minimizes the weighted sum of *squared* distances, and `_grad` is the derivative of that. `_loss` returned unsquared distances, so the value it reported was not the quantity being minimized and did not match its own gradient.

**`_grad` diagonalized a non-symmetric matrix with `eigh` for `metric="riemann"`.** It computed `M = X^-1 Q Y Q^T` and passed it to `numpy.linalg.eigh`. `M` is a product of two SPD matrices, which is generally not symmetric, and `eigh` reads only one triangle, so the result was the log of a different matrix. The fix uses the standard similarity transform:

```
logm(X^-1 S) = X^-1/2 logm(X^-1/2 S X^-1/2) X^1/2
```

where `X^-1/2 S X^-1/2` is SPD and safe to diagonalize.

**Tests.** `tests/test_optimization.py` gets `test_grassmann_loss`, which pins the loss to the weighted sum of squared distances, and `test_grassmann_grad`, which checks `_grad` against a central finite difference of `_loss` for both metrics. Against unpatched `master` all 4 parametrizations fail. With the fix the file is 4 passed, and `tests/test_transfer.py` plus `tests/test_optimization.py` together are 196 passed, 127 skipped.

https://claude.ai/code/session_0189zxefNUxZmXLRED6jPeMy

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -64,7 +64,7 @@ v0.13.dev
   sum of squared distances it is documented to minimize and that the gradient
   is derived from, and the gradient for ``metric="riemann"`` no longer
   diagonalizes a non-symmetric matrix with ``numpy.linalg.eigh``.
-  :pr:`484` by :user:`adityasingh2400`
+  :pr:`492` by :user:`adityasingh2400`
 
 v0.12 (July 2026)
 -----------------
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -59,6 +59,13 @@ v0.13.dev
   for two matrices, which were using a wrong position on the geodesic when ``sample_weight`` is not uniform.
   :pr:`480` by :user:`adityasingh2400`
 
+- Fix the loss and the gradient minimized by
+  :class:`pyriemann.transfer.TLRotate` in manifold: the loss now returns the
+  sum of squared distances it is documented to minimize and that the gradient
+  is derived from, and the gradient for ``metric="riemann"`` no longer
+  diagonalizes a non-symmetric matrix with ``numpy.linalg.eigh``.
+  :pr:`484` by :user:`adityasingh2400`
+
 v0.12 (July 2026)
 -----------------
 
--- a/pyriemann/optimization/grassmann.py
+++ b/pyriemann/optimization/grassmann.py
@@ -4,6 +4,7 @@ import warnings
 
 import numpy as np
 
+from ..geometry.base import invsqrtm, logm, sqrtm
 from ..geometry.distance import distance
 from ..utils._check import check_weights
 
@@ -34,7 +35,7 @@ def _retract(X, v):
 def _loss(Q, X, Y, weights, metric="euclid"):
     """Loss function for estimating the rotation matrix."""
 
-    return weights @ distance(X, Q @ Y @ Q.T, metric=metric)
+    return weights @ distance(X, Q @ Y @ Q.T, metric=metric, squared=True)
 
 
 def _grad(Q, X, Y, weights, metric="euclid"):
@@ -44,11 +45,13 @@ def _grad(Q, X, Y, weights, metric="euclid"):
         return np.einsum("a,abc->bc", weights, -4 * (X - Q @ Y @ Q.T) @ Q @ Y)
 
     elif metric == "riemann":
-        M = np.linalg.solve(X, Q) @ Y @ Q.T
-        eigvals, eigvecs = np.linalg.eigh(M)
-        logeigvals = np.expand_dims(np.log(eigvals), -2)
-        inveigvecs = np.linalg.solve(eigvecs, np.eye(eigvecs.shape[-1]))
-        logM = (eigvecs * logeigvals) @ inveigvecs
+        # logm of X^-1 S, with S = Q Y Q^T. This product of two SPD matrices
+        # is not symmetric, so it cannot be diagonalized by eigh. It is
+        # however similar to the SPD matrix X^-1/2 S X^-1/2, which is:
+        # logm(X^-1 S) = X^-1/2 logm(X^-1/2 S X^-1/2) X^1/2
+        X_invsqrt, X_sqrt = invsqrtm(X), sqrtm(X)
+        S = Q @ Y @ Q.T
+        logM = X_invsqrt @ logm(X_invsqrt @ S @ X_invsqrt) @ X_sqrt
         return np.einsum("a,abc->bc", weights, 4 * logM @ Q)
 
     else:
--- a/tests/test_optimization.py
+++ b/tests/test_optimization.py
@@ -7,6 +7,9 @@ from pyriemann.geometry.distance import distance
 from pyriemann.optimization.grassmann import _grad, _loss
 
 
+pytestmark = pytest.mark.numpy_only
+
+
 @pytest.mark.parametrize("metric", ["euclid", "riemann"])
 def test_grassmann_loss(metric, get_mats, get_weights):
     """Test that loss is the weighted sum of squared distances"""
--- a/tests/test_optimization.py
+++ b/tests/test_optimization.py
@@ -1,0 +1,47 @@
+import numpy as np
+from numpy.testing import assert_array_almost_equal
+import pytest
+from pytest import approx
+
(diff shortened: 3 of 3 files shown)


doc/whatsnew.rst
```diff
@@ -59,6 +59,13 @@ v0.13.dev
   for two matrices, which were using a wrong position on the geodesic when ``sample_weight`` is not uniform.
   :pr:`480` by :user:`adityasingh2400`
 
+- Fix the loss and the gradient minimized by
```
qbarthelemy:
```suggestion
- Correct the loss and the gradient minimized by
```

tests/test_optimization.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@pytest.mark.parametrize("metric", ["euclid", "riemann"])
+def test_grassmann_loss(rndstate, metric):
+    """Test that loss is the weighted sum of squared distances"""
+    n_matrices, n_channels = 3, 4
+    X = make_matrices(n_matrices, n_channels, "spd", rs=rndstate)
+    Y = make_matrices(n_matrices, n_channels, "spd", rs=rndstate)
+    weights = np.array([0.5, 0.3, 0.2])
+    Q = np.linalg.qr(rndstate.randn(n_channels, n_channels))[0]
```
qbarthelemy:
```suggestion
def test_grassmann_loss(metric, get_mats, get_weights):
    """Test that loss is the weighted sum of squared distances"""
    n_matrices, n_channels = 3, 4
    X = get_mats(n_matrices, n_channels, "spd")
    Y = get_mats(n_matrices, n_channels, "spd")
    weights = get_weights(n_matrices)
    Q = get_mats(1, n_channels, "orth")[0]
```

tests/test_optimization.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@pytest.mark.parametrize("metric", ["euclid", "riemann"])
+def test_grassmann_grad(rndstate, metric):
+    """Test that gradient is the derivative of the loss"""
+    n_matrices, n_channels = 3, 4
+    X = make_matrices(n_matrices, n_channels, "spd", rs=rndstate)
+    Y = make_matrices(n_matrices, n_channels, "spd", rs=rndstate)
+    weights = np.array([0.5, 0.3, 0.2])
+    Q = np.linalg.qr(rndstate.randn(n_channels, n_channels))[0]
```
qbarthelemy:
```suggestion
def test_grassmann_grad(metric, get_mats, get_weights):
    """Test that gradient is the derivative of the loss"""
    n_matrices, n_channels = 3, 4
    X = get_mats(n_matrices, n_channels, "spd")
    Y = get_mats(n_matrices, n_channels, "spd")
    weights = get_weights(n_matrices)
    Q = get_mats(1, n_channels, "orth")[0]
```

pyriemann/optimization/grassmann.py
qbarthelemy, line 208:
```suggestion
    .. versionadded:: 0.4
    .. versionchanged:: 0.13
        Correct loss and gradient.
```

qbarthelemy (changes requested):

adityasingh2400:
All four applied in `861066c`, thanks.

The two tests now use the repo fixtures rather than calling `make_matrices` and `rndstate` directly, including `get_mats(1, n_channels, "orth")[0]` for `Q` and `get_weights(n_matrices)` for the weights. That also picks up the backend parametrization for free, so the file is now 4 passed and 4 skipped instead of 4 passed.

Wording changed to "Correct the loss and the gradient", and `versionchanged:: 0.13` added next to `versionadded:: 0.4` in `_get_rotation_manifold`.

I also merged master in, since #483 landed and collided on the `whatsnew` entry. Same for #479, #481 and #493, which hit the same collision.

Reverting the source still fails all 4 parametrizations, and `test_optimization.py` plus `test_transfer.py` together are 196 passed, 131 skipped.

tests/test_optimization.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
+from numpy.testing import assert_array_almost_equal
+import pytest
+from pytest import approx
+
+from pyriemann.geometry.distance import distance
+from pyriemann.optimization.grassmann import _grad, _loss
+
+
```
qbarthelemy, line 13:
```suggestion
pytestmark = pytest.mark.numpy_only


```

adityasingh2400: Applied. Confirmed it is the right call rather than just deferring to it: `grassmann.py` is numpy throughout, `np.linalg`, `np.einsum` and the finite-difference loop, so there is nothing for the torch backend to exercise here. This also matches how `test_artifactdetection.py` and four other test modules declare it.

The count is unchanged at 4 passed, 4 skipped, the difference being that the skips are now explicit rather than incidental.

tests/test_optimization.py
```diff
@@ -0,0 +1,49 @@
...
+            step = np.zeros((n_channels, n_channels))
+            step[i, j] = eps
+            grad_num[i, j] = (
+                _loss(Q + step, X, Y, weights, metric=metric)
+                - _loss(Q - step, X, Y, weights, metric=metric)
+            ) / (2 * eps)
+
+    grad = _grad(Q, X, Y, weights, metric=metric)
+    assert grad.dtype == grad_num.dtype
+    assert_array_almost_equal(grad, grad_num, decimal=5)
```
qbarthelemy, line 57: A last comment.
Can you add a test `test_get_rotation_manifold` testing that output is a rotation matrix for `metric` in {"euclid", "riemann"}, using
```
def _is_orth(X):
    return X @ X.T == approx(np.eye(X.shape[0]))
```

adityasingh2400:
`build_docs` went red on `0d2a824` and it is not from this branch, so flagging it before it costs you a look.

The failure is `plot_clusterer_comparison.py`, and the traceback is entirely inside `GaussianMixture`:

```
plot_clusterers("riemann") -> clt.fit(X)
  clustering.py:873  self._components[k].update_mean(...)
  clustering.py:658  update_mean -> gmean -> mean_riemann
  base.py:175        expm -> _matrix_operator
ValueError: Matrices must be positive definite.
```

Four things put it outside this PR:

1. This branch touches `grassmann.py`, `tests/test_optimization.py` and `whatsnew.rst`. It does not touch `clustering.py`, and nothing on that traceback imports `grassmann`.
2. `build_docs` passed on `861066c`, the commit immediately before, on this same branch. The only change since is one line adding `pytestmark` to a test module, which no example imports.
3. `build_docs` is currently green on #479, #481 and #493, which carry the same merge of master.
4. The clusterer is built as `GaussianMixture(n_components=3)` with no `random_state`, and the class documents that argument as the generator used to initialize the Gaussian models. The dataset is seeded with `RandomState(2025)` but the EM initialization is not, so a run can land on an initialization where `mean_riemann` hits a non positive definite intermediate.

So it looks like an unseeded-initialization flake in that example rather than a regression. Seeding the clusterer there would make it reproducible, though that is a separate change from this PR and I did not want to fold it in. Happy to open it separately if you think it is worth pinning down.

qbarthelemy:
Yes, this issue with GaussianMixture has been observed for some time. It should be fixed in a separate PR. Thanks a lot!

tests/test_optimization.py
adityasingh2400: Added in `586448a`, parametrized over both metrics.

I used your `_is_orth` as the only check rather than also asserting `det == +1`. `_get_rotation_manifold` returns whichever of its two minimizations gives the smaller loss, the one started on rotations and the one started on reflections, so an output with `det == -1` is valid by design. I tried the stricter assertion first and it failed on the euclid case with exactly that, which matches the two branches.

qbarthelemy (commented):
Thanks, @adityasingh2400 , for your corrections—as insightful as ever!

PR #493, opened 2026-09-07
Fix __init__ of KmeansPerClassTransform not storing its parameters

adityasingh2400:
Split out of #483 at @qbarthelemy's request. #483 now carries only the `MeanShift` bandwidth fix.

`KmeansPerClassTransform.__init__` built a `Kmeans` and kept only that, so the constructor arguments were never stored on the estimator. `BaseEstimator.get_params()` reads them back by attribute, so anything relying on it raised `AttributeError: 'KmeansPerClassTransform' object has no attribute 'n_clusters'`. That covers `get_params()`, `set_params()`, `clone()`, and therefore `cross_val_score()` and every scikit-learn search.

This uses the approach you suggested, which avoids restating the `Kmeans` signature:

```python
def __init__(self, **params):
    defaults = Kmeans().get_params()
    defaults.update(params)
    self.params = defaults
    for name, value in defaults.items():
        setattr(self, name, value)
```

with `get_params`/`set_params` reading and writing `self.params`, and `fit` constructing `Kmeans(**self.params)`. One consequence worth flagging: `n_clusters` is no longer an explicit keyword in the signature, it now comes from the `Kmeans` defaults like every other parameter. `KmeansPerClassTransform(n_clusters=3)` still works.

**Tests.** `test_clustering_get_params` checks `get_params`, `set_params` and `clone` round-trip for every clustering estimator, and `test_kmeansperclasstransform_get_params` pins the exposed keys to `Kmeans`'s. Both fail on unpatched `master` and pass here. Whole file is 62 passed, 60 skipped.

Checked end to end as well:

```
get_params keys == Kmeans keys: True
clone works: True
set_params: riemann
cross_val_score: [0.643 0.538 0.462]
```

https://claude.ai/code/session_0189zxefNUxZmXLRED6jPeMy

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -76,6 +76,12 @@ v0.13.dev
   each source domain onto the target domain.
   :pr:`481` by :user:`adityasingh2400`
 
+- Fix ``__init__`` of
+  :class:`pyriemann.clustering.KmeansPerClassTransform`, which did not store
+  its parameters, so that ``get_params()``, ``set_params()``, ``clone()`` and
+  ``cross_val_score()`` raised an ``AttributeError``.
+  :pr:`493` by :user:`adityasingh2400`
+
 v0.12 (July 2026)
 -----------------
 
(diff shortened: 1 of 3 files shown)


qbarthelemy (approved):
Thx @adityasingh2400 !

PR #494, opened 2026-09-10
Speed up covariances_EP for the sample covariance

gcattan:
Some optim, for the new version of QSD in pyRiemmann qiskit:

Note, I cannot apply the same to OAS that lies in sklearn. 
That makes further comparison to xdawncov unfair for P300.

@toncho11 @qbarthelemy fyi

covariances_EP broadcast the prototype P over the matrices and concatenated before estimating, which recomputed the constant P.P^T block once per matrix and materialized an (..., n_channels_proto + n_channels, n_times) temporary. The sample covariance is bilinear, so it splits into blocks

    [[P.P^T, P.X^T],
     [X.P^T, X.X^T]]

where only the last two depend on the matrix: P.P^T is estimated once and nothing is concatenated. Flops drop from N(p+c)^2 t to N(2pc + c^2) t, about 2x when p == c, and the measured gain is larger because the concatenation is also avoided: about 3x on a set of 5000 matrices.

The fast path is limited to "scm", the only estimator that is this decomposition, and to real 3d NumPy input with no extra keyword arguments. Shrinkage estimators are nonlinear functions of the whole matrix, and they, complex and 2d input, and the PyTorch namespace all take the general path. Results are unchanged: verified against the previous implementation over estimators scm, cov, oas, lwf, corr and sch.

### DIFF
--- a/doc/whatsnew.rst
+++ b/doc/whatsnew.rst
@@ -10,6 +10,13 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.13.dev
 ---------
 
+- Speed up :func:`pyriemann.geometry.covariance.covariances_EP` for the
+  ``"scm"`` estimator, by computing the covariance blockwise rather than
+  concatenating a broadcast prototype, so that the constant prototype block
+  is estimated once rather than once per matrix: about 3x faster on a set of
+  5000 matrices. Results are unchanged.
+  :pr:`XXX` by :user:`gcattan`
+
 - Update pyRiemann from Python 3.10 - 3.12 to 3.11 - 3.13.
   :pr:`462` by :user:`qbarthelemy`
 
@@ -358,7 +365,7 @@ v0.7 (October 2024)
 - Enhance :class:`pyriemann.spatialfilters.CSP` adding parameter ``ajd_method``.
   :pr:`313` by :user:`qbarthelemy`
 
-- Add :func:`pyriemann.geometry.distance.distance_poweuclid` and 
+- Add :func:`pyriemann.geometry.distance.distance_poweuclid` and
   :func:`pyriemann.geometry.mean.mean_poweuclid` to use power Euclidean metric.
   :pr:`312` by :user:`qbarthelemy`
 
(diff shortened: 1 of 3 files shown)


pyriemann/geometry/covariance.py
```diff
@@ -1,6 +1,7 @@
 from functools import wraps
 import warnings
 
+import numpy as np
```
qbarthelemy: You cannot import NumPy. The proposed change must use the array API.

pyriemann/geometry/covariance.py
```diff
@@ -518,13 +519,62 @@ def covariances_EP(X, P, estimator="cov", **kwds):
     if n_times_p != n_times:
         raise ValueError(
             f"X and P do not have the same n_times: {n_times} and {n_times_p}")
+
+    fast = _covariances_EP_blockwise(X, P, estimator, **kwds)
+    if fast is not None:
+        return fast
```
qbarthelemy:
```suggestion
    if _is_fast_possible(X, estimator) 
        return _covariances_EP_fast(X, P, estimator, **kwds)
```

doc/whatsnew.rst
```diff
@@ -10,6 +10,13 @@ A catalog of new features, improvements, and bug-fixes in each release.
 v0.13.dev
 ---------
 
+- Speed up :func:`pyriemann.geometry.covariance.covariances_EP` for the
+  ``"scm"`` estimator, by computing the covariance blockwise rather than
+  concatenating a broadcast prototype, so that the constant prototype block
+  is estimated once rather than once per matrix: about 3x faster on a set of
+  5000 matrices. Results are unchanged.
+  :pr:`XXX` by :user:`gcattan`
```
qbarthelemy: Must be at the bottom of the section.

qbarthelemy (commented):
Thx @gcattan for this suggestion.

I have a lot of hesitation about adding this code: the time saved is undeniable, but it's a lot of code to maintain in `pyriemann` for such a limited use case: estimator=scm, without weights, only for real-valued data and only for NumPy.

gcattan:
Sure,no problem. I understand.

